PR #33, opened 2019-10-25
adding include in ios ua

sayan-rudder:


sayan-rudder:
merging as no major change

PR #36, opened 2019-11-07
v0 - Dest segment

Sampath-vi:


v0/segment/transform.js
```diff
@@ -0,0 +1,116 @@
+const get = require("get-value");
+let { destinationConfigKeys, batchEndpoint } = require("./config");
+const { defaultPostRequestConfig } = require("../util");
+
+function responseBuilderSimple(payload, segmentConfig) {
+  let basicAuth = new Buffer(`${segmentConfig.writeKey}:`).toString("base64");
```
SumanthPuram: const

v0/segment/transform.js
```diff
@@ -0,0 +1,116 @@
...
+function makePayload(
+  type,
+  userId,
+  event,
+  traits,
+  properties,
+  timeStamp,
+  segmentConfig
+) {
+  let eventPayload = {};
```
SumanthPuram: const eventPayload = {
  type,
  userId ...
}

PR #48, opened 2019-12-06
Updated test case and results

prabrisha-rudder:


v0/ga/transform.js
```diff
@@ -50,14 +51,15 @@ function responseBuilderSimple(
...
 
-  const response = {
-    endpoint: GA_ENDPOINT,
-    requestConfig: defaultGetRequestConfig,
-    header: {},
-    userId: message.anonymousId,
-    payload: finalPayload
-  };
-  //console.log("response ", response);
+  const response = JSON.parse(JSON.stringify(defaultRequestConfig));
```
lokey: We can remove JSON.parse and JSON.stringify here and use defaultRequestConfig. We can make defaultRequestConfig a function that returns the same object instead of making a copy using parse and stringify.

PR #94, opened 2020-03-16
Release for LeanPlum & Generic Webhook

arnabp92:
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-transformer/94)


v0/customerio/transform.js
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    set(
+      rawPayload,
+      "created_at",
+      Math.floor(new Date(message.originalTimestamp).getTime() / 1000)
+    );
+    //console.log(rawPayload);
+
+    if (userId) {
```
SumanthPuram: Let's move this check to the beginning of the if block. As a practice, whenever we can return from a function those checks should be at the top.

v0/customerio/transform.js
```diff
@@ -0,0 +1,222 @@
...
+function responseBuilder(message, evType, evName, destination) {
+  const rawPayload = {};
+  let endpoint;
+  let requestConfig = defaultPostRequestConfig;
+  //console.log(message);
+  let userId =
+    message.userId && message.userId != "" ? message.userId : undefined;
+
+  const response = defaultRequestConfig();
+  response.userId = message.userId ? message.userId : message.anonymousId;
```
SumanthPuram: response.userId = message.userId || message.anonymousId.

v0/customerio/transform.js
```diff
@@ -0,0 +1,222 @@
...
+
+      if (deviceDeleteRelatedEventName == evName) {
+        const token = get(message, "context.device.token");
+        if (userId && token) {
+          endpoint = DEVICE_DELETE_ENDPOINT.replace(":id", userId).replace(
+            ":device_id",
+            token
+          );
+
+          response.endpoint = endpoint;
```
SumanthPuram: Setting everything together is more readable.

{
  endpoint,
  method: "DELETE",
  headers: {
  }
}

v0/customerio/transform.js
```diff
@@ -0,0 +1,222 @@
...
+          response.method = "DELETE";
+          response.headers = {
+            Authorization:
+              "Basic " +
+              btoa(destination.Config.siteID + ":" + destination.Config.apiKey)
+          };
+
+          return response;
+        } else {
```
SumanthPuram: Else can be skipped. Anyways, it's better to return error in the if.
I like this pattern.

https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

v0/customerio/transform.js
```diff
@@ -0,0 +1,222 @@
...
+      evType = "identify";
+      break;
+    case EventType.PAGE:
+      evType = "page"; // customerio mandates sending 'page' for pageview events
+      evName = message.name ? message.name : message.properties.url;
+      break;
+    case EventType.SCREEN:
+      evType = "event";
+      evName =
+        "Viewed " + message.name
```
SumanthPuram: Can we put the part after Viewed in ().
Also since JS has no type, we could merge some ternary operator to || statements.

v0/leanplum/transform.js
```diff
@@ -0,0 +1,177 @@
...
+async function startSession(message, destination) {
+  let retryCount = 0;
+  let success = false;
+  const payload = constructPayload(
+    message,
+    ConfigCategory.START.name,
+    destination
+  );
+  // console.log(payload);
+  const url = ENDPOINT + "?action=start";
```
SumanthPuram: Let's send as array of two events back to router. We can remove the RETRY logic,

v0/leanplum/transform.js
```diff
@@ -0,0 +1,177 @@
...
+  // all event types except idetify requires startSession
+  if (messageType !== EventType.IDENTIFY) {
+    await startSession(message, destination);
+  }
+
+  return response;
+}
+
+async function process(event) {
+  const resp = await processSingleMessage(event.message, event.destination);
```
SumanthPuram: Merge into 1 line? 
It is okay this way as well, but just less number of lines make it easier to read. Might be my obsession, you decide whichever way.

v0/webhook/transform.js
```diff
@@ -0,0 +1,35 @@
+const {
+  defaultPostRequestConfig,
+  defaultRequestConfig,
+  isURL
+} = require("../util");
+
+function process(event) {
+  const { message, destination } = event;
+  const response = defaultRequestConfig();
+  if (destination.Config.webhookUrl && isURL(destination.Config.webhookUrl)) {
```
SumanthPuram: Regex checks are generally costly. Let's move this check to config-backend/webapp and assume this is a valid URL.

v0/webhook/transform.js
```diff
@@ -0,0 +1,35 @@
...
+    const configHeaders = destination.Config.header;
+    if (configHeaders) {
+      configHeaders.forEach(header => {
+        response.headers[header.from] = header.to;
+      });
+    }
+
+    response.userId = message.userId ? message.userId : message.anonymousId;
+    response.body.JSON = { ...message };
+    if (!response.statusCode) {
```
SumanthPuram: Do we need if here?

v0/webhook/transform.js
```diff
@@ -0,0 +1,35 @@
...
+  const response = defaultRequestConfig();
+  if (destination.Config.webhookUrl && isURL(destination.Config.webhookUrl)) {
+    response.endpoint = destination.Config.webhookUrl;
+    response.method = defaultPostRequestConfig.requestMethod;
+    // add content-type by default as our payload is json
+    response.headers = {
+      "Content-Type": "application/json"
+    };
+
+    const configHeaders = destination.Config.header;
```
SumanthPuram: Should we change the config-backend param to headers as well (instead of header)

v0/webhook/transform.js
```diff
@@ -0,0 +1,35 @@
...
+    if (configHeaders) {
+      configHeaders.forEach(header => {
+        response.headers[header.from] = header.to;
```
SumanthPuram: Shouldn't this be header.name and header.value?

v0/webhook/transform.js
```diff
@@ -0,0 +1,35 @@
...
+} = require("../util");
+
+function process(event) {
+  const { message, destination } = event;
+  const response = defaultRequestConfig();
+  if (destination.Config.webhookUrl && isURL(destination.Config.webhookUrl)) {
+    response.endpoint = destination.Config.webhookUrl;
+    response.method = defaultPostRequestConfig.requestMethod;
+    // add content-type by default as our payload is json
+    response.headers = {
```
SumanthPuram: Our server's network.go support FORM body too. Do we want to restrict webhook only for JSON. 
That is good infact, when someone asks we can enable support for FORM.

v0/customerio/transform.js
arnabp92: Done

v0/customerio/transform.js
arnabp92: Done

arnabp92:
@SumanthPuram Review comments are taken care of. Please check.

v0/customerio/transform.js
arnabp92: The post is really good. I'd try to enforce this pattern going forward

v0/customerio/transform.js
arnabp92: Left it as it is now. Once we come back here for someone, will do a round of code cleanup.

v0/customerio/transform.js
arnabp92: Updated

v0/leanplum/transform.js
arnabp92: Updated

v0/leanplum/transform.js
arnabp92: Merging in one line is giving a lint error. Kept it to avoid that.

v0/webhook/transform.js
arnabp92: Removed the regex check

v0/webhook/transform.js
arnabp92: statusCode is not needed. Removed it

v0/webhook/transform.js
arnabp92: Didn't get you completely. But I think we are not adding extra keys name and value. Kept it as it is

v0/webhook/transform.js
arnabp92: Makes sense

v0/webhook/transform.js
arnabp92: Updated both control-plane and here

v0/customerio/transform.js
```diff
@@ -158,40 +147,31 @@ function responseBuilder(message, evType, evName, destination) {
...
-      evName = message.name ? message.name : message.properties.url;
+      evName = message.name || message.properties.url;
       break;
     case EventType.SCREEN:
       evType = "event";
-      evName =
-        "Viewed " + message.name
-          ? message.name
-          : message.properties.name + " Screen";
+      evName = "Viewed " + message.name || message.properties.name + " Screen";
```
SumanthPuram: "Viewed " + (message.name || message.properties.name) + " Screen" 
For better readability?

v0/leanplum/transform.js
```diff
@@ -96,42 +92,7 @@ function responseBuilderSimple(message, category, destination) {
 }
 
 async function startSession(message, destination) {
```
SumanthPuram: This not an async function anymore. We can remove async and the corresponding await.

v0/webhook/transform.js
SumanthPuram: Ideally these fields should be named header.name and header.value. 

For a header, _from_ and _to_ is not the right convention.

v0/leanplum/transform.js
```diff
@@ -161,12 +122,14 @@ async function processSingleMessage(message, destination) {
   // build the response
   const response = responseBuilderSimple(message, category, destination);
 
+  let respList = [];
+
   // all event types except idetify requires startSession
   if (messageType !== EventType.IDENTIFY) {
-    await startSession(message, destination);
+    respList = [await startSession(message, destination)];
```
SumanthPuram: We don't need await now.

v0/webhook/transform.js
arnabp92: Yes. But `from` and `to` is coming from our `config-backend` response.

PR #97, opened 2020-03-22
Handle reserved keywords for warehouses

psrikanth88:
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-transformer/97)


psrikanth88:
> Very elaborate. I'd have just added one special reserved keys file and crammed everything into it, instead of three. Nicely done. But, we got to remember to add any new keywords(hopefully none) in 3 files.
> 
> Does it handle backward compatibility for torpedo etc?

@varkalablr Each warehouse had totally different set of reserved keywords, so placed them separately. Torpedo did not use any reserve keywords till now, so should work for it

PR #133, opened 2020-04-20
MixPanel bug Fixes

sajalrudderlabs:
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-transformer/133)


sajalrudderlabs:
If any nested properties sent along with a `track` call , it is not getting flattened

PR #170, opened 2020-06-22
WH lint fix

arnabp92:
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-transformer/170)


v0/destinations/util.js
```diff
@@ -66,6 +53,7 @@ const updatePayload = (currentKey, eventMappingArr, value, payload) => {
     if (obj.rudderKey === currentKey) {
```
SumanthPuram, line 53: Instead of map, we should be using a while/for-each construct here.

versionedRouter.js
```diff
@@ -114,7 +118,7 @@ if (startDestTransformer) {
 
       const transformedEvents = [];
       await Promise.all(
-        Object.entries(groupedEvents).map(async ([dest, destEvents]) => {
+        Object.entries(groupedEvents).map(async ([destEvents]) => {
```
SumanthPuram, line 121: This needs to be fixed. The first argument will be destID and the second is list of events.

Example here: https://repl.it/repls/SmoggyNextCrash

warehouse/v0/util.js
```diff
@@ -10,9 +10,10 @@ const toSnakeCase = str => {
 };
 
 const toSafeDBString = str => {
-  if (parseInt(str[0]) > 0) str = `_${str}`;
-  str = str.replace(/[^a-zA-Z0-9_]+/g, "");
-  return str.substr(0, 127);
+  let parsedStr = str;
+  if (parseInt(str[0], 10) > 0) parsedStr = `_${str}`;
```
SumanthPuram, line 14: Make it more readable?
```
if (parseInt(str[0], 10) > 0) {
  parsedStr = `_${str}`;
}
 
```

v0/destinations/util.js
arnabp92: Done

versionedRouter.js
arnabp92: Done

warehouse/v0/util.js
arnabp92: Done

PR #174, opened 2020-06-25
alter string to text for rs if char limit is more than 512

gane5hvarma:
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-transformer/174)


warehouse/index.js
```diff
@@ -10,6 +9,8 @@ const whPageConfigJson = require("./config/WHPageConfig.json");
 const whScreenConfigJson = require("./config/WHScreenConfig.json");
 const whGroupConfigJson = require("./config/WHGroupConfig.json");
 const whAliasConfigJson = require("./config/WHAliasConfig.json");
+// redshift destination string limit, if the string length crosses 512 we will change data type to text which is varchar(max) in redshift
+const RSStringLimit = 512;
```
SumanthPuram, line 13: This file should be generic, if we add destination specific code in this then it becomes hard to read/maintain going forward.

SumanthPuram: One approach could be to change the signature of processSingleMessage.

```
*** warehouse/index.js
--  function processWarehouseMessage(provider, message, schemaVersion, RSAlterStringToText = "false") 

++ function processWarehouseMessage(provider, message, options = {}) 

-- const getDataType = (val)

++ const getDataType = (val, options) => {
  if options.getDataType && typeof options.getDataType === "function" {
     return options.getDataType(val)
  }
}
```

```
*** rs/transform.js
++ const RSStringLimit = 512;

++ func getDataType(val) {
++ // Do redshift specific handling here. Pass this function to warehouse index.js
++ }

-- processWarehouseMessage("rs", message, schemaVersion, RSAlterStringToText);

++ processWarehouseMessage("rs", message, {schemaVersion, getDataType});
```

warehouse/index.js
```diff
@@ -52,6 +52,9 @@ const getDataType = val => {
   if (validTimestamp(val)) {
     return "datetime";
   }
+  if (options.getDataTypeOverride && typeof options.getDataTypeOverride === "function") {
+    return options.getDataTypeOverride(val, options);
```
psrikanth88, line 56: name the func string specific? this will override only the string datatype right

v0/destinations/rs/transform.js
```diff
@@ -1,12 +1,27 @@
...
+  }
+  return "string"
 }
 
 function process(event) {
   const schemaVersion = event.request.query.whSchemaVersion || "v1";
-  return processSingleMessage(event.message, schemaVersion);
+  const RSAlterStringToText = event.request.query.RSAlterStringToText || "false";
+  const provider = redshift
+  return processSingleMessage(event.message, { schemaVersion, getDataTypeOverride, provider, RSAlterStringToText });
```
psrikanth88, line 24: need to change the function signature in other warehouse transform.js also

psrikanth88: change query parameter to start with lower case letter?

warehouse/index.js
gane5hvarma: yes. but kept it generic

PR #176, opened 2020-06-26
Api Version for transformer <-> backend compatibility

sivashanmukh:
Sending an apiVersion as a response header for destination transformations to check against in rudder-server for compatibility

---
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-transformer/176)

### DIFF
--- a/versionedRouter.js
+++ b/versionedRouter.js
@@ -6,6 +6,7 @@ const { lstatSync, readdirSync } = require("fs");
 const logger = require("./logger");
 
 const versions = ["v0"];
+const API_VERSION = '1'
 
 const transformerMode = process.env.TRANSFORMER_MODE;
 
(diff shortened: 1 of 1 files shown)


versionedRouter.js
```diff
@@ -83,6 +83,7 @@ async function handleDest(ctx, destHandler) {
   );
   logger.debug(`[DT] Output events: ${JSON.stringify(respList)}`);
   ctx.body = respList;
+  ctx.set('apiVersion', '1')
```
arnabp92, line 86: Instead of a constant string, should we take from `.version`?

sivashanmukh: So the idea of this is that only if this matches with the version defined in this PR -> https://github.com/rudderlabs/rudder-server/pull/380/files

With changing versions in transformer, we don't want an update in rudder-backend for every transformer version change.

We should look at this as a version that forces a change in rudder-backend.
That being said, instead of using a constant here, picking it up from somewhere else might make sense.

arnabp92: Makes sense!

versionedRouter.js
```diff
@@ -83,6 +84,7 @@ async function handleDest(ctx, destHandler) {
   );
   logger.debug(`[DT] Output events: ${JSON.stringify(respList)}`);
   ctx.body = respList;
+  ctx.set('apiVersion', API_VERSION)
```
psrikanth88, line 87: we need to add this in custom transformer and source transformer response also

PR #195, opened 2020-07-10
Auth0 source implementation

arnab-p:


v0/sources/auth0/transform.js
```diff
@@ -0,0 +1,60 @@
...
+      // responses.push({
+      //   statusCode: 400,
+      //   error: error.message || "Unknwon error"
+      // });
+    }
+  });
+  if (responses.length === 0) {
+    throw new Error("All requests in the batch failed");
+  } else {
+    return { batch: responses };
```
psrikanth88, line 56: can we move this adding batch key to versionedRouter. Otherwise, someone might forget adding it one of the other source transformers and it will cause an issue

arnab-p: Done

PR #221, opened 2020-07-30
Indicative Destination

arnab-p:


__tests__/data/indicative_output.json
```diff
@@ -0,0 +1,187 @@
...
+      "traits": {
+        "address": {
+          "city": "Dhaka",
+          "country": "Bangladesh"
+        },
+        "anonymousId": "c82cbdff-e5be-4009-ac78-cdeea09ab4b1"
+      }
+    },
+    "event": "spin_result",
+    "integrations": {
```
sayan-mitra, line 27: will this will be part of output?

v0/destinations/indicative/data/INIdentifyConfig.json
```diff
@@ -0,0 +1,12 @@
+[
+  {
+    "destKey": "uniqueId",
+    "sourceKeys": ["userId", "context.traits.userId", "context.traits.id"],
```
sayan-mitra, line 4: should we take userId in case of identify from anywhere else except toplevel? Bcz context is modifiable by user but top level userId is always set by sdks for identify

v0/destinations/indicative/data/INAliasConfig.json
```diff
@@ -0,0 +1,12 @@
+[
+  {
+    "destKey": "previousId",
+    "sourceKeys": ["previousId", "anonymousId"],
+    "required": true
+  },
+  {
+    "destKey": "newId",
+    "sourceKeys": ["userId", "context.traits.userId", "context.traits.id"],
```
sayan-mitra, line 9: similar to below cooment

v0/destinations/indicative/data/INPageConfig.json
```diff
@@ -0,0 +1,29 @@
...
+      "userId",
+      "context.traits.userId",
+      "context.traits.id",
+      "anonymousId"
+    ],
+    "required": true
+  },
+  {
+    "destKey": "eventName",
+    "sourceKeys": "event",
```
sayan-mitra, line 14: page doesn't have event correct? should be name?

v0/destinations/indicative/data/INPageConfig.json
```diff
@@ -0,0 +1,29 @@
...
+    "required": true,
+    "metadata": {
+      "template": "Visited {{value}} page"
```
sayan-mitra, line 17: what does segment do here?

v0/destinations/indicative/transform.js
```diff
@@ -0,0 +1,63 @@
...
+const responseBuilderSimple = (message, category, destination) => {
+  const payload = constructPayload(message, MAPPING_CONFIG[category.name]);
+  const responseBody = { ...payload, apiKey: destination.Config.apiKey };
+  const response = defaultRequestConfig();
+  response.endpoint = category.endPoint;
+  response.method = defaultPostRequestConfig.requestMethod;
+  response.headers = {
+    "Content-Type": "application/json"
+  };
+  response.userId = getFieldValueFromMessage(message, "userId");
```
sayan-mitra, line 19: this I feel should always be anonymousId, that way may help in server side to not switch between rt job workers, wdyt

v0/destinations/indicative/transform.js
```diff
@@ -0,0 +1,63 @@
+const { EventType } = require("../../../constants");
```
sayan-mitra, line 1: does indicative has id resolution thing? did we check with the user flow from unidentified tracking to identified tracking

v0/util/data/message.json
```diff
@@ -0,0 +1,8 @@
+{
```
sayan-mitra, line 1: have a better name for this file?

v0/util/index.js
```diff
@@ -0,0 +1,341 @@
...
+        }
+      } else if (required) {
+        // throw error if reqired value is missing
+        throw new Error(
+          `Missing required value from ${JSON.stringify(sourceKeys)}`
+        );
+      }
+    });
+  }
+  return null;
```
sayan-mitra, line 235: return payload ?

__tests__/data/indicative_output.json
arnab-p: No. Output file is not updated. Doing that.

v0/destinations/indicative/data/INIdentifyConfig.json
arnab-p: Agreed. This is for fail-safety given HTTP APIs can send any data. Idea is to maximize the possibilities.

v0/destinations/indicative/data/INPageConfig.json
arnab-p: Updated

v0/destinations/indicative/transform.js
arnab-p: Agreed, but that way, we have to be consistent across destinations. Let's consider this as a part of refactoring.

v0/util/data/message.json
arnab-p: Suggestions please!

v0/destinations/indicative/data/INIdentifyConfig.json
sayan-mitra: okay

v0/destinations/indicative/transform.js
sayan-mitra: yes..let's take a note

v0/util/data/message.json
sayan-mitra: GenericFieldMapping?

arnab-p: 👍

v0/util/index.js
arnab-p: Fixed

v0/destinations/indicative/transform.js
arnab-p: Yes. Updated identify logic to include alias call to map the userId correctly

v0/util/index.js
```diff
@@ -0,0 +1,351 @@
...
+const handleMetadataForValue = (value, metadata) => {
+  let formattedVal = value;
+  const { type, typeFormat, template } = metadata;
+
+  // handle type and format
+  // skipping check for typeFormat to support default for each type
+  if (type) {
+    switch (type) {
+      case "timestamp":
+        formattedVal = formatTimeStamp(formattedVal, typeFormat);
```
prabrishac, line 200: as there could be multiple timestamp formats, we may either put different type for them in metadata or put a subtype as "UTC" and based on that call different methods or pass it in the same method as argument.

PR #223, opened 2020-07-31
Transfomations code flexibility

sivashanmukh:
Pool management for reusing ivms out of a resource pool to reduce over head of creation and destructions of ivms


logger.js
```diff
@@ -5,9 +5,7 @@ const levelInfo = 1; // Logs about state of the application
 const levelError = 2; // Logs about errors which dont immediately halt the application
 const levelNone = 3; // Nothing is logged
 
-const logLevel = process.env.LOG_LEVEL
-  ? parseInt(process.env.LOG_LEVEL, 10)
-  : levelInfo;
+const logLevel = 0;
```
arnab-p, line 8: You might want to revert this?

PR #229, opened 2020-08-03
made changes for heap

ruchiramoitra:


v0/destinations/heap/transform.js
```diff
@@ -39,20 +83,32 @@ function commonPayload(message, rawPayload, type) {
...
+  let email;
   switch (type) {
     case EventType.TRACK:
       propsArray = get(message, "properties")
         ? Object.keys(message.properties)
         : null;
       rudderPropertiesObj = message.properties;
-      rawPayload.identity = message.context.traits.email;
+      email =
+        (message.context.traits ? email : null) || message.properties.email;
```
sayan-mitra, line 94: (message.context.traits ? email : null)  ==> email is undefined here right?
so rawPayload.identity = null is valid but rawPayload.identity = undefined (previous version) is not?

v0/destinations/heap/transform.js
```diff
@@ -39,20 +83,32 @@ function commonPayload(message, rawPayload, type) {
...
+      else
+        throw Error("Email for identity is required for sending track calls.");
       break;
     case EventType.IDENTIFY:
       propsArray = get(message.context, "traits")
         ? Object.keys(message.context.traits)
         : null;
+      email =
+        message.context.traits.email ||
+        (message.properties ? message.properties.email : null);
```
sayan-mitra, line 105: will identify have message.properties?

v0/destinations/heap/transform.js
```diff
@@ -7,7 +7,51 @@ const {
   defaultRequestConfig
 } = require("../util");
 
-function responseBuilder(payload, message, heapConfig) {
+function flatten(target, opts) {
```
sayan-mitra, line 10: heap needs  flattened properties?
we are not using opts, can this be removed? Also, move this method to utils?
can we have a description and an example of what it does with javascript object, object array and array types

ruchiramoitra: this function is also used in another destination Kissmetrics. there opts is used so making it generic and keeping it in utils also adding comments.

v0/destinations/heap/transform.js
ruchiramoitra: properties can be sent through identify calls right? 
Thus checking once and then assigning.

v0/destinations/heap/transform.js
ruchiramoitra: Yes heap was not accepting nested properties.

v0/destinations/heap/transform.js
sayan-mitra: do we have that provision in our sdks?

v0/destinations/heap/transform.js
ruchiramoitra: changed the logic

sayan-mitra:
👍

PR #239, opened 2020-08-04
Iterbale refactor

ruchiramoitra:


__tests__/data/iterable_input.json
```diff
@@ -908,5 +908,58 @@
       },
       "Enabled": true
     }
+  },
+  {
```
arnab-p, line 912: This test is not passing

v1/destinations/iterable/data/IterableIdentifyConfig.json
```diff
@@ -2,14 +2,16 @@
   {
     "destKey": "email",
     "sourceKeys": [
-      "context.traits.email"
+      "context.traits.email",
```
arnab-p, line 5: Need to put top-level traits with higher precedence. The same comment goes for similar changes.

__tests__/data/iterable_input.json
ruchiramoitra: done

v1/destinations/iterable/data/IterableIdentifyConfig.json
ruchiramoitra: changed

PR #248, opened 2020-08-05
add timestamp and tests fix

sayan-mitra:


v0/destinations/customerio/transform.js
```diff
@@ -100,7 +101,16 @@ function responseBuilder(message, evType, evName, destination) {
         Math.floor(new Date(message.originalTimestamp).getTime() / 1000)
       );
     }
-
+    if (getFieldValueFromMessage(message, "timestamp")) {
```
arnab-p, line 104: We're running the same loop again. Should we reduce that?

sayan-mitra: for toplevel keys with single sourcekey candidate, I think it won't affect
wdyt

arnab-p: Agreed

PR #254, opened 2020-08-12
Zendesk - Top level traits fix

dhawal1248:


dhawal1248:
> @dhawal1248 Should we update all the usage of `message.userId` with `getFieldValueFromMessage`?

For Zendesk, we need just `message.userId` and not other fallbacks like `context.traits.userId` or `anonymousId`.So we can't use `getFieldValueFromMessage `

arnab-p (approved):

PR #260, opened 2020-08-19
Webhook- Get Support

dhawal1248:


v1/destinations/webhook/transform.js
```diff
@@ -1,25 +1,41 @@
...
-    response.endpoint = destination.Config.webhookUrl;
-    response.method = defaultPostRequestConfig.requestMethod;
+  const url = destination.Config.webhookUrl;
+  const method = destination.Config.webhookMethod;
+  const headers = destination.Config.headers;
 
-    response.headers = {
-      "Content-Type" : "application/json"
-    };
+  if(url){
```
sayan-mitra, line 18: let's add a url format validator? (regex)

v1/destinations/webhook/transform.js
```diff
@@ -1,25 +1,41 @@
...
+    Object.assign(response.headers, getHashFromArray(headers));
     response.userId = getFieldValueFromMessage(message, "userId");
-
-    response.body.JSON = { ...message };
+    response.endpoint = url;
+    
     return response;
   }
-  throw new Error("Invalid Url in destination");
+  throw new Error("Invalid URL in destination config");
```
sayan-mitra, line 38: since we are adding some logic now, let's wrap the code in try and catch and use error.message || "failed"

sayan-mitra (changes requested):

PR #261, opened 2020-08-19
Add empty checks to Utils

dhawal1248:


v1/util/index.js
```diff
@@ -353,7 +355,9 @@ function flattenJson(data) {
       if (isEmpty && prop) result[prop] = {};
     }
   }
-  recurse(data, "");
+  if (data) {
```
sayan-mitra, line 358: should we add the validation for primitive types as well, like 
flattenJson("abc")/ flattenJson([1,2,3]) etc?

v1/util/index.js
```diff
@@ -42,9 +42,11 @@ const formatValue = value => {
 // Format the destination.Config.dynamicMap arrays to hashMap
 const getHashFromArray = (arrays, fromKey = "from", toKey = "to") => {
   const hashMap = {};
-  arrays.forEach(array => {
-    hashMap[array[fromKey]] = array[toKey];
-  });
+  if (arrays) {
```
sayan-mitra, line 45: let's do Array.isArray(arrays) here?
it wlll then safeguard objects too

sayan-mitra (changes requested):

v1/util/index.js
dhawal1248: Done

v1/util/index.js
dhawal1248: Done

sayan-mitra (approved):

PR #262, opened 2020-08-19
Develop-Webhook/get

dhawal1248:


v0/destinations/webhook/transform.js
```diff
@@ -1,39 +1,51 @@
...
+  if (message.type === EventType.IDENTIFY) {
+    return flattenJson(getFieldValueFromMessage(message, "traits"));
   }
-  throw new Error("Invalid Url in destination");
-}
+  return flattenJson(message.properties);
+};
 
-const process = event => {
+function process(event) {
```
arnab-p, line 18: Any benefit from moving from es6 to es5?

v0/destinations/webhook/transform.js
```diff
@@ -1,39 +1,51 @@
...
-    return handleEvent(event);
-  } catch (error) {
-    return {
-      statusCode: 400,
-      error: `${error.message}`
-    };
+    const { message, destination } = event;
+    const response = defaultRequestConfig();
+    const url = destination.Config.webhookUrl;
+    const method = destination.Config.webhookMethod;
```
arnab-p, line 23: Should we handle default value as well? 
```suggestion
    const method = destination.Config.webhookMethod || "POST";
```

v0/destinations/webhook/transform.js
```diff
@@ -1,39 +1,51 @@
...
+      } else {
+        response.method = defaultPostRequestConfig.requestMethod;
+        response.body.JSON = message;
+        response.headers = {
+          "Content-Type": "application/json"
+        };
+      }
+
+      Object.assign(response.headers, getHashFromArray(headers));
+      response.userId = getFieldValueFromMessage(message, "userId");
```
arnab-p, line 40: We decided to use `anonymousId` here always.

v0/destinations/webhook/transform.js
```diff
@@ -1,39 +1,51 @@
...
+    return flattenJson(getFieldValueFromMessage(message, "traits"));
   }
-  throw new Error("Invalid Url in destination");
-}
+  return flattenJson(message.properties);
+};
 
-const process = event => {
+function process(event) {
   try {
```
arnab-p, line 19: Do we need the `try-catch` block here?

v0/destinations/webhook/transform.js
```diff
@@ -1,39 +1,51 @@
...
-    return handleEvent(event);
-  } catch (error) {
-    return {
-      statusCode: 400,
-      error: `${error.message}`
-    };
+    const { message, destination } = event;
+    const response = defaultRequestConfig();
+    const url = destination.Config.webhookUrl;
```
arnab-p, line 22: Renaming the variables to `webhookUrl` and `webhookMethod` and using destructurization is more readable to me. 
```suggestion
    const { webhookUrl, webhookMethod } = destination.Config;
```
WDYT?

arnab-p (changes requested):

v0/destinations/webhook/transform.js
dhawal1248: `webhookMethod` wont be present for old destinations, which would cause an error here

v0/destinations/webhook/transform.js
dhawal1248: Since we have added some logic now(earlier it was very simple), so i guess its better to have a try catch

v0/destinations/webhook/transform.js
dhawal1248: Okay

v0/destinations/webhook/transform.js
dhawal1248: No, we could use either one

v0/destinations/webhook/transform.js
dhawal1248: It is handled in the code, we only check if the method is `GET`..if yes then we convert properties or traits into query params..otherwise we handle it as a `POST` request

dhawal1248: WDYT should we change?

v0/destinations/webhook/transform.js
arnab-p: Yes. Then let's stick to a particular one? ES6 being the latest, lets try to use ES6 as much as possible?

PR #272, opened 2020-08-25
WH:  Fix bug in setting column types

psrikanth88:


psrikanth88:
Bug:
For track and identify calls, common properties for both output tables are set and then properties specific to the table are set. But the same `columnTypes` object was used to store the datatype of each column which caused the bug. To repro, send a track event with key `id` inside `properties` and a boolean value. This will result in setting the correct `id` value (i.e messageId) in output tables but the datatype is incorrectly set to boolean.

SumanthPuram (approved):

PR #322, opened 2020-09-28
add boundaries for valid timestamps

gane5hvarma:
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-transformer/322)

### DIFF
--- a/warehouse/index.js
+++ b/warehouse/index.js
@@ -11,8 +11,8 @@ const whScreenConfigJson = require("./config/WHScreenConfig.json");
 const whGroupConfigJson = require("./config/WHGroupConfig.json");
 const whAliasConfigJson = require("./config/WHAliasConfig.json");
 
-const minTimeInMs = -62135596800000; // Date.parse("0001-01-01T00:00:00Z");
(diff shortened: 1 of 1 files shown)


psrikanth88 (approved):

warehouse/index.js
```diff
@@ -29,7 +31,17 @@ const timestampRegex = new RegExp(
 );
 
 function validTimestamp(input) {
-  return timestampRegex.test(input);
+  if (timestampRegex.test(input)) {
+    // check if date value lies in between min time and max time. if not then it's not a valid timestamp
+    const date = new Date(input).toISOString();
+    if (
+      Date.parse(minTimestamp) <= Date.parse(date) &&
```
SumanthPuram, line 38: We don't need to parse minTimestamp and maxTimestamp for every timestamp field. Precompute them at top of the file. This gets called for every event's timestamp fields, let's try to reduce the computation.

Don't need to parse `Date.parse(date)` twice.

SumanthPuram (commented):
Can be optimized.

SumanthPuram (approved):

chandumlg (approved):

PR #333, opened 2020-10-01
Facebook App Events fixes

dhawal1248:


v0/destinations/fb/transform.js
```diff
@@ -97,43 +115,75 @@ function sanityCheckPayloadForTypesAndModifications(updatedEvent) {
...
-          updatedEvent.custom_events[0][fbEventPath][length] =
-            intendValue || "";
-
-          length++;
-          count--;
+  const { properties } = message;
+  if (properties) {
+    if (properties.revenue && !properties.currency) {
+      throw new Error(
+        "If properties.revenue is present, properties.currency is required."
```
arnab-p, line 144: Can we set to something default like `USD`? Does FB do it?

arnab-p (changes requested):
There are bunch of TODO comments in the code. Should we remove it or kept for reference for future efforts?

v0/destinations/fb/transform.js
sayan-mitra: I think better to have explicit ?
https://developers.facebook.com/docs/marketing-api/app-event-api (fb_currency)

sayan-mitra:
> There are bunch of TODO comments in the code. Should we remove it or kept for reference for future efforts?

yeah

arnab-p (approved):

PR #345, opened 2020-10-09
WH: Changes to context_ip behavior

psrikanth88:


warehouse/config/WHDefaultConfig.js
```diff
@@ -0,0 +1,22 @@
+const { coalesce } = require("./helpers");
```
SumanthPuram, line 1: How about we reverse the config?

{
  "id": "messageId",  
  "user_id": "userId", 
  "context_request_ip": "request_ip", 
   ".." , 
  "context_ip": (msg) => msg .context.ip ||   msg.request_ip 
}

We check if the value is a function, If it is we fill it with the return value of the function. Makes it more readable.

SumanthPuram: This seems easier to read. Basically, we have columns in DB to fill and we are filling this way.

SumanthPuram (commented):
Minor comments

warehouse/config/WHDefaultConfig.js
```diff
@@ -1,22 +1,17 @@
...
-  }
+  id: "messageId",
+  anonymous_id: "anonymousId",
+  user_id: "userId",
+  sent_at: "sentAt",
+  timestamp: "timestamp",
+  received_at: "receivedAt",
+  original_timestamp: "originalTimestamp",
+  channel: "channel",
+  context_ip: message => coalesce(message, ["context.ip", "request_ip"]),
```
SumanthPuram, line 12: coalesce might not be the right name?
getFirstValidValue() or something that is easy to know the meaning from name.

warehouse/config/helpers.js
```diff
@@ -1,15 +1,21 @@
...
+  return _.isEmpty(_.toString(value));
+};
+
 const coalesce = (message, props) => {
-  return props.reduce(
-    (accumulator, current) =>
-      isNull(accumulator) ? get(message, current) : accumulator,
-    null
-  );
+  for (let index = 0; index < props.length; index += 1) {
```
SumanthPuram, line 13:
```
let currVal = null
for (let index = 0; index < props.length; index += 1) {
  currVal = get(message, props[index]) 
  if (!isBlank(currVal)) {
    break
  }
}

return currVal
```

SumanthPuram: Easier to read than calling get with message and props.

warehouse/index.js
```diff
@@ -122,41 +123,32 @@ function setDataFromColumnMappingAndComputeColumnTypes(
...
+    } else {
+      val = get(input, columnMapping[key]);
+    }
 
-  const setInOutput = (key, val) => {
+    const columnName = utils.safeColumnName(options.provider, key);
     // do not set column if val is null/empty
     if (isBlank(val)) {
+      // eslint-disable-next-line no-param-reassign
+      delete output[columnName];
```
SumanthPuram, line 140: Do we need to delete the output?

SumanthPuram (approved):

SumanthPuram (approved):

PR #363, opened 2020-10-27
Added Unit tests in build spec

arnab-p:


SumanthPuram (commented):
Let's remove unsafe-perm check and also remove the postInstall script?

buildspec.saas.yml
```diff
@@ -14,6 +14,8 @@ phases:
     commands:
       - VERSION="$(date '+%d%m%Y.%H%M%S')"
       - ls
+      - npm install --unsafe-perm
+      - npm test
```
snarkychef, line 18: We are running tests on AWS build container and building the application in Dockerfile on alpine. There can be disparity between the builds. Maybe we should run tests in Dockerfile itself?

@SumanthPuram @sivashanmukh What do you guys think?

sivashanmukh: Yea. Makes sense. We should move it to DockerFile

Dockerfile
```diff
@@ -7,9 +7,9 @@ RUN mkdir -p /home/node/app/node_modules && chown -R node:node /home/node/app
 WORKDIR /home/node/app
 USER node
 
-COPY package*.json ./
-COPY build.js ./
-RUN npm install --unsafe-perm
+ADD . /home/node/app
```
sayan-mitra, line 10: should we run the tests on the image after it's built with docker run?

buildspec.master.yml
```diff
@@ -11,8 +11,7 @@ phases:
     commands:
       - VERSION=master
       - ls
-      - npm install --unsafe-perm
-      - npm test
+      - npm install
```
sayan-mitra, line 14: is this step required if we do npm install in dockerfile?

package.json
```diff
@@ -37,7 +37,7 @@
     "handlebars": "^4.7.6",
     "is": "^3.1.0",
     "is-ip": "^3.1.0",
-    "isolated-vm": "^3.3.3",
+    "isolated-vm": "^2.0.1",
```
sayan-mitra, line 40: should we fix the isolate-vm version?
also if we are not committing package-lock file, should we remove it from the source?

Dockerfile
snarkychef: Sure. Let's do this. Great suggestion

buildspec.saas.yml
SumanthPuram: Definitely a great option if it's not a big effort.

package.json
SumanthPuram: We should commit package-lock.json file unless we have a good reason not to. 
package-lock should automatically take care of this.

sayan-mitra:
looks good 👍

sayan-mitra (approved):

PR #369, opened 2020-10-29
Amplitude batching

arnab-p:


arnab-p:
The code is merged through separate PR. Not required anymore.

PR #386, opened 2020-11-11
Google Analytics revisit

arnab-p:


__tests__/ga.test.js
```diff
@@ -1,34 +1,34 @@
...
-      output = transformer.process(input);
-      expected = expectedData[index];
-    } catch (error) {
-      output = error.message;
-      expected = expectedData[index].message;
-    }
-    expect(output).toEqual(expected);
-  });
+  Object.keys(testData).forEach(key => {
+    test(`${name} Tests : payload: ${key}`, () => {
```
sayan-mitra, line 19: let's name the keys like test-1 or something for easier lookup ?
1/2 etc seems too generic. Going forward we can name them around the specific functionality they are for.

also. add source name to jest test name.

sayan-mitra (commented):
LGTM.
one minor comment.

PR #422, opened 2020-12-22
Hubspot batch implementation

arnab-p:


v0/destinations/hs/transform.js
```diff
@@ -172,6 +175,41 @@ async function processSingleMessage(message, destination) {
...
+      respList.push(formatBatchResponse(message, [metadata], destination));
+    } else {
+      // unprocessed event
+      try {
+        const singleResponse = await processSingleMessage(message, destination);
+        respList.push(
+          formatBatchResponse(singleResponse, [metadata], destination)
+        );
+      } catch (error) {
+        logger.debug(error);
```
SumanthPuram, line 205: Earlier, failed transformations were caught in `proc_errors` and we can notify customers. 

With batch API and failed transformation, we are losing the error message. We are silently dropping the error message here. Not easy to associate this with jobs.

Router is treating this as 500 and retries for 3 hours. So all invalid events will stay in our system for 3 hours.

We missed this while designing batch API earlier. Something we should fix.

SumanthPuram: CC: @chandumlg

SumanthPuram (approved):

arnab-p:
Closing as it is a stale PR

PR #426, opened 2020-12-28
Adds transform for Userlist

benedikt:
## Description of the change

This pull requests adds [Userlist](https://userlist.com) as a new destination. 

## Type of change

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/__tests__/userlist.test.js
+++ b/__tests__/userlist.test.js
@@ -1,0 +1,29 @@
+const integration = "userlist";
+const name = "Userlist";
+
+const fs = require("fs");
+const path = require("path");
+const version = "v0";
+
+const transformer = require(`../${version}/destinations/${integration}/transform`);
+
+const inputDataFile = fs.readFileSync(
+  path.resolve(__dirname, `./data/${integration}_input.json`)
+);
+const outputDataFile = fs.readFileSync(
+  path.resolve(__dirname, `./data/${integration}_output.json`)
+);
+
+const inputData = JSON.parse(inputDataFile);
+const expectedData = JSON.parse(outputDataFile);
+
+inputData.forEach((input, index) => {
+  it(`${name} Tests: payload: ${index}`, async () => {
+    try {
+      const output = await transformer.process(input);
+      expect(output).toEqual(expectedData[index]);
+    } catch (error) {
+      expect(error.message).toEqual(expectedData[index].error);
+    }
+  });
+});
--- a/__tests__/userlist.test.js
+++ b/__tests__/userlist.test.js
@@ -18,9 +18,9 @@ const inputData = JSON.parse(inputDataFile);
 const expectedData = JSON.parse(outputDataFile);
 
 inputData.forEach((input, index) => {
-  it(`${name} Tests: payload: ${index}`, async () => {
+  it(`${name} Tests: payload: ${index}`, () => {
     try {
-      const output = await transformer.process(input);
+      const output = transformer.process(input);
       expect(output).toEqual(expectedData[index]);
     } catch (error) {
       expect(error.message).toEqual(expectedData[index].error);
--- a/v0/destinations/userlist/transform.js
+++ b/v0/destinations/userlist/transform.js
@@ -1,0 +1,52 @@
+const _ = require("lodash");
+
+const { EventType } = require("../../../constants");
+const {
+  defaultRequestConfig,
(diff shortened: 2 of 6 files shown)


v0/destinations/userlist/transform.js
```diff
@@ -0,0 +1,52 @@
...
+  response.body.JSON = message;
+  response.statusCode = 200;
+
+  return response;
+}
+
+function processSingleMessage(message, destination) {
+  const messageType = message.type.toLowerCase();
+
+  if (isBlank(message.userId)) {
```
arnab-p, line 34: We support passing the `userId` in multiple places. Check [here](https://github.com/rudderlabs/rudder-transformer/blob/ca66ab93ea4c9b41b3d7d66b2c5711f9747a2e41/v0/util/data/GenericFieldMapping.json#L41)

Please use the method `getFieldValueFromMessage` from [here](https://github.com/rudderlabs/rudder-transformer/blob/ca66ab93ea4c9b41b3d7d66b2c5711f9747a2e41/v0/util/index.js#L520) to check for a valid value of the `userId`

On a different note: does it make sense to use `anonymousId` instead of `userId` here?

v0/destinations/userlist/transform.js
```diff
@@ -0,0 +1,52 @@
...
+    Authorization: `Push ${pushKey}`
+  };
+  response.body.JSON = message;
+  response.statusCode = 200;
+
+  return response;
+}
+
+function processSingleMessage(message, destination) {
+  const messageType = message.type.toLowerCase();
```
arnab-p, line 32:
```suggestion
  const messageType = message.type && message.type.toLowerCase();
```

v0/destinations/userlist/transform.js
```diff
@@ -0,0 +1,52 @@
...
+
+const { EventType } = require("../../../constants");
+const {
+  defaultRequestConfig,
+  defaultPostRequestConfig
+} = require("../../util");
+
+const { endpoint } = require("./config");
+
+function isBlank(value) {
```
arnab-p, line 11: Please move this method to `v0/util/index.js`.

__tests__/userlist.test.js
```diff
@@ -0,0 +1,29 @@
...
+);
+const outputDataFile = fs.readFileSync(
+  path.resolve(__dirname, `./data/${integration}_output.json`)
+);
+
+const inputData = JSON.parse(inputDataFile);
+const expectedData = JSON.parse(outputDataFile);
+
+inputData.forEach((input, index) => {
+  it(`${name} Tests: payload: ${index}`, async () => {
```
arnab-p, line 21: Please remove the `async/await` as the `transformer.process` method is not `async`.

arnab-p (approved):
LGTM

__tests__/userlist.test.js
benedikt: Ok, removed. FYI: This pattern is common in other tests.

v0/destinations/userlist/transform.js
benedikt: ✅

v0/destinations/userlist/transform.js
benedikt: I'd expect that the existence of required fields (according to the [spec](https://docs.rudderstack.com/rudderstack-api-spec#common-fields)) like this would be checked _before_ the message gets passed to the transforms. Otherwise what's the point of having a spec? 😄 

Anyways, fixed this.

v0/destinations/userlist/transform.js
benedikt: We explicitly don't process messages from unauthenticated users, so this is intentional.

arnab-p: 👍

arnab-p (approved):

v0/destinations/userlist/transform.js
arnab-p: I agree with you. We don't have a top-level validator for the events, and at some times, we saw that some of the events were messed up because of a bad transformation function. This is for fail-safety. We're working on the top level validator. Then we'll clean-up these sections of the code.

__tests__/userlist.test.js
arnab-p: Yeah. We're in the process of the code cleanups for this.

PR #452, opened 2021-01-27
GA - Ecommerce logic update

ruchiramoitra:
## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/v0/destinations/ga/transform.js
+++ b/v0/destinations/ga/transform.js
@@ -732,7 +732,7 @@ function processSingleMessage(message, destination) {
   } = destination.Config;
   enableServerSideIdentify = enableServerSideIdentify || false;
   enhancedEcommerce = enhancedEcommerce || false;
-  ecommerce = ecommerce || true;
+  ecommerce = ecommerce !== undefined ? ecommerce : true;
   switch (messageType) {
     case EventType.IDENTIFY:
       if (enableServerSideIdentify) {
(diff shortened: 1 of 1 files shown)


v0/destinations/ga/transform.js
```diff
@@ -725,9 +725,14 @@ function processSingleMessage(message, destination) {
...
   let category;
-  let { enableServerSideIdentify, enhancedEcommerce } = destination.Config;
+  let {
+    enableServerSideIdentify,
+    enhancedEcommerce,
+    ecommerce
+  } = destination.Config;
   enableServerSideIdentify = enableServerSideIdentify || false;
   enhancedEcommerce = enhancedEcommerce || false;
+  ecommerce = ecommerce || false;
```
sayan-mitra, line 735: let's make this true by default?
then we can support present production users.

ruchiramoitra: yes we should do that.

sayan-mitra: thinking more, actually this is tricky.
making the default true here will lead to always ignoring the config. 
Is the webapp/config BE PR merged? We can make the default value true there (to support present production users)
And check if the config value is undefined in transformer (instead of present falsy check), then only make true (tp support present production users) else take it from config.
wdyt?

ruchiramoitra: I did not understand the 2 line.. 
yes they are merged but there I think the default value is false.

sayan-mitra (approved):

PR #454, opened 2021-01-28
Fix for async function calls inside transformEvent

lokey:


lokey:
_Test for timeout_ tests were not timing out and taking forever. Saw the same behavior before this fix as well. So skipped those tests. But something we should fix.

sivashanmukh (commented):
LGTM. Do add some async test cases to the user-transformations-test

gane5hvarma:
> _Test for timeout_ tests were not timing out and taking forever. Saw the same behavior before this fix as well. So skipped those tests. But something we should fix.

@lokey I have wrote tests for v1 timeout code. Also we will loose event ordering if we use async in `transformEvent` function

util/ivmFactory.js
```diff
@@ -46,10 +46,12 @@ async function createIvm(code, libraryVersionIds) {
           outputEvents = transformBatch(eventMessages, metadata).map(transformedEvent => ({transformedEvent, metadata: metadata(transformedEvent)}))
           break;
         case "transformEvent":
-          outputEvents = eventMessages.map(ev => {
-            let transformedEvent = transformEvent(ev, metadata)
+
+          outputEvents = await Promise.all(eventMessages.map(async ev => {
```
sivashanmukh, line 50: As @gane5hvarma pointed out in an earlier conversation, this will mess up the event ordering if we run transformEvent in an async fashion.

sivashanmukh:
> LGTM. Do add some async test cases to the user-transformations-test

@lokey While at it, do add some test around event ordering as well.

util/ivmFactory.js
lokey: With promise.all, returned values will be in the order of the Promises passed, regardless of the completion order.

lokey: But, if any of the promises is rejected, it rejects the whole array. I will change the logic here.
Also will add more test cases around when some of the promises getting rejected.

gane5hvarma: @lokey so if any of the promise is rejected, should we drop whole batch and raise an error ?

sivashanmukh (approved):
LGTM

gane5hvarma (approved):

gane5hvarma (approved):

PR #462, opened 2021-02-08
fix(ga): Do not fallback to md5(userId) for cid, keep it undefined

lxcid:
## Description of the change

Rudderstack GA transformer default empty `cid` to a md5 hash of user ID, this have cause us problem when unifying session. Due to the debug cycle for session unification, our hypothesis cannot be 100% verified but I believe rudderstack should not even fallback to md5 hash of user ID when anonymousId is not availabe.

I understand that GA spec said the if user ID is not available, cid is needed, that for me should be a concern of the implementor, rudderstack shouldn't make such guess as there are use case for not wanting to provide cid to GA.

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/v0/destinations/ga/transform.js
+++ b/v0/destinations/ga/transform.js
@@ -1,5 +1,4 @@
 const get = require("get-value");
-const md5 = require("md5");
 const { EventType } = require("../../../constants");
 const {
   Event,
@@ -218,7 +217,7 @@ function responseBuilderSimple(
     integrationsClientId ||
     getDestinationExternalID(message, "gaExternalId") ||
     message.anonymousId ||
-    md5(message.userId);
+    undefined;
 
   finalPayload.uip = getParsedIP(message);
 


lxcid:
just want to provide another argument for why we shouldn’t fallback to a a md5 hash of userId if anonymousId is not provided, it will cause a double count of session because that is treated as a separate session instead of being ignore by ga in session view.

this will cause inaccurate count of number of active session.

SumanthPuram:
Should we make this a configuration variable (whether to use md5 or just leave it undefined)?

lxcid:
configurable is good. :)

arnab-p:
@lxcid We're handling your use case here: https://github.com/rudderlabs/rudder-transformer/pull/522

lxcid:
@arnab-p its a slightly different implementation as it default to userId, it should default to undefined and not provide any id, this in some system create a new session, thus double session issue.

PR #464, opened 2021-02-09
Marketo Caching

arnab-p:


v0/destinations/marketo/transform.js
```diff
@@ -9,6 +9,11 @@ const {
   getFieldValueFromMessage
 } = require("../../util");
 const { getAxiosResponse, postAxiosResponse } = require("../../util/network");
+const Cache = require("../../util/cache");
+
+const userIdLeadCache = new Cache(24 * 60 * 60); // 1 day
+const emailLeadCache = new Cache(24 * 60 * 60); // 1 day
+const authCache = new Cache(60 * 60); // 1 hr
```
SumanthPuram, line 16: We want to configure with these while testing. Can you set these from env variables?

SumanthPuram (approved):

v0/destinations/marketo/transform.js
arnab-p: 👍

PR #488, opened 2021-02-23
Added support for incremental update for Salesforce

arnab-p:


v0/destinations/salesforce/transform.js
```diff
@@ -62,7 +62,8 @@ function responseBuilderSimple(traits, salesforceMap, authorizationData) {
   // get traits from the message
   let rawPayload = traits;
```
sayan-mitra, line 63: for update scenario, how are we handling new custom columns or updating existing columns that we have pushed with _c earlier?
we need to append _c for this scenario as well?

sayan-mitra (approved):

PR #505, opened 2021-03-09
marketo lead lookup fix

arnab-p:


v0/destinations/marketo/transform.js
```diff
@@ -181,14 +182,18 @@ const processIdentify = async (
     }
   }
 
+  leadId = null;
```
ryanmccrary, line 185: Will this not always throw an error in the block below?

chandumlg (approved):

PR #552, opened 2021-04-08
Convert userId to string before sending

Shivam4415:
## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/v0/destinations/posthog/transform.js
+++ b/v0/destinations/posthog/transform.js
@@ -79,7 +79,12 @@ const responseBuilderSimple = (message, category, destination) => {
     payload.distinct_id = payload.distinct_id.toString();
   }
   if (payload.properties) {
-    if (!(payload.distinct_id === null || payload.distinct_id === undefined)) {
+    if (
+      !(
+        payload.properties.distinct_id === null ||
+        payload.properties.distinct_id === undefined
+      )
+    ) {
       payload.properties.distinct_id = payload.properties.distinct_id.toString();
     }
   }
--- a/v0/destinations/posthog/transform.js
+++ b/v0/destinations/posthog/transform.js
@@ -75,11 +75,13 @@ const responseBuilderSimple = (message, category, destination) => {
   };
 
   // Convert the distinct_id to string as that is the needed type in destinations.
-  if (payload.distinct_id) {
+  if (!(payload.distinct_id === null || payload.distinct_id === undefined)) {
     payload.distinct_id = payload.distinct_id.toString();
   }
-  if (payload.properties.distinct_id) {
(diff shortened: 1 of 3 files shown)


v0/destinations/posthog/transform.js
```diff
@@ -74,6 +74,14 @@ const responseBuilderSimple = (message, category, destination) => {
     ...payload.properties
   };
 
+  // Convert the distinct_id to string as that is the needed type in destinations.
+  if (payload.distinct_id) {
```
arnab-p, line 78: This will fail for falsy values. Let's check for not null and not undefined.

arnab-p (changes requested):

v0/destinations/posthog/transform.js
Shivam4415: Javascript supports 7 falsy values :: 
1. the number 0, -0
2. the BigInt 0n
3. the keyword null
4. the keyword undefined
5. the boolean false
6. the number NaN
7. the empty string "" (equivalent to '' or ``)
I tested for all of them and it never fails.
Should I still change it the way u are suggesting.

arnab-p: It's improbable, but if `payload.distinct_id` is `0`, it won't be converted to String. Can we make sure that we're handling all the cases?

Shivam4415: Actually what I was thinking is  if payload.distinct_id is 0, which should never happen actually but will happen only if wrong payload is sent or sdk failed to generate correct anonymousId/userId. Keeping that in my mind I thought why I should handle logic for wrong value. Anyways I will add logic as u suggested and commit.

v0/destinations/posthog/transform.js
```diff
@@ -74,6 +74,21 @@ const responseBuilderSimple = (message, category, destination) => {
     ...payload.properties
   };
 
+  // Convert the distinct_id to string as that is the needed type in destinations.
+  if (!(payload.distinct_id === null || payload.distinct_id === undefined)) {
```
arnab-p, line 78: Let's use the method `isDefinedAndNotNull` from https://github.com/rudderlabs/rudder-transformer/blob/738121e00997ebc438cc41318781f8a9b5ab53fb/v0/util/index.js#L30

arnab-p (changes requested):

v0/destinations/posthog/transform.js
Shivam4415: fixed as suggested

arnab-p (approved):

PR #556, opened 2021-04-12
Clevertap enhancement

arajguha:
## Type of change
- [x] Fixed usage of removeUndefinedAndNullValues function
- [x] Added functionality to map list of provided values in metadata to a target value.

### DIFF
--- a/v0/util/index.js
+++ b/v0/util/index.js
@@ -512,23 +512,28 @@ const handleMetadataForValue = (value, metadata) => {
   //   }
   // ]
   if (multikeyMap) {
+    let foundVal = false;
     if (Array.isArray(multikeyMap)) {
       multikeyMap.some(map => {
         if (!map.sourceVal || !map.destVal || !Array.isArray(map.sourceVal)) {
           logger.warn(
             "multikeyMap skipped: sourceVal and destVal must be of valid type"
           );
+          foundVal = true;
           return true;
         }
 
         if (map.sourceVal.includes(formattedVal)) {
           formattedVal = map.destVal;
+          foundVal = true;
           return true;
         }
       });
     } else {
       logger.warn("multikeyMap skipped: multikeyMap must be an array");
     }
+    if (!foundVal)
+      formattedVal = undefined
   }
 
   return formattedVal;
--- a/v0/util/index.js
+++ b/v0/util/index.js
@@ -521,8 +521,7 @@ const handleMetadataForValue = (value, metadata) => {
           return true;
         }
 
-        const keySet = new Set(map.sourceVal);
-        if (keySet.has(formattedVal)) {
+        if (map.sourceVal.includes(formattedVal)) {
           formattedVal = map.destVal;
           return true;
         }
--- a/v0/util/index.js
+++ b/v0/util/index.js
@@ -399,7 +399,14 @@ const handleMetadataForValue = (value, metadata) => {
   }
 
   // get infor from metadata
-  const { type, typeFormat, template, defaultValue, excludes } = metadata;
+  const {
+    type,
+    typeFormat,
+    template,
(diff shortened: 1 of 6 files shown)


v0/util/index.js
```diff
@@ -488,6 +495,42 @@ const handleMetadataForValue = (value, metadata) => {
...
+          formattedVal = map.destVal;
+          return true;
+        }
+      });
+    } else {
+      logger.warn("multikeyMap skipped: multikeyMap must be an array");
+    }
+  }
+
   return formattedVal;
```
ruchiramoitra, line 534: If formattedValue is neither F or M now we should put formattedVal as undefined iff required is false. Otherwise we return formattedVal

v0/util/index.js
```diff
@@ -488,6 +495,42 @@ const handleMetadataForValue = (value, metadata) => {
...
+    if (Array.isArray(multikeyMap)) {
+      multikeyMap.some(map => {
+        if (!map.sourceVal || !map.destVal || !Array.isArray(map.sourceVal)) {
+          logger.warn(
+            "multikeyMap skipped: sourceVal and destVal must be of valid type"
+          );
+          return true;
+        }
+
+        if (map.sourceVal.includes(formattedVal)) {
```
ruchiramoitra, line 524: can we 
 var upperCaseSourceValMap=  map.sourceVal.map(a => {
          return a.toUpperCase();
        });
  
        if (upperCaseSourceValMap.includes(formattedVal.toUpperCase())) {
          formattedVal = map.destVal;
          return true;
        }

So that even if "FEMALE" comes it is handled. and from sourceVal we can only set "m" "f" and "M" "F" is not required? WDYT?

utsabc: Yes, we can do this we can reduce case-specific keys. Either we can convert all keys to upper or lower-case and then compare the formattedVal in that case, we should mention this in the comment docs for ease of understanding Even though code is self-explanatory. We should do this @arajguha

arajguha: W.r.t I am getting a few things in mind. 
`
var upperCaseSourceValMap= map.sourceVal.map(a => {
return a.toUpperCase();
});
`
This map function to convert to upper or lower case would iterate through **all** the values and return the new array and on that new array we are looking for the match. So, in terms of iteration , wouldn't just one linear search be better ?

Second thing is, in this particular case, it happens to be an array of strings. For non-string values, this would crash. Would it better to generalise it?
What do you think ?

v0/util/index.js
arajguha: Looking into this. **required** is not in the scope of **handleMetadataForValue** function. So have to think about a work around.

v0/util/index.js
utsabc: I agree, for other `data type` and considering the time to iterate and create a new map of `uppercase values`. We should just accept an array of keys. It should be up to the integration developer how he wants to match. @ruchiramoitra  WDYT seems like a valid point?

ruchiramoitra: yes.. let's leave this. in the data json file add "FEMALE" and "MALE"

v0/util/index.js
ruchiramoitra: yes no need to check required. Just assign undefined. If it is required and still undefined is going it will throw error.

v0/util/index.js
arajguha: okay. got it.

v0/util/index.js
arajguha: ok got it

ruchiramoitra (approved):

ruchiramoitra (approved):

utsabc (approved):

arnab-p (approved):

PR #597, opened 2021-05-10
Added maskRules to webhook destination

arnab-p:


arnab-p:
Redundant PR as we decided to do it in a different path

PR #649, opened 2021-06-16
[hotfix] Webhook dynamic header support

arnab-p:


ryanmccrary (approved):
Looks good to me. This should do the trick

PR #810, opened 2021-10-16
Response transform

utsabc:
Implementation of Response transform. The idea for this is to transform any non compatible response from destination for server.

### DIFF
--- a/v0/util/index.js
+++ b/v0/util/index.js
@@ -1083,7 +1083,7 @@ function populateErrStat(error, destination, isStageTransform = true) {
       scope: TRANSFORMER_METRIC.MEASUREMENT_TYPE.EXCEPTION.SCOPE
     };
     // eslint-disable-next-line no-ex-assign
-    error = { ...error, statTags };
+    error.statTags = statTags;
   }
   return error;
 }


adapters/network.js
```diff
@@ -4,75 +4,144 @@ const http = require("http");
...
-const send = async options => {
-  // here the options argument K-Vs will take priority over requestOptions
-  const requestOptions = {
-    ...{
-      // `method` is the request method to be used when making the request
-      method: "get",
+const agentArgs = removeUndefinedAndNullValues({
+  keepAlive: true,
+  maxSockets: Number.isNaN(
+    Number.parseInt(process.env.NETWORK_AGENT_MAX_SOCKET, 10)
```
sanpj2292, line 13: Should we still have `maxSockets`/`maxFreeSockets`/`keepAlive` to be set ?

v0/util/error.js
```diff
@@ -0,0 +1,64 @@
+function ErrorBuilder() {
```
sanpj2292, line 1: https://github.com/rudderlabs/rudder-transformer/blob/2aa28b7785858dcfe64c314f718eac8566050868/v0/util/response-builders.js

Why can't we write something like this ? Just wanted to know the reason you've implemented here

versionedRouter.js
```diff
@@ -587,12 +609,12 @@ router.post("/batch", ctx => {
...
 });
 
-router.get("/heapdump", ctx => {
-  heapdump.writeSnapshot((err, filename) => {
-    logger.debug("Heap dump written to", filename);
-  });
-  ctx.body = "OK";
-});
+// router.get("/heapdump", ctx => {
+//   heapdump.writeSnapshot((err, filename) => {
```
sanpj2292, line 613: This shouldn't be commented right ?!

versionedRouter.js
```diff
@@ -514,24 +517,43 @@ if (startSourceTransformer) {
...
 
-async function handleDestinationNetwork(version, ctx) {
-  const { destination } = ctx.request.body;
+function handleResponseTransform(version, destination, ctx) {
   const destNetHandler = getDestNetHander(version, destination);
   // flow should never reach the below (if) its a desperate fall-back
-  if (!destNetHandler || !destNetHandler.sendData) {
+  if (!destNetHandler || !destNetHandler.responseTransform) {
     ctx.status = 404;
     ctx.body = `${destination} doesn't support transformer proxy`;
```
sanpj2292, line 525: Can we change `transformer proxy` to `response transformation` ?

versionedRouter.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable no-prototype-builtins */
 /* eslint-disable import/no-dynamic-require */
 /* eslint-disable global-require */
-const heapdump = require("heapdump");
+// const heapdump = require("heapdump");
```
sanpj2292, line 4: This also shouldn't be commented I guess, can we check it ?

adapters/network.js
utsabc: Since we moved this config via transformer env vars and we are using this function for transformer api calls (example Marketo bulk upload) so kept it for that use case

v0/util/error.js
utsabc: Yeah since I was using the error part only I kept it here but yeah we can write like this

versionedRouter.js
utsabc: We are not using this and every time we debug we have to commented this hence I have commented this. We can remove this actually

versionedRouter.js
utsabc: Yep

v0/destinations/marketo/networkResponseHandler.js
```diff
@@ -0,0 +1,251 @@
...
+  } catch (err) {
+    clientResponse = { success: false, response: err };
+  }
+  return clientResponse;
+};
+
+const responseTransform = destResponse => {
+  let respBody;
+  try {
+    respBody = JSON.parse(destResponse.Body);
```
sanpj2292, line 168: I think we should use a lowercase starting character for properties names in an Object. Can we follow that convention here ?

v0/destinations/marketo/networkResponseHandler.js
```diff
@@ -0,0 +1,251 @@
...
+  } catch (err) {
+    respBody = isEmpty(!destResponse.Body) ? destResponse.Body : null;
+  }
+  if (respBody && !respBody.success) {
```
sanpj2292, line 172: How do we know if the response is a success or failure here ? Where is this success property being populated ?

adapters/networkhandler/genericNetworkResponseHandler.js
```diff
@@ -0,0 +1,35 @@
+const { isEmpty } = require("../../v0/util/index");
+/**
+ * network handler as a fall back for all destination nethandlers, this file provides abstraction
+ * for all the network comms btw dest transformer along with dest specific reqeusts from server to actual APIs
```
sanpj2292, line 4: > dest specific requests

What does it mean in this context ?

v0/destinations/marketo/networkResponseHandler.js
utsabc: We are actually getting destResponse as the string representation of a rudder-server structure where we follow the convention of having upper-case properties

v0/destinations/marketo/networkResponseHandler.js
utsabc: respBody is the destination response for Marketo which will contain the success key in it specifying if the request is successful or failure. As it returns 200 for every request we need to check the success/failure scenario using this key

adapters/networkhandler/genericNetworkResponseHandler.js
utsabc: Fixed it

adapters/network.js
sanpj2292: I would say, let's clear this out now & add it at a later point of time, if there is a requirement as such

adapters/networkhandler/genericNetworkResponseHandler.js
```diff
@@ -0,0 +1,35 @@
+const { isEmpty } = require("../../v0/util/index");
+/**
+ * network handler as a fall back for all destination nethandlers, this file provides abstraction
+ * for all the network comms btw dest transformer.
+ *
+ * --responseTransform-- this is a function which can be used to handle responses which are not-compatible with
```
sanpj2292, line 6: We probably would need to mention what `not compatible with rudder-server` means. Take an example of a destination like BQStream/Marketo wherein the status codes are not rudder-server friendly(I mean rudder-server by itself cannot conclude if the error is retryable/abortable/throttable). Let's add it a bit later

`Adding the comment here as a note`

v0/destinations/marketo/networkResponseHandler.js
sanpj2292: Got it

v0/destinations/marketo/networkResponseHandler.js
sanpj2292: Why don't we do a `json.Marshal(destinationResponse)` & send it to Transformer ?
Also for the type of destinationResponse we can define the json while defining the struct itself
For example

```golang

type DestinationResponseT struct {
 Body string `json:"body"`
 .
 .
 .
}
```

v0/destinations/marketo/networkResponseHandler.js
```diff
@@ -0,0 +1,251 @@
...
+    // marketo application response level failure
```
sanpj2292, line 173: Shall we move this to a new method ?

v0/destinations/marketo/transform.js
```diff
@@ -93,24 +104,27 @@ const createOrUpdateLead = async (
...
-        action: "createOrUpdate",
-        input: [attribute],
-        lookupField: userId ? "userId" : "anonymousId"
-      },
-      {
-        headers: {
-          Authorization: `Bearer ${token}`,
-          "Content-type": "application/json"
+    const response = marketoResponseHandler({
+      clientResponse: await sendPostRequest(
```
sanpj2292, line 108: We are not using `httpPOST` here ?

v0/destinations/marketo/transform.js
```diff
@@ -127,17 +141,20 @@ const createOrUpdateLead = async (
...
-      {
-        params: { filterValues: email, filterType: "email" },
-        headers: { Authorization: `Bearer ${token}` }
-      },
-      "During lead look up using email"
-    );
-    if (resp) {
-      const { result } = resp;
+    const response = marketoResponseHandler({
+      clientResponse: await sendGetRequest(
```
sanpj2292, line 145: Similarly we are not using `httpGET` method here ?

adapters/networkhandler/genericNetworkResponseHandler.js
utsabc: yep

v0/destinations/marketo/transform.js
utsabc: sendPostRequest is a wrapper on top of httpPost for Marketo.  It returns response which is compatible with reesponsehandler

v0/destinations/marketo/networkResponseHandler.js
utsabc: This is handled

v0/destinations/marketo/networkResponseHandler.js
sanpj2292: Ok

v0/destinations/marketo/transform.js
utsabc: Just like post to we are using a wrapper on top of httpGET for marketo which is compatible with responseHandler

sanpj2292 (dismissed):

sanpj2292 (approved):

PR #858, opened 2021-11-18
Bug: Included check for payload.user and userEmail for Zendesk

abandonned:
## Description of the change
Included check for payload.user and throwing error if userEmail is not found.

> Notion -> https://www.notion.so/rudderstacks/Zendesk-Aellocred-6b12bb67dafd484e8b4188feaa6685f5

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/__tests__/data/zendesk_input.json
+++ b/__tests__/data/zendesk_input.json
@@ -43,8 +43,7 @@
         "screen": {
           "density": 2
         },
-        "traits": {
-        },
+        "traits": {},
         "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.149 Safari/537.36"
       },
       "integrations": {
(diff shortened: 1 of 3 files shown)


__tests__/data/zendesk_input.json
```diff
@@ -1,4 +1,65 @@
...
+        "locale": "en-GB",
+        "os": {
+          "name": "",
+          "version": ""
+        },
+        "screen": {
+          "density": 2
+        },
+        "traits": {
+        },
```
saikumarrs, line 47: For track events, there is also an edge case where a missing `name` field in `traits` can cause an exception (while trying to create a new user). So, can we modify this test case input exactly to cover the missing `email` case?

abandonned: this test case is for the first error, where if none of the traits in identify mapping is present user is not created inside payload.
We can add test case to show the 2nd error as well.

saikumarrs: For the first error, just the absence of `email` is enough. I believe this data is good enough for the second error but might need more work to cover that.

abandonned: They both are of different eventType (track and identify) therefore included a new test case to demonstrate email misssing in traits error.

saikumarrs: I was referring to the exception thrown in `createUser` method.

abandonned: Sorry. `createUser` method is being in two different places. Inside `processTrack` and in `processGroup`. 

- For `processTrack` if we don't have `userEmail` we are throwing error so it won't reach `createUser` without `userEmail` . And we don't want to make call without a valid `userEmail` .
- For `processGroup` there is already a check present for `traits.name` and `traits.email`

saikumarrs: Agreed. In order to cover the first error, the input should just not contain `email` field and for the second one (in `createUser`), at least the `name` field should be missing in the `traits` object.
We should have test cases to explicitly cover each of them.

saikumarrs (approved):

PR #863, opened 2021-11-22
HS - feat: optimize "properties" endpoint

akashrpo:
## Description of the change

> Optimize the call for "properties" endpoint to avoid reaching its rate-limit.

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/v0/destinations/hs/transform.js
+++ b/v0/destinations/hs/transform.js
@@ -335,7 +335,20 @@ const processRouterDest = async inputs => {
   // using the first destination config for transforming the batch
   const { destination } = inputs[0];
   // reduce the no. of calls for properties endpoint
-  const propertyMap = await getProperties(destination);
+  let propertyMap;
+  const traitsFound = inputs.some(input => {
+    if (
+      input.message.traits ||
+      input.message.context.traits ||
+      input.message.properties
+    ) {
+      return true;
+    }
+    return false;
+  });
+  if (traitsFound) {
+    propertyMap = await getProperties(destination);
+  }
   await Promise.all(
     inputs.map(async input => {
       try {
--- a/v0/destinations/hs/transform.js
+++ b/v0/destinations/hs/transform.js
@@ -343,10 +343,7 @@ const processRouterDest = async inputs => {
   // reduce the no. of calls for properties endpoint
   let propertyMap;
   const traitsFound = inputs.some(input => {
-    if (getTraits(input.message)) {
-      return true;
-    }
-    return false;
+    return getTraits(input.message) !== undefined;
   });
   if (traitsFound) {
     propertyMap = await getProperties(destination);
--- a/v0/destinations/hs/transform.js
+++ b/v0/destinations/hs/transform.js
@@ -66,7 +66,12 @@ async function getProperties(destination) {
   return hubSpotPropertyMap;
 }
 
-async function getTransformedJSON(message, mappingJson, propertyMap) {
+async function getTransformedJSON(
+  message,
+  mappingJson,
+  destination,
+  propertyMap
+) {
   const rawPayload = {};
   const sourceKeys = Object.keys(mappingJson);
   let traits = getFieldValueFromMessage(message, "traits");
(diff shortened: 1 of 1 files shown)


v0/destinations/hs/transform.js
```diff
@@ -324,6 +326,8 @@ const processRouterDest = async inputs => {
   const errorRespList = [];
   // using the first destination config for transforming the batch
   const { destination } = inputs[0];
+  // reduce the no. of calls for properties endpoint
+  const propertyMap = await getProperties(destination);
```
saikumarrs, line 330: Properties are needed if at least one of the events has `traits` defined. So, let's further optimize it.

v0/destinations/hs/transform.js
```diff
@@ -218,8 +217,11 @@ async function processSingleMessage(message, destination) {
...
 }
 
-function process(event) {
-  return processSingleMessage(event.message, event.destination);
+// has been deprecated - using routerTransform for both the versions
+async function process(event) {
+  // no. of calls for properties endpoint remains the same
+  const propertyMap = await getProperties(event.destination);
+  return processSingleMessage(event.message, event.destination, propertyMap);
 }
```
saikumarrs, line 225: Although not really a bad change, I would revert these changes and just send `undefined` in place of `propertyMap` to align with my comment below.

v0/destinations/hs/transform.js
```diff
@@ -76,7 +76,6 @@ async function getTransformedJSON(message, mappingJson, destination) {
 
   if (traits) {
     const traitsKeys = Object.keys(traits);
-    const propertyMap = await getProperties(destination);
```
saikumarrs, line 79: I would still keep the `destination` argument and calculate `propertyMap` if the incoming value is not defined.

akashrpo: added the check

v0/destinations/hs/transform.js
akashrpo: sent undefined. The code looks cleaner this way.

v0/destinations/hs/transform.js
akashrpo: added the `traits` check

v0/destinations/hs/transform.js
```diff
@@ -324,6 +334,21 @@ const processRouterDest = async inputs => {
...
+  let propertyMap;
+  const traitsFound = inputs.some(input => {
+    if (
+      input.message.traits ||
+      input.message.context.traits ||
+      input.message.properties
+    ) {
+      return true;
+    }
+    return false;
```
saikumarrs, line 347: The logic above is not properly replicated here. Let's refactor that original logic into a method and use it in two places.

akashrpo: The above original logic extracts the traits and here I think we just need to add the check for the presence of traits.

saikumarrs: I can think of two reasons here for the re-factoring:
- Extracting traits (in general) is not a costly operation.
- Having this logic in two places is a potential risk going forward.

akashrpo: refactored the logic into a method and used it in two different places.

v0/destinations/hs/transform.js
```diff
@@ -324,6 +340,17 @@ const processRouterDest = async inputs => {
...
   // using the first destination config for transforming the batch
   const { destination } = inputs[0];
+  // reduce the no. of calls for properties endpoint
+  let propertyMap;
+  const traitsFound = inputs.some(input => {
+    if (getTraits(input.message)) {
+      return true;
+    }
+    return false;
+  });
```
saikumarrs, line 350: Let's simplify this logic a bit,

```suggestion
  const traitsFound = inputs.some(input => {
    getTraits(input.message) !== undefined
    }
  });
```

akashrpo: simplified

saikumarrs (approved):

PR #904, opened 2021-12-14
Feat: Klaviyo ecommerce Specs

abandonned:
## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


v0/destinations/klaviyo/transform.js
```diff
@@ -2,13 +2,19 @@
 /* eslint-disable no-underscore-dangle */
 /* eslint-disable  array-callback-return */
 const get = require("get-value");
+const fs = require("fs");
+const path = require("path");
```
akashrpo, line 6: Let's remove these unused imports

v0/destinations/klaviyo/transform.js
```diff
@@ -160,13 +163,75 @@ const trackRequestHandler = (message, category, destination) => {
...
-    customerProperties._id = getFieldValueFromMessage(message, "userId");
-  }
-  payload.customer_properties = customerProperties;
-  if (message.timestamp) {
-    payload.time = toUnixTimestamp(message.timestamp);
+  return customerProperties;
+};
+const trackRequestHandler = (message, category, destination) => {
+  let payload = {};
+  let event = getValueFromMessage(message, "event");
```
akashrpo, line 170: Let's use simple get() call to extract `event`

akashrpo:
Let's add required test cases for new ecomm specs so we don't miss out any edge cases during merge.

v0/destinations/klaviyo/transform.js
abandonned: Everywhere in transform.js getValueFromMessage is being used , so let's maintain parity if there is no advantage in changing that.

akashrpo: Actually we are using only `get()` and `getFieldValueFromMessage()` in transform.js so directly calling `get()` here would be better.

v0/destinations/klaviyo/config.js
```diff
@@ -9,9 +9,49 @@ const CONFIG_CATEGORIES = {
   IDENTIFY: { name: "KlaviyoIdentify", apiUrl: "/api/identify" },
   SCREEN: { name: "KlaviyoTrack", apiUrl: "/api/track" },
   TRACK: { name: "KlaviyoTrack", apiUrl: "/api/track" },
-  GROUP: { name: "KlaviyoGroup" }
+  GROUP: { name: "KlaviyoGroup" },
+  STARTED_CHECKOUT: { type: "STARTED_CHECKOUT", name: "StartedCheckout" },
+  VIEWED_PRODUCT: { type: "VIEWED_PRODUCT", name: "ViewedProduct" },
+  ADDED_TO_CART: { type: "ADDED_TO_CART", name: "AddedToCart" },
+  ITEMS: { type: "ITEMS", name: "Items" }
```
saikumarrs, line 16: Just curious on why `type` was added to only the new entries

v0/destinations/klaviyo/transform.js
saikumarrs: If we need to extract value from a key that does not have multiple sources (say `userId` --- `userId`, `traits.userId`, `traits.id` etc.), then `get` is ideal, as in this case. Otherwise, `getFieldValueFromMessage` should be used.

v0/destinations/klaviyo/transform.js
```diff
@@ -160,13 +161,75 @@ const trackRequestHandler = (message, category, destination) => {
...
+    }
+    const customerProperties = createCustomerProperties(message);
+    if (destination.Config.enforceEmailAsPrimary) {
+      delete customerProperties.$id;
+      customerProperties._id = getFieldValueFromMessage(message, "userId");
+    }
+    payload.customer_properties = customerProperties;
+    if (message.timestamp) {
+      payload.time = toUnixTimestamp(message.timestamp);
+    }
```
saikumarrs, line 232: The entire logic is cramped here with no visual separation or code documentation. Please improve this.

v0/destinations/klaviyo/transform.js
```diff
@@ -160,13 +160,75 @@ const trackRequestHandler = (message, category, destination) => {
...
+    const eventName = eventNameMapping[event];
+    payload.event = eventName;
+    payload.token = destination.Config.publicApiKey;
+    const eventMap = jsonNameMapping[eventName];
+    payload.customer_properties = createCustomerProperties(message);
+    if (
+      !payload.customer_properties.$email &&
+      !payload.customer_properties.$phone_number
+    ) {
+      throw new CustomError("customer_properties not found", 400);
```
saikumarrs, line 181: The error is not appropriate here. Let's make it a bit more specific

v0/destinations/klaviyo/transform.js
```diff
@@ -160,13 +160,75 @@ const trackRequestHandler = (message, category, destination) => {
...
+      delete payload.properties.revenue;
+    }
+    const customerProperties = createCustomerProperties(message);
+    if (destination.Config.enforceEmailAsPrimary) {
+      delete customerProperties.$id;
+      customerProperties._id = getFieldValueFromMessage(message, "userId");
+    }
+    payload.customer_properties = customerProperties;
+    if (message.timestamp) {
+      payload.time = toUnixTimestamp(message.timestamp);
```
saikumarrs, line 236: Why are these not needed for e-com events?

abandonned: These are pre existing track call. didn't change that, and hence its inside else block

saikumarrs: So, not applicable for e-com events?

saikumarrs: Actually, there is a good amount of common logic b/w two categories. I would unify the logic and put the event type-specific code after that (as already there are `if` conditions to perform specific actions)

abandonned: its not related to ecom events

v0/destinations/klaviyo/transform.js
abandonned: have added comments and separated different things with linebreak

v0/destinations/klaviyo/transform.js
abandonned: Have updated the error code to provide more information

v0/destinations/klaviyo/transform.js
```diff
@@ -160,11 +160,85 @@ const trackRequestHandler = (message, category, destination) => {
...
+    const customerProperties = createCustomerProperties(message);
+    if (destination.Config.enforceEmailAsPrimary) {
+      delete customerProperties.$id;
+      customerProperties._id = getFieldValueFromMessage(message, "userId");
+    }
+    payload.customer_properties = customerProperties;
   }
-  payload.customer_properties = customerProperties;
   if (message.timestamp) {
     payload.time = toUnixTimestamp(message.timestamp);
```
akashrpo, line 243: Let's use `getFieldValueFromMessage()` to extract timestamp and also insert a check whether timestamp is already in unix format otherwise this function will truncate the last 3 digits and contain invalid data.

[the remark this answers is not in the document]
abandonned: this part of the code is already in production.  @saikumarrs should we update it or keep the original one?

akashrpo (approved):

v0/destinations/klaviyo/transform.js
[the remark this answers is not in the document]
saikumarrs: I wouldn't go for this update as `timestamp` and `originalTimestamp` by definition are not same. 
Refer to https://rudderstack.com/docs/data-warehouse-integrations/warehouse-schemas/#standard-rudderstack-properties.

For the second concern, as both the timestamp fields are calculated/generated by RudderStack and wouldn't be in a different format than expected.

saikumarrs (approved):

PR #911, opened 2021-12-21
Refactor network handler

arajguha:
## Description of the change

> Description here
The main change is related to moving error builder and network handler factory implementations to a class-based approach.
This is more like a suggestive approach, but maybe worth looking into.

Motivation:
- remove redundancy caused mainly by the networkHandler function and enforce structure.
- easier onboarding for new destinations.  For new destinations, the following steps can be followed:
   - create Destination specific network handler class by extending the GenericNetworkHandler
   - override the responseHandler method in that class 
   - add the class to the map in networkHandlersMap.js

Another important point to note:
`networkHandlerFactory.getNetworkHandler(
    destination
)`
This seems to create a new instance for every incoming request.
With this refactor, we should have the following:
- one single networkHandlerFactory instance in versionedRouter
- that instance would  contain a map to store destination specific handler instances. The map initially would contain only the genericHandler instance
- as new requests start coming, based on the destination it will create a handler instance and add it to the map.
- so now, each destination would have a single handler instance that would be reused.

Verified that single network handler instance is created at the start for each worker and being re used for all incoming requests. ✅ 

**Need to loadtest**

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


adapters/networkHandlerFactory.js
```diff
@@ -1,26 +1,38 @@
...
   }
-};
+}
+
+// Alternate option 1: expose the class and make the member method as static
+// Alternate option 2: export the class and initialize the
+// singleton object inside versioned router
+
+// directly exporting a singleton factory object
+module.exports = new NetworkHandlerFactory();
```
utsabc, line 38: If node modules are cached we will only have 1 instance of `NetworkHadlerFactory` after initialization of the server. Need to verify by generating some member property like objectId it - and printing it in the versioned router. WDYT?

utsabc (approved):

adapters/networkHandlerFactory.js
arajguha: yes that should help in verifying that only one instance is being created.

adapters/networkHandlerFactory.js
```diff
@@ -1,26 +1,38 @@
...
-  networkHandler: BqStreamNetworkHandler
-} = require("../v0/destinations/bqstream/util");
+const NetWorkHandlerClassMap = require("./networkHandlersMap");
+
+class NetworkHandlerFactory {
+  constructor() {
+    this.info = "NetworkHandlerFactory";
+    // intially map would contain only the generic handler
+    this.factoryMap = new Map();
+    this.factoryMap.set("generic", new GenericNetworkHandler());
```
saikumarrs, line 11: We should avoid creating generic handler by default (make it on-demand)

adapters/networkHandlersMap.js
```diff
@@ -0,0 +1,9 @@
+const { BqStreamNetworkHandler } = require("../v0/destinations/bqstream/util");
+const { BrazeNetworkHandler } = require("../v0/destinations/braze/util");
+const { MarketoNetworkHandler } = require("../v0/destinations/marketo/util");
+
+module.exports = {
+  braze: BrazeNetworkHandler,
+  marketo: MarketoNetworkHandler,
+  bqstream: BqStreamNetworkHandler
+};
```
saikumarrs, line 9: Instead of a map, how about dynamically `require`ing the module directly in the factory? Additional, overhead for the developer to add an entry here.

v0/destinations/marketo/util.js
```diff
@@ -126,50 +125,46 @@ const sendPostRequest = async (url, data, options) => {
...
+        .setStatTags({
+          destination: DESTINATION,
+          stage: TRANSFORMER_METRIC.TRANSFORMER_STAGE.RESPONSE_TRANSFORM,
+          scope: TRANSFORMER_METRIC.MEASUREMENT_TYPE.API.SCOPE,
+          meta: getDynamicMeta(status)
+        })
+        .build();
+    }
+    // check for marketo application level failures
+    marketoApplicationErrorHandler(
```
saikumarrs, line 151: Why are we naming methods against the destination? Also, we are exporting it separately. Shouldn't the exported class take care of that?

v0/destinations/marketo/util.js
```diff
@@ -126,50 +125,46 @@ const sendPostRequest = async (url, data, options) => {
...
-      .setStatTags({
-        destination: DESTINATION,
-        stage: TRANSFORMER_METRIC.TRANSFORMER_STAGE.RESPONSE_TRANSFORM,
-        scope: TRANSFORMER_METRIC.MEASUREMENT_TYPE.API.SCOPE,
-        meta: getDynamicMeta(status)
-      })
-      .build();
+class MarketoNetworkHandler extends GenericNetworkHandler {
+  // eslint-disable-next-line class-methods-use-this
+  responseHandler(destinationResponse) {
```
saikumarrs, line 130: How is this different from the responseHandler defined in Ln74?

arajguha: I can see that those are being used for the intermediate axios calls. @utsabc can you share some more insight on this?

adapters/networkHandlerFactory.js
arajguha: IMHO the map should at least start off with the generic handler. Anyway that is the main one that would be used. Only for few specific destinations we are going to have a specific/custom handler

adapters/networkHandlersMap.js
arajguha: dynamically requiring wrt to the `destination` name? I think it can be done. The idea here was to maintain a list of the specific handlers in one place.
Another reason why I was trying to avoid dynamic `require` is this --> https://eslint.org/docs/rules/global-require
We currently do it in `versionedRouter` but it really is an anti-pattern, It seems.
WDYT?

v0/destinations/marketo/util.js
arajguha: yes we should maybe refactor it further. The problem of keeping those specific methods in class is that, then an instance needs to be created in `transform.js` to use that. We can think of `marketoApplicationErrorHandler` as more of a utility function being used in transform.js.  
I am not sure if we should be looking towards creating instances of this class elsewhere other than the factory (at least as of now).
@utsabc  wdyt ? I think @saikumarrs  has a good point.

adapters/networkHandlerFactory.js
saikumarrs: Got it. But why create if we don't know for sure?

v0/destinations/marketo/util.js
saikumarrs: If they are strictly different, we should avoid name conflicts

v0/destinations/marketo/util.js
saikumarrs: IMO, it should be the factory's responsibility to instantiate and cache these objects. On a tangent, should these be converted to static classes (not even singleton), if they don't really hold any "instance" specific data

PR #936, opened 2022-01-07
Bugfix: To support field with mapping in dynamic config

abandonned:
## Description of the change
Dynamic Config was only supporting fields with direct values in {{ }}. We are trying to extend this support to all types of input fields such as for key inside array of objects.

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


util/dynamicConfig.js
```diff
@@ -1,24 +1,38 @@
...
-        if (value.includes("||")) {
-          const path = value.split("||")[0].trim();
-          const getFieldVal = get(event, path);
-          if (getFieldVal) {
-            Config[field] = getFieldVal;
-            unset(event, path);
-          } else {
-            Config[field] = JSON.parse(value.split("||")[1].trim()).toString();
+      const value = Config[field].toString().trim();
+      if (Array.isArray(value)) {
```
sanpj2292, line 25: Is there a possibility that `Config[field]` can be an object in any case ?

util/dynamicConfig.js
```diff
@@ -1,24 +1,38 @@
...
+    } else {
+      Config[field] = JSON.parse(value.split("||")[1].trim()).toString();
+    }
+  }
+}
+
 function getDynamicConfig(event) {
   const { Config } = event.destination;
   if (Config) {
     Object.keys(Config).forEach(field => {
```
sanpj2292, line 23: Just a thought
How about write a new method with same signature using `recursion`, which makes things easier to read

util/dynamicConfig.js
```diff
@@ -1,24 +1,38 @@
+/* eslint-disable no-param-reassign */
 const get = require("get-value");
 const unset = require("unset-value");
 
+function recurse(event, value, Config, field) {
```
sanpj2292, line 5:
```suggestion
function setDynamicConfigValue(event, value, config, field) {
```
This name looks more appropriate rather than `recurse` & also don't use first letter upper-cased character in method args

util/dynamicConfig.js
abandonned: No I don't think we can have that case. @ruchiramoitra

util/dynamicConfig.js
abandonned: Sorry didn't get this. What do we want to include in new method?

sanpj2292: Leave it, it seems unnecessary now

util/dynamicConfig.js
abandonned: Updated.

util/dynamicConfig.js
```diff
@@ -1,29 +1,48 @@
+/* eslint-disable no-param-reassign */
 const get = require("get-value");
 const unset = require("unset-value");
 
+function setDynamicConfigValue(event, value, config, field) {
+  value = value.replace("{{", "");
+  value = value.replace("}}", "");
```
saikumarrs, line 7:
```suggestion
  value = value.replace("{{", "").replace("}}", "");
```

abandonned: updated

util/dynamicConfig.js
ruchiramoitra: you are checking only if it is not string. this would not work if the value is a boolean. Check the other scenarios too.

ruchiramoitra: like normal string not matching the patter that value is passed. boolean should also go as it is

sanpj2292:
Can we also add separate `test-cases` for dynamic config alone to check out all the scenarios ?

util/dynamicConfig.js
abandonned: Have updated the check to look for `object` instead of !== "string" . js treats array as objects only. Therefore.

abandonned:
> @ruchiramoitra @abandonned Can we also add separate `test-cases` for dynamic config alone to check out all the scenarios ?

Have added new file for dynamicConfig test cases.

util/dynamicConfig.js
```diff
@@ -1,29 +1,47 @@
...
+    if (getFieldVal) {
+      config[field] = getFieldVal;
+      unset(event, path);
+    } else {
+      config[field] = JSON.parse(value.split("||")[1].trim()).toString();
+    }
+  }
+}
+
 function getDynamicConfig(event) {
```
saikumarrs, line 19: @abandonned we need to recursively replace the dynamic config values.
CC @ruchiramoitra

abandonned: Sorry didn't get this. How do we want to use recursion? Currently our approach is to get the field and check for its type. If type and pattern matches we send it to dynamically Configure its value.

ruchiramoitra:
> > @ruchiramoitra @abandonned Can we also add separate `test-cases` for dynamic config alone to check out all the scenarios ?
> 
> Have added new file for dynamicConfig test cases.

Need to add for destination having different kinds of inputs. Best Example GA/Fb Pixel/ Adobe

util/dynamicConfig.js
ruchiramoitra: yes discussed this with @abandonned

util/dynamicConfig.js
```diff
@@ -1,29 +1,108 @@
...
+      // first checking whether the value is array/object -> in that case we recurse and send that object/array as value
+      if (typeof value[obj] === "object") {
+        getVal(value[obj], event);
+      } // if we encounter string (actual values), then we start configuring it
+      else if (typeof value[obj] === "string") {
+        let val = value[obj];
+        // we need to configure the value here itself else we will not have path
+        if (val && val.startsWith("{{") && val.endsWith("}}")) {
+          val = val.replace("{{", "").replace("}}", "");
+          if (val.includes("||")) {
```
ruchiramoitra, line 35: This is repetitive. So can we replace this with the function `processDynamicConfig`

ruchiramoitra:
1. Clean up the code (check the commented parts out)
2. Add test cases for GA 
3. Add test cases for dynamic config
4. remove the heap and delighted test cases

util/dynamicConfig.js
abandonned: we already have `processDynamiConfig` , can we instead use `processDynamicValue`  as function name?

ruchiramoitra: I think `processDynamicConfig` has the same logic as this piece of code. Can you find a way of reusing it?

util/dynamicConfig.js
```diff
@@ -1,29 +1,108 @@
...
+            const path = val.split("||")[0].trim();
```
sanpj2292, line 36:
```suggestion
           const [pathVal, defaultVal] = val.split("||");
           const path = pathVal.trim();
           ....
           //May be we can add this too(but not sure if we need to add)
           // starts
           if (Array.isArray(defaultVal) || typeof JSON.parse(defaultVal) === 'object') {
               throw new CustomError('Invalid Default Value', 400);
           }
           // ends
           value[obj] = JSON.parse(defaultVal.trim()).toString();
```

abandonned: If not necessary we can skip this because JSON.parse will give error everytime it encouters some string/random value.

abandonned:
Created branch from production instead of production-staging.

util/dynamicConfig.js
abandonned: Updated.

util/dynamicConfig.js
abandonned: Updated

util/dynamicConfig.js
```diff
@@ -1,29 +1,62 @@
...
+    if (getFieldVal) {
+      config[field] = getFieldVal;
+      unset(event, path);
+    } else {
+      config[field] = JSON.parse(value.split("||")[1].trim()).toString();
+    }
+  }
+}
+
+function getVal(value, event) {
```
saikumarrs, line 19: This function is named `getVal` but doesn't return anything.

util/dynamicConfig.js
saikumarrs: IMO the only data types we should avoid are `function` and `symbol`. All other combinations should be supported.

util/dynamicConfig.js
abandonned: renaming it to `configureVal`

util/dynamicConfig.js
abandonned: Updated to support `objects` as well

util/dynamicConfig.js
abandonned: updated

abandonned:
Have Added some test cases.

util/dynamicConfig.js
```diff
@@ -1,29 +1,64 @@
...
-            Config[field] = JSON.parse(value.split("||")[1].trim()).toString();
-          }
+      let value = Config[field];
+      if (Array.isArray(value)) {
+        value.forEach(obj => {
+          configureVal(obj, event);
+        });
+      } else if (typeof value === "object") {
+        configureVal(value, event);
+      } else {
```
saikumarrs, line 51: Add an `else if` block to skip the data types that are not supported (`function` etc.)

__tests__/data/dynamicConfig_input.json
```diff
@@ -0,0 +1,106 @@
...
+          "channel": "web",
+          "context": {},
+          "traits": {
+            "anonymousId": "sampath",
+            "email": "sampath@gmail.com",
+            "appId": "testAppId"
+          }
+        }
+      }
+    ]
```
ruchiramoitra, line 104: add cases for booleans

__tests__/data/dynamicConfig_input.json
```diff
@@ -0,0 +1,106 @@
...
+            "anonymousId": "sampath",
+            "email": "sampath@gmail.com",
+            "appId": "testAppId"
+          }
+        }
+      },
+      {
+        "destination": {
+          "Config": {
+            "appId": [{"from" :"{{ message.traits.appId || \"<APP_ID>\" }}"} ,{ "to": "{{ message.traits.appId || \"<APP_ID>\" }}"}]
```
ruchiramoitra, line 27: give different values. Both are same objects.

ruchiramoitra: and from to should be within the same object. Create another object with from to with different values

saikumarrs: @abandonned make sure the test cases fully cover the module.

util/dynamicConfig.js
abandonned: Updated. please take a look

__tests__/data/dynamicConfig_input.json
abandonned: Have added some more test cases . Please check .

abandonned: @ruchiramoitra  , as per suggestion of @saikumarrs  we have made code more generic to support all kinds of possibilities. Please check the `dynamicConfig.js `file as well.

util/dynamicConfig.js
```diff
@@ -1,29 +1,62 @@
...
+function getDynamicConfigValue(event, value) {
+  value = value.replace("{{", "").replace("}}", "");
+  if (value.includes("||")) {
+    const path = value.split("||")[0].trim();
+    const fieldVal = get(event, path);
+    if (fieldVal) {
+      value = fieldVal;
+      unset(event, path);
+    } else {
+      value = JSON.parse(value.split("||")[1].trim());
```
abandonned, line 22: We removed .toString() since we don’t want to restrict the value to strings only. So that defaultValue can be numbers/boolean as well. @ruchiramoitra @saikumarrs

saikumarrs: @abandonned I don't think we need JSON parsing here.

util/dynamicConfig.js
```diff
@@ -1,29 +1,62 @@
...
+      typeof value !== "symbol" &&
+      typeof value !== "boolean"
+    ) {
+      const val = value.toString().trim();
+      if (isValidPattern(val)) {
+        value = getDynamicConfigValue(event, val);
       }
-    });
+    }
+    return value;
```
saikumarrs, line 49: This is not required.

abandonned: If value parameter is a boolean then it does logical operation in the if condition therefore I gave a return condition here. We can put it inside a condition.

util/dynamicConfig.js
abandonned: yeah we can remove JSON.parse. But should we keep `toString()` as suggested by @ruchiramoitra di?

util/dynamicConfig.js
```diff
@@ -1,29 +1,62 @@
...
+      });
+    } else if (
+      typeof value !== "function" &&
+      typeof value !== "symbol" &&
+      typeof value !== "boolean"
+    ) {
+      const val = value.toString().trim();
+      if (isValidPattern(val)) {
+        value = getDynamicConfigValue(event, val);
       }
```
saikumarrs, line 47:
```suggestion
      value = getDynamicConfigValue(event, value.toString().trim());
```

__tests__/data/dynamicConfig_input.json
```diff
@@ -0,0 +1,221 @@
+{
```
saikumarrs, line 1: You may have covered some of the cases already but here is a list of minimum test cases:
- "{{ a.b.c || defaultVal }}"
- "  {{ a.b.c   ||   defaultVal   }}  "
- "{{ a.b.c || "  default value  " }}"
- "{{ a.b.c || "default value }}"
- "{{ a.b.c || default value" }}"
- "{{ a.b.c || 1234 }}"
- "{{ a.b.c || true }}"
- "{{ a.b.c || 45.92340 }}"
- "{ key1: { key2: "{{ a.b.c || defaultVal }}" } }"
- "{ key1: { key2: "{{ a.b.c || defaultVal }}", key3: "{{ d.e || defaultVal }}" } }"
- "{ key1: [ "{{ a.b.c || defaultVal }}" ] }"
- "{ key1: [ "{{ a.b.c || defaultVal }}", "{{ d.e || defaultVal }}" ] }"
- "{ key1: [ { key2: { key3: "{{ a.b.c || defaultVal }}" } } ] }"
- "{ key1: [ [ { key2: { key3: "{{ a.b.c || defaultVal }}" } }, {key4: "{{ d.e || defaultVal }}" } ], "defaultVal" ] }"

If it makes sense, combine some of them in the same case.
@ruchiramoitra @abandonned add if I've missed anything.

util/dynamicConfig.js
saikumarrs: Yes

util/dynamicConfig.js
```diff
@@ -1,29 +1,62 @@
...
+    } else if (typeof value === "object") {
+      Object.keys(value).forEach(obj => {
+        value[obj] = configureVal(value[obj], event);
+      });
+    } else if (
+      typeof value !== "function" &&
+      typeof value !== "symbol" &&
+      typeof value !== "boolean"
+    ) {
+      const val = value.toString().trim();
```
saikumarrs, line 44: If string is the only data type left, we should invert the logical check and remove the `.toString()`

util/dynamicConfig.js
saikumarrs: It would still be not required as,
- `true` - boolean data type check passes and would return
- `false` - the check would fail and would return

abandonned: yes in case of fail it would return but without any value thus leading it to be `undefined`

util/dynamicConfig.js
abandonned: We should provide this information in doc that even though we accept all types of value as defaultVal, we are converting to string while configuring.

util/dynamicConfig.js
abandonned: updated

__tests__/data/dynamicConfig_input.json
abandonned: added the suggested test cases

util/dynamicConfig.js
```diff
@@ -1,29 +1,62 @@
...
+    const fieldVal = get(event, path);
+    if (fieldVal) {
+      value = fieldVal;
+      unset(event, path);
+    } else {
+      value = JSON.parse(value.split("||")[1].trim());
+    }
+  }
+  return value;
+}
```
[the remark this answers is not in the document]
abandonned: incorporated this in current code

util/dynamicConfig.js
abandonned: incorporated this in current code

util/dynamicConfig.js
saikumarrs: It would return `value` (false), right?

abandonned: oh sorry. you are talking about return statement inside the if condition.

abandonned: updated

saikumarrs (approved):

ruchiramoitra (approved):

PR #952, opened 2022-01-17
Adding Cluster mode env variable

Jayachand:
## Description of the change

> Changing default cluster mode to false.
> Currently even though cluster mode is enabled, only one replica is used. Dropping cluster mode increases ingestion rate too

> Right now CLUSTER_ENABLED is not set explicitly i.e. running in cluster mode. 
> There is no compulsory requirement for cluster setup. 

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


destTransformer.js
```diff
@@ -8,7 +8,9 @@ const { testRouter } = require("./testRouter");
 const cluster = require("./util/cluster");
 const { addPrometheusMiddleware } = require("./middleware");
 
-const clusterEnabled = process.env.CLUSTER_ENABLED !== "false";
+const clusterEnabled = process.env.CLUSTER_MODE
```
sanpj2292, line 11: Currently in deployment do we have `CLUSTER_MODE` explicitly mentioned ?
It seems like we need to explicitly mention if we want cluster mode, is that fine ? @saikumarrs @utsabc

destTransformer.js
```diff
@@ -8,7 +8,9 @@ const { testRouter } = require("./testRouter");
...
+  ? process.env.CLUSTER_MODE.toLowerCase() === "true"
+  : false;
```
saikumarrs, line 13: Let's use simpler syntax,

```suggestion
const clusterEnabled = process.env.CLUSTER_MODE?.toLowerCase() === "true";
```

destTransformer.js
saikumarrs: I'm having a deja vu moment. I think I have seen this exact comment before but the timestamp on this comment is just a few mins old.

saikumarrs: Anyway, that's a good point. We should check the configuration as this inverts the condition.

Jayachand: changed to CLUSTER_ENABLED

destTransformer.js
Jayachand: Right, transformer is in node 14 now. Added change

destTransformer.js
saikumarrs: If the CLUSTER_ENABLED is `undefined`, previously it would translate to cluster mode being enabled. The current changes don't support it anymore.
Why are we doing this? 
If it's an intended decision, have we updated/intend to update all the deployment configruations?

Jayachand: correct with this change, by default cluster mode is disabled. Also everywhere transformer is deployed, it is always associated with env variable NUM_PROCS=1 i.e. having single replica. Considering there is no specific requirement for cluster mode and with observation that memory usage is somewhat less compared to that of cluster and request processing is also faster. Hence proposing default behaviour as disabling cluster.

sanpj2292: So with this change now, cluster mode will not be available anymore & we are ok with it
Also we would not need to make change in devops or anywhere with this change in transformer
Whenever we need the cluster mode to be enabled we would need the changes to be made in devops repo for transformer service. Is that right ?

saikumarrs: Makes sense. Thanks, @Jayachand.

saikumarrs (approved):

destTransformer.js
Jayachand: Will collect metrics on cpu/memory usage with/without cluster to validate

Jayachand: With isolate-vm & several other packages being updated, Will take reading of fresh load tests (after total package updation) and evaluate this config validity.

Jayachand:
Closing this PR. Not a requirement right now

PR #977, opened 2022-02-04
Branch - add checks for undefined cases

akashrpo:
## Description of the change

> add checks for undefined cases where it was throwing error of cannot read property `variable_name` of undefined

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


saikumarrs (approved):

v0/destinations/branch/transform.js
```diff
@@ -81,25 +81,28 @@ function getCategoryAndName(rudderEventName) {
...
+    android_id: get(message, "context.android_id"),
+    idfa: get(message, "context.idfa")
+      ? get(message, "context.android_id")
+      : null,
+    idfv: get(message, "context.idfv")
+      ? get(message, "context.android_id")
+      : null,
+    aaid: get(message, "context.aaid")
+      ? get(message, "context.android_id")
+      : null,
```
utsabc, line 105: Lets not use ternary lets do `get(message, "context.aaid") || get(message, "context.android_id")`

utsabc (dismissed):

saikumarrs:
@akashrpo @utsabc we've discovered another issue in the Branch destination. See the Notion card for more details.

saikumarrs (dismissed):

v0/destinations/branch/transform.js
```diff
@@ -81,72 +59,93 @@ function getCategoryAndName(rudderEventName) {
...
-    );
+  const os = get(message, "context.os.name");
+
+  if (isDefinedAndNotNull(os)) {
+    if (isAppleFamily(os)) {
+      userData.idfa = get(message, "context.device.advertisingId");
+      userData.idfv = get(message, "context.device.id");
+    } else if (os.toLowerCase() === "android") {
+      userData.android_id =
+        get(message, "context.device.id") || get(message, "context.android_id");
```
saikumarrs, line 83: Invert the precedence order here for backward compatibility.

v0/destinations/branch/transform.js
```diff
@@ -81,72 +59,93 @@ function getCategoryAndName(rudderEventName) {
...
+      userData.aaid = get(message, "context.device.advertisingId");
```
saikumarrs: For `idfa`, `idfv` and `aaid` fields, let's keep the previous mapping as default and higher in the order.

v0/destinations/branch/transform.js
```diff
@@ -81,72 +59,98 @@ function getCategoryAndName(rudderEventName) {
...
+        get(message, "context.device.advertisingId") ||
+        get(message, "context.idfa");
+      userData.idfv =
+        get(message, "context.device.id") || get(message, "context.idfv");
+    } else if (os.toLowerCase() === "android") {
+      userData.android_id =
+        get(message, "context.device.id") || get(message, "context.android_id");
+      userData.aaid =
+        get(message, "context.device.advertisingId") ||
+        get(message, "context.aaid");
```
saikumarrs, line 89: For `idfa`, `idfv` and `aaid` fields, let's keep the previous mapping as default and higher in the order.

v0/destinations/branch/transform.js
```diff
@@ -81,72 +59,98 @@ function getCategoryAndName(rudderEventName) {
...
+    developer_identity: getFieldValueFromMessage(message, "userId"),
+    user_agent: get(message, "context.userAgent")
+  };
 
-  if (!os || !isDefinedAndNotNullAndNotEmpty(os.name)) {
-    throw new CustomError(
-      "os name is missing in the payload and please make sure to insert it at context.os.name",
-      400
-    );
+  const os = get(message, "context.os.name");
```
utsabc, line 75: Redundant code here line 75 can be done at 62 and the os variable can be reused, to avoid multiple get() usage

v0/destinations/branch/transform.js
```diff
@@ -81,72 +59,98 @@ function getCategoryAndName(rudderEventName) {
...
+
```
utsabc, line 76: Add some comments why the os is not a mandatory field for us

v0/destinations/branch/transform.js
akashrpo: IMO since taking this config from the 1st precedence is the standard way for all other dest and in docs we have and the 2nd precedence is just for back compatibility. Keeping this way may boost the user to use the standard one instead of the deprecated ones?

saikumarrs: If we always prefer the new source locations, it wouldn't be exactly backward compatible. It makes sense when we are building something new. 
As an immediate effect, it might be a breaking change. Rather, we can take the softer route where the docs point only to the recommended ones.

saikumarrs (approved):

PR #1007, opened 2022-02-23
Transformer CDK destination onboarding

shrouti1507:
## Description of the change
Onboarded statsig, variance, autopilot for new transformer CDK
> Description here

- It contains config files for all the destinations
- Unit tests are passing
- E2E is yet to be done

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


__tests__/autopilot.test.js
```diff
@@ -1,49 +1,6 @@
...
-  
-  describe("Router Tests", () => {
-    it("Payload", async () => {
-      const routerOutput = await transformer.processRouterDest(inputRouterData);
-      expect(routerOutput).toEqual(expectedRouterData);
-    });
-  });
-  
-});
+executeTransformationTest('autopilot', 'processor')
```
saikumarrs, line 5: Let's include the router test as well?

saikumarrs, line 5: There is a difference b/w the display name and the name of integration.
`Facebook Custom Audience` vs `fb_custom_audience`
The support for that is missing here.

__tests__/utilities/test-utils.js
```diff
@@ -0,0 +1,80 @@
...
+
+function executeTransformationTest(dest, transformAt) {
+  const testParams = formTestParams(dest, transformAt);
+  const { iscdkDest, transformer, expected, input } = testParams;
+  
+  const basePath = path.resolve(__dirname, "../../cdk");
+  const factory = new RudderCDK.ConfigFactory(basePath);
+
+  describe(`${dest} Tests`, () => {
+    it(`${dest} - ${transformAt} tests`, async () => {
```
saikumarrs, line 43: Introduce a new `description` field in test case data and use it here.

__tests__/utilities/test-utils.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+  const inputDataFile = fs.readFileSync(
+    path.resolve(__dirname, `../data/${dest}_${trCat}input.json`)
+  );
+  const outputDataFile = fs.readFileSync(
+    path.resolve(__dirname, `../data/${dest}_${trCat}output.json`)
+  );
+  const inputData = JSON.parse(inputDataFile);
+  const expected = JSON.parse(outputDataFile);
+	const cdkDest = ['variance'].filter(({ name }) => name === dest)[0];
```
saikumarrs, line 26: This kind of check is error-prone. Keep a list of reserved function names (`process`, `processAtRouter` etc.) and check whether the module exports it.
CC @sanpj2292

__tests__/utilities/test-utils.js
```diff
@@ -0,0 +1,80 @@
...
+const _ = require('lodash');
+const path = require("path");
+const RudderCDK = require("rudder-transformer-cdk");
+
+function formTestParams(dest, transformAt) {
+	const version = "v0";
+  
+  const transformer = require(
+    path.resolve(__dirname + `../../../${version}/destinations/${dest}/transform`)
+  );
```
saikumarrs, line 11: Why are we expecting `transform.js` to be always present?

__tests__/utilities/test-utils.js
```diff
@@ -0,0 +1,80 @@
...
+            } catch (error) {
+              console.log(error.message)
+              return error.message
+            }
+          })
+        );
+        actualData.map((actData, index) => {
+          console.log("Testing Payload no ",index)
+          if (expected[index].error) {
+            expect(actData).toEqual(expected[index].error);
```
saikumarrs, line 66: Why are we comparing only the `error` string?

__tests__/utilities/test-utils.js
```diff
@@ -0,0 +1,80 @@
...
+              )
+              expect(actualData).toEqual(expected);
+            } catch (error) {
+              console.log(error.message)
+              return error.message
+            }
+          })
+        );
+        actualData.map((actData, index) => {
+          console.log("Testing Payload no ",index)
```
saikumarrs, line 64: Test case number should be made part of the description.

__tests__/utilities/test-utils.js
```diff
@@ -0,0 +1,80 @@
...
+          input.map(async inp => {
+            try {
+              if (!iscdkDest) {
+                return await transformer.process(inp)
+              }
+              return await RudderCDK.Executor.executeStages(
+                inp,
+                factory.getConfig(dest)
+              )
+              expect(actualData).toEqual(expected);
```
saikumarrs, line 56: What's the purpose of this statement?

__tests__/utilities/test-utils.js
```diff
@@ -0,0 +1,80 @@
...
+            } catch (error) {
+              console.log(error.message)
+              return error.message
```
saikumarrs, line 59: This should result in force test case failure.

__tests__/utilities/test-utils.js
```diff
@@ -0,0 +1,80 @@
...
+          console.log("Testing Payload no ",index)
+          if (expected[index].error) {
+            expect(actData).toEqual(expected[index].error);
+          } else {
+            expect(actData).toEqual(expected[index])
+          }
+        });
+      } else {
+        actualData = await transformer.processRouterDest(input);
+        expect(actualData).toEqual(expected);
```
saikumarrs, line 73: Why isn't any elaborate checking made for router tests? Ideally, except for the API to be called on the transformer, everything should be identical.

cdk/autopilot/mapping/identifyMapping.yaml
```diff
@@ -0,0 +1,39 @@
+- destKey: Email
```
saikumarrs, line 1: This file is already under a `mapping` directory. We should not add redundant context with naming it again as `identifyMapping.yaml'.

cdk/autopilot/mapping/identifyMapping.yaml
```diff
@@ -0,0 +1,39 @@
...
+  sourceKeys:
+  - traits.email
+  - context.traits.email
+  required: false
```
saikumarrs, line 5: Keep `required: false` as the default behavior and set it here to override that.

cdk/autopilot/mapping/identifyMapping.yaml
```diff
@@ -0,0 +1,39 @@
...
+- destKey: FirstName
```
saikumarrs, line 6: Please add at least 1 line gap b/w mapping sections for improving the readability.

package.json
```diff
@@ -59,8 +59,9 @@
     "node-cache": "^5.1.2",
     "node-fetch": "^2.6.1",
     "object-hash": "^2.2.0",
-    "prom-client": "^14.0.1",
     "parse-static-imports": "^1.1.0",
+    "prom-client": "^14.0.1",
+    "rudder-transformer-cdk": "file:../rudder-transformer-cdk",
```
saikumarrs, line 64: Leaving a comment here as a reminder to remove this.

cdk/statsig/config.yaml
```diff
@@ -0,0 +1,21 @@
...
+    - page
+    - screen
+    - alias
+    - track
+    - group
+  default:
+    transformation:
+      # Here no mapping is mentioned, so we'd check for mappings existence
+      # Since no mapping file(mapperPath) is found, we will simply send message here
+      # If spreadMessage is also not present, we should probably throw an error
```
saikumarrs, line 13: The comments here are not relevant. Please remove.

cdk/variance/config.yaml
```diff
@@ -0,0 +1,19 @@
...
+  # Here since supportedMessageTypes is not present
+  # We'd assume that the below mentioned transformation
+  # is valid for all message types
+  default:
+    transformation:
+      # Here first the message would be spread on that
+      # the mappings would be overridden
+      # ResponsePayload = { ...message, ...mappedPayload }
+      spreadMessage: true
+      mapperPath: ./mappings/default.yaml
```
saikumarrs, line 11: What is the need to explicitly mention it if it can be auto-deduced?
CC @sanpj2292

saikumarrs (changes requested):
Please add a sample test and coverage reports.

__tests__/utilities/test-utils.js
sanpj2292: Is that not a big change for every destination ?

__tests__/utilities/test-utils.js
sanpj2292: Check the transformed data and validate if it matches the expected data. Can you explain what is not clear here ?

__tests__/utilities/test-utils.js
sanpj2292: Currently we're doing the same thing, probably here we can additionally check what is the statusCode that we're getting & validate it as well.

__tests__/utilities/test-utils.js
saikumarrs: We should do it nevertheless. Start with CDK destinations.

__tests__/utilities/test-utils.js
sanpj2292: Sorry I don't understand this. What do you mean by elaborate checking in this context ?

__tests__/utilities/test-utils.js
saikumarrs: It's right after a return statement.

__tests__/utilities/test-utils.js
saikumarrs: Compare the entire payload just like any other successful test case data.

cdk/autopilot/mapping/identifyMapping.yaml
sanpj2292: Agreed. This is to to serve as an example for the user that they can name it in anyway they deem fit

cdk/autopilot/mapping/identifyMapping.yaml
sanpj2292: Sure

cdk/autopilot/mapping/identifyMapping.yaml
sanpj2292: That is the default behavior here we were just explicit. But yeah, we can remove it here

__tests__/utilities/test-utils.js
saikumarrs: There should not be any difference b/w the way in which processor and router tests are handled except for the API (`process` vs `processRouterDest`).

cdk/variance/config.yaml
sanpj2292: With this path thing, we're giving user the flexibility as that they can mention the yaml anywhere inside the <destination> folder
I mean inside the folder `/v0/destinations/<destination>` they can mention the mappings yaml anywhere they like & with whatever name they're comfortable with

cdk/statsig/config.yaml
sanpj2292: Sure

cdk/variance/config.yaml
saikumarrs: It makes sense to specify when it is different from anything other than the defaults (default.yaml, track.yaml, identify.yaml, etc under mappings directory).

__tests__/utilities/test-utils.js
sanpj2292: Because that is how we implement a destination & it is how we're even getting a handler in `versionedRouter.js`
Please take a look at https://github.com/rudderlabs/rudder-transformer/blob/master/versionedRouter.js#L42-L47

__tests__/utilities/test-utils.js
sanpj2292: Got it. Wrong code

__tests__/utilities/test-utils.js
saikumarrs: Why do we expect it to be present for CDK-type destinations?

__tests__/utilities/test-utils.js
sanpj2292: What if we're checking for the failure in the test-case itself ?

__tests__/utilities/test-utils.js
sanpj2292: We are just requiring it in a more like a higher context but may be we can only require it for the destinations other than cdk

cdk/variance/config.yaml
sanpj2292: That decision is left to the user or developer

__tests__/utilities/test-utils.js
saikumarrs: We can catch the appropriate exception type and evaluate accordingly.

__tests__/utilities/test-utils.js
sanpj2292: `process` takes input as a single event and returns an array of transformed payload
`processRouterDest` takes input as an array of events and returns an array of transformed payload

The transformed payload for both of them are different, it'll be difficult to bring them under same roof.

__tests__/utilities/test-utils.js
saikumarrs: Even for CDK, you can have `transform.js`, right? which will host extraneous functions referenced in the config. The concern is that no error handling is done.

__tests__/utilities/test-utils.js
sanpj2292: What do you mean by exception type here, do you mean the class or function which throws an exception ?

cdk/variance/config.yaml
saikumarrs: I didn't see any piece of code in the CDK for assuming the defaults (till now). Anyway, in this case, it is redundantly specified in the config.

__tests__/utilities/test-utils.js
sanpj2292: But let me check if such kind of thing is possible, we can leverage it

saikumarrs: No, I meant something like this.

```
// Here "input" is an array of events for processor tests and
// array of array of events for router tests.
input.forEach(tcInput => {
  it(`test name`, async () => {
    try {
      if (transformAt == "processor") {
        actualData = transformer.process(tcInput);
      } else {
        actualData = tranformer.processRouterDest(tcInput);
      }
      // Compare actual and expected data
    }
    catch {
      // Force fail the test case if the expected exception is not raised
    }
  )};
}
```

__tests__/utilities/test-utils.js
saikumarrs: We throw exceptions of type `CustomError` for handled errors throughout the transformer. So, we should be able to compare the message, status code, and metadata.

cdk/variance/config.yaml
sanpj2292: I don't understand what defaults are you pointing about ?

cdk/autopilot/mapping/identifyMapping.yaml
arajguha: @saikumarrs The point here is to give the user the flexibility to mention the path to the file where identify mapping yaml is present. They can give any name for that matter. So the useful thing about this is, say all mappings are same for all calls, you can have one default.yaml and point to it. 
Now on top of it, we can maybe add a layer of logic to autodeduce it (which it is not currently). But this being more direct would help the user to debug in case payload transformation is not working as expected.

saikumarrs: I totally get you. I'm just saying at least the examples we give should be the right ones.
`/mappings/identifyMapping.yaml` should just have been `/mappings/identify.yaml`. We can talk about the flexibility in the documentation.
It is understood that if you name an identify mapping file as `my_identify_mapping_file.yaml`, it needs to be specified in the config.

saikumarrs: @sanpj2292 To also answer this comment,
https://github.com/rudderlabs/rudder-transformer/pull/1007#discussion_r812612019

If the mapping files are simply named after the message types, `identify.yaml`, `track.yaml` etc in mappings directly, CDK should not expect the file paths in the config.
Isn't it intuitive that if the user names the files in a way CDK doesn't understand, it should be explicitly called out in the config?

sanpj2292: @saikumarrs 
Sure, we will implement the intuitive behavior for deducing the files

__tests__/utilities/test-utils.js
sanpj2292: Removed

__tests__/autopilot.test.js
shrouti1507: done

__tests__/utilities/test-utils.js
sanpj2292: Changed this. Can you once check and let me know if there are anymore concerns ?

saikumarrs (dismissed):

temp.js
```diff
@@ -0,0 +1,222 @@
+const path = require('path')
+const RudderCDK = require('rudder-transformer-cdk')
+
+const main = () => {
+  try {
```
sanpj2292, line 5: Do we need to commit this file ?

__tests__/utilities/test-utils.js
```diff
@@ -0,0 +1,114 @@
...
+          // Compare actual and expected data
+          expect(actualData).toEqual(expected[index])
+        } catch (error) {
+          // Force fail the test case if the expected exception is not raised
+          expect(error.message).toEqual(expected[index].error)
+        }
+      });
+    });
+  });
+  if (transformAt == "router") {
```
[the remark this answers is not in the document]
shrouti1507: To add common test case for the generic block `if (!Array.isArray(inputs) || inputs.length <= 0) {
    const respEvents = getErrorRespEvents(null, 400, "Invalid event array");
    return [respEvents];` in `porcessRouterDest` function
  }

sanpj2292:
**NOTE**: Here we used the node v10 version. We'd fix the conflict in a separate branch just to make things easier for us to handle

PR #1022, opened 2022-03-01
Mp ignore lastseen

ujjwal-ab:
## Description of the change

Ignored last seen property

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


v0/destinations/mp/data/MPIdentifyConfig.json
```diff
@@ -29,6 +29,12 @@
     "required": false,
     "sourceFromGenericMap": true
   },
+  {
+    "destKey": "$ignore_time",
+    "sourceKeys": "ignoreTime",
+    "required": false,
+    "sourceFromGenericMap": true
```
utsabc, line 36: We don't have `ignoreTime` in GenericFieldMapping.json the mapping will fail at this point.

utsabc: Segment supports this by setting context.active = false for identify calls. Doing this will set $ignore_time to true for mix panel.
So rather than using the same name `ignoreTime` for source can we use `active`?

ItsSudip: My idea was, by default set ignoreTime to true. And if the user sends any other value we will accept that.

utsabc: I don't think keeping default true is good idea as during creation this will cause disparity on updated TS also we should update the updated TS always unless user sends this attribute as true

utsabc (dismissed):

v0/destinations/mp/transform.js
```diff
@@ -163,6 +163,10 @@ function getTransformedJSON(message, mappingJson, useNewMapping) {
     set(rawPayload, "$name", getFullName(message));
   }
 
+  if (!message.context.active && message.context.active === false) {
```
saikumarrs, line 166: Isn't this sufficient?

```suggestion
  if (message.context?.active === false) {
```

ujjwal-ab: Yes, it is sufficient. Both the previous conditions are actually doing the same thing.

utsabc (approved):

ItsSudip (approved):

saikumarrs (approved):

PR #1065, opened 2022-03-21
Garl hotfix

ItsSudip:
## Description of the change

We are using an npm module called sha256 to encrypt the user properties. But that module is unable to handle null values so put a check.

## Type of change
- [X] Bug fix (non-breaking change that fixes an issue)

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/v0/destinations/google_adwords_remarketing_lists/transform.js
+++ b/v0/destinations/google_adwords_remarketing_lists/transform.js
@@ -22,7 +22,7 @@ const {
 
 const hashEncrypt = object => {
   Object.keys(object).forEach(key => {
-    if (hashAttributes.includes(key)) {
+    if (hashAttributes.includes(key) && object[key]) {
       // eslint-disable-next-line no-param-reassign
       object[key] = sha256(object[key]);
     }
--- a/v0/destinations/google_adwords_remarketing_lists/transform.js
+++ b/v0/destinations/google_adwords_remarketing_lists/transform.js
@@ -142,7 +142,7 @@ const createPayload = (message, destination) => {
       );
       if (userIdentifiersList.length === 0) {
         logger.info(
-          `[Google_adwords_remarketing_list]:: No attributes are there in the ${key}.`
+          `Google_adwords_remarketing_list]:: No attributes are present in the '${key}' property.`
         );
         return;
       }
@@ -215,7 +215,7 @@ const processEvent = async (metadata, message, destination) => {
 
     if (!Object.keys(createdPayload).length) {
       throw new CustomError(
-        "[Google_adwords_remarketing_list]:: add or remove property is not present inside listData or there is not any attribute inside add or remove matching with the schemafields. Aborting message.",
+        "[Google_adwords_remarketing_list]:: Neither 'add' nor 'remove' property is present inside 'listData' or there are no attributes inside 'add' or 'remove' properties matching with the schema fields. Aborting message.",
         400
       );
     }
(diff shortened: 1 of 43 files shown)


v0/destinations/google_adwords_remarketing_lists/transform.js
```diff
@@ -22,7 +22,7 @@ const {
 
 const hashEncrypt = object => {
   Object.keys(object).forEach(key => {
-    if (hashAttributes.includes(key)) {
+    if (hashAttributes.includes(key) && object[key]) {
```
utsabc, line 25: if `object[key]` is undefined/null we are skipping here. Hence only hashing is skipped but the `null` / `undefined` values are sent to google. Does google accept those?
Secondly, if an identifier is having null/undefined shouldn't we skip sending those to google entirely?

ItsSudip: No we are not sending null values we are discarding those values.

saikumarrs: It's also skipping all the "falsy" values a.k.a `undefined`, `NaN`, ``, `false`, and `0`.

saikumarrs: What are the repercussions if we don't send the invalid values to the destination?

ItsSudip: There are not any repercussions if we don't send the invalid values.

utsabc: I think the targeting will be weak as it will be hard to identify a person with less identifiable attributes

ItsSudip: There is nothing to do with us as the customer is unable to collect the user data properly.

saikumarrs: Re-iterating this,

> It's also skipping all the "falsy" values a.k.a undefined, NaN, ``, false, and `0`.

ItsSudip: The values we are sending as user traits, can't be any falsy value.

saikumarrs: Why can't a trait be `false` or `0`?

utsabc: somebody can send `email` : 0, in that case we will skip it As Far As I understand

saikumarrs: Makes sense but we can't apply that to all the traits.
Ex: 
`numberOfVisits`: `0`
`isActive`: `false`

utsabc: Yes agreed, but from lines 78 to 117 as we are filtering over user identifiers we are safe there.

utsabc: @ItsSudip can you confirm

ItsSudip: @saikumarrs we are not sending this type of trait. Google Adwords required only the following [userIdentifiers](https://developers.google.com/google-ads/api/reference/rpc/v10/UserIdentifier).

saikumarrs: Sounds good.

utsabc (approved):

PR #1097, opened 2022-04-06
Transformer CDK: solving cdk new destination onboarding problem

shrouti1507:
## Description of the change

Previously, only old destinations were being onboarded to CDK. While onboarding new destinations the file path was not matching with new cdk format.

## Type of change
- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally
- [ ] The code changed/added as part of this pull request has been covered with tests
- [ ] All tests related to the changed code pass in development

### Code review 

- [ ]  This pull request has a descriptive title and information useful to a reviewer. There may be a screenshot or screencast attached
- [ ] "Ready for review" label attached to the PR and reviewers mentioned in a comment
- [ ] Changes have been reviewed by at least one other engineer
- [ ] Issue from task tracker has a link to this pull request

PR #1102, opened 2022-04-07
feature(new integration):DCM cloud mode Integration onboarding

akashrpo:
## Description of the change

> DCM cloud mode Integration onboarding
   This integration uses CDK to transform the events

## Type of change

## Related issues

> Fix [#1](This PR has a dependency on https://github.com/rudderlabs/rudder-transformer/pull/1097) 

### Development

- [ ] Lint rules pass locally

### Code review 


cdk/dcm_floodlight/mapping/track.yaml
```diff
@@ -0,0 +1,15 @@
+- destKey: dc_rdid
+  sourceKeys: context.device.advertisingId
+  required: true
+
+- destKey: ord
+  sourceKeys: properties.orderId
+  required: false
+
+- destKey: qty
+  sourceKeys: properties.quantity
```
utsabc, line 10: In other place it is also calculated from the sum of products array, anyway we can try that?

cdk/dcm_floodlight/mapping/track.yaml
```diff
@@ -0,0 +1,15 @@
+- destKey: dc_rdid
+  sourceKeys: context.device.advertisingId
+  required: true
+
+- destKey: ord
+  sourceKeys: properties.orderId
```
utsabc, line 6:
```suggestion
  sourceKeys: properties.orderId, properties.order_id
```

cdk/dcm_floodlight/mapping/track.yaml
```diff
@@ -0,0 +1,15 @@
...
+  required: false
+
+- destKey: qty
+  sourceKeys: properties.quantity
+  required: false
+
```
utsabc, line 12: We are missing `u$`

cdk/dcm_floodlight/transform.js
```diff
@@ -0,0 +1,136 @@
...
+
+function trackPostMapper(input, mappedPayload, rudderContext) {
+  const { message, destination } = input;
+  const { advertiserId, conversionEvents } = destination.Config;
+  let { activityTag, groupTag } = destination.Config;
+
+  const baseEndpoint = "https://ad.doubleclick.net/ddm/activity/";
+
+  let event = getValueFromMessage(message, "event");
+  if (!event) {
```
utsabc, line 59:
```suggestion
  if (!isEmpty(event)) {
```

cdk/dcm_floodlight/transform.js
```diff
@@ -0,0 +1,136 @@
...
+    throw new CustomError(
+      "[DCM Floodlight]:: event is required for track call",
+      400
+    );
+  }
+
+  let customFloodlightVariable;
```
utsabc, line 66: Lets do the variable declarations at the beginning of the function makes code easy to read

cdk/dcm_floodlight/transform.js
```diff
@@ -0,0 +1,136 @@
...
+  if (!salesTag) {
+    // for counter tag
+    mappedPayload.ord = get(message, "messageId");
+    delete mappedPayload.qty;
+    delete mappedPayload.cost;
+  }
+
+  const integrationsObj = getIntegrationsObj(message, "dcm_floodlight");
+  if (isDefinedAndNotNull(integrationsObj.COPPA)) {
+    mappedPayload.tag_for_child_directed_treatment = isValidFlag(
```
utsabc, line 103: Please provide some inline documentation on how this would be present in integrations object. Also ddm api ref

cdk/dcm_floodlight/transform.js
```diff
@@ -0,0 +1,136 @@
...
+      "COPPA",
+      integrationsObj.COPPA
+    );
+  }
+
+  if (isDefinedAndNotNull(integrationsObj.GDPR)) {
```
utsabc, line 109: Please add ddm api ref here

cdk/dcm_floodlight/transform.js
```diff
@@ -0,0 +1,136 @@
...
+    mappedPayload.tfua = isValidFlag("GDPR", integrationsObj.GDPR);
+  }
+
+  if (isDefinedAndNotNull(integrationsObj.npa)) {
+    mappedPayload.npa = isValidFlag("npa", integrationsObj.npa);
```
utsabc, line 114: Please add ddm api ref here

cdk/dcm_floodlight/transform.js
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  mappedPayload.dc_lat = get(message, "context.device.adTrackingEnabled");
+  if (isDefinedAndNotNull(mappedPayload.dc_lat)) {
+    mappedPayload.dc_lat = isValidFlag("dc_lat", mappedPayload.dc_lat);
+  }
```
utsabc, line 120: This part is bit confusing
- Why are we not using mapping.yml to map this 
- Seems like we are using isValidFlag quite a few time if multikey map is available lets use that let not go back and forth after cdk updates

cdk/dcm_floodlight/transform.js
```diff
@@ -0,0 +1,136 @@
...
+    customFloodlightVariable
+  );
+
+  rudderContext.endpoint = appendProperties(baseEndpoint, mappedPayload);
+  rudderContext.endpoint = appendProperties(
+    rudderContext.endpoint,
+    customFloodlightVariable
+  ).slice(0, -1);
+
+  return {};
```
utsabc, line 133: Lets use cdk help here, get call if we could avoid returning empty json

cdk/dcm_floodlight/transform.js
```diff
@@ -0,0 +1,136 @@
...
+  mappedPayload = removeUndefinedAndNullValues(mappedPayload);
+  customFloodlightVariable = removeUndefinedAndNullValues(
+    customFloodlightVariable
+  );
+
+  rudderContext.endpoint = appendProperties(baseEndpoint, mappedPayload);
+  rudderContext.endpoint = appendProperties(
+    rudderContext.endpoint,
+    customFloodlightVariable
+  ).slice(0, -1);
```
utsabc, line 131:
```suggestion
  const dcmEndpoint = appendProperties(baseEndpoint, mappedPayload);
  dcmEndpoint = appendProperties(
    dcmEndpoint,
    customFloodlightVariable
  ).slice(0, -1);
  
  rudderContext.endpoint = dcmEndpoint;
```
Lets not pass reference of rudderContext rather use temp variables

utsabc (changes requested):

cdk/dcm_floodlight/mapping/track.yaml
akashrpo: added support for snake case

cdk/dcm_floodlight/transform.js
akashrpo: `getValueFromMessage()` takes care of this

cdk/dcm_floodlight/transform.js
akashrpo: Done

cdk/dcm_floodlight/transform.js
akashrpo: Added comment

cdk/dcm_floodlight/transform.js
akashrpo: Added above

cdk/dcm_floodlight/transform.js
akashrpo: Added above

cdk/dcm_floodlight/transform.js
akashrpo: CDK doesn't yet support this and it requires a patch update which is being worked on.

cdk/dcm_floodlight/transform.js
akashrpo: We can do so after CDK patch fix

cdk/dcm_floodlight/mapping/track.yaml
akashrpo: DCM simply requires quantity value

cdk/dcm_floodlight/mapping/track.yaml
```diff
@@ -9,7 +9,7 @@
   required: false
 
 - destKey: qty
-  sourceKeys: properties.quantity
+  sourceKeys: properties.products
```
utsabc, line 12: No lets keep it `peoperties.quantity` here

cdk/dcm_floodlight/transform.js
```diff
@@ -91,7 +91,17 @@ function trackPostMapper(input, mappedPayload, rudderContext) {
     ...mappedPayload
   };
 
-  if (!salesTag) {
+  if (salesTag) {
```
utsabc, line 94: Here if the `products` array is present then override the `mappedPayload.qty` with  quantity calculated from number of products

saikumarrs (approved):
LGTM

cdk/dcm_floodlight/transform.js
```diff
@@ -0,0 +1,171 @@
...
+          if (product.quantity) {
+            return accumulator + product.quantity;
+          }
+          return accumulator;
+        }, 0);
+        if (quantities) {
+          mappedPayload.qty = quantities;
+        }
+      }
+    }
```
utsabc, line 123:
```suggestion
  const products = get(message, "properties.products");
  if (!isEmpty(products) && Array.isArray(products)) {
    const quantities = products.reduce((accumulator, product) => {
      if (product.quantity) {
        return accumulator + product.quantity;
      }
      return accumulator;
    }, 0);
    if (quantities) {
      mappedPayload.qty = quantities;
    }
  }
```

cdk/dcm_floodlight/transform.js
```diff
@@ -0,0 +1,171 @@
...
+      400
+    );
+  }
+  rudderContext.userAgent = userAgent;
+
+  // find conversion event
+  // some() stops execution if at least one condition is passed and returns bool
+  event = event.trim().toLowerCase();
+  const conversionEventFound = conversionEvents.some(conversionEvent => {
+    if (conversionEvent.eventName.trim().toLowerCase() === event) {
```
utsabc, line 81:
```suggestion
    if (get(conversionEvent, "eventName") === event.trim().toLowerCase() ) {
```
We are already securing event above with defined check

cdk/dcm_floodlight/transform.js
```diff
@@ -0,0 +1,171 @@
...
+      "[DCM Floodlight]:: userAgent is required for track call",
+      400
+    );
+  }
+  rudderContext.userAgent = userAgent;
+
+  // find conversion event
+  // some() stops execution if at least one condition is passed and returns bool
+  event = event.trim().toLowerCase();
+  const conversionEventFound = conversionEvents.some(conversionEvent => {
```
utsabc, line 80:
```suggestion
  const conversionEventFound = conversionEvents.some(conversionEvent => {
  if(conversionEvent) { ..
```
If conversionEvent exisits do the nextsteps, 
`conversionEvents = [undefined, null, etc]`
This can be a possibility considering control plane as black box

cdk/dcm_floodlight/transform.js
```diff
@@ -0,0 +1,171 @@
...
+    if (conversionEvent.eventName.trim().toLowerCase() === event) {
+      if (
+        !isEmpty(conversionEvent.floodlightActivityTag) &&
```
utsabc, line 83: Use `get` insted of [dot]

cdk/dcm_floodlight/transform.js
```diff
@@ -0,0 +1,171 @@
...
+
+  return endpoint;
+}
+
+// transform webapp dynamicForm custom floodlight variable
+// into {u1: value, u2: value, ...}
+function transformCustomVariable(customFloodlightVariable) {
+  const customVariable = {};
+  customFloodlightVariable.forEach(item => {
+    if (!isEmpty(item.from) && !isEmpty(item.to)) {
```
utsabc, line 25: Check validity of item before accessing its propery

cdk/dcm_floodlight/transform.js
```diff
@@ -0,0 +1,171 @@
...
+        `u${item.from.trim().replace(/u/g, "")}`
+      ] = `[${item.to.trim()}]`;
+    }
+  });
+
+  return customVariable;
+}
+
+// valid flag should be provided [1|true] or [0|false]
+function isValidFlag(key, value) {
```
utsabc, line 37: We should consider value as undefined case

cdk/dcm_floodlight/transform.js
```diff
@@ -0,0 +1,171 @@
...
+    mappedPayload.ord = get(message, "messageId");
+    delete mappedPayload.qty;
+    delete mappedPayload.cost;
+  }
+
+  // COPPA, GDPR, npa must be provided inside integration object ("DCM FLoodlight")
+  // Ref - https://support.google.com/displayvideo/answer/6040012?hl=en
+  const integrationsObj = getIntegrationsObj(message, "dcm_floodlight");
+  if (integrationsObj) {
+    if (isDefinedAndNotNull(integrationsObj.COPPA)) {
```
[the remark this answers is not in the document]
akashrpo: we already have its check above

cdk/dcm_floodlight/transform.js
akashrpo: if the `item` is not present then forEach simply won't traverse

cdk/dcm_floodlight/transform.js
akashrpo: If you check the condition before calling this function it has been handled. We don't require to double-check it.

cdk/dcm_floodlight/transform.js
akashrpo: get seems to be an overkill to access only 1st level properties.

cdk/dcm_floodlight/transform.js
akashrpo: `some` won't traverse if a particular conversion event does not exist.

cdk/dcm_floodlight/transform.js
akashrpo: We do not need `get` for 1st level children and we should keep the previous logic as it trims and lowercases the receiving events from the web app to bring it in common ground.

utsabc: We should consider the scenario `conversionEvent` being undefined its very well possible as we are using dynamic form

cdk/dcm_floodlight/transform.js
utsabc: Screenshot 2022-04-13 at 1 52 09 PM
Seeing its iteration actually

cdk/dcm_floodlight/transform.js
akashrpo: If we check webapp code for dynamicForm it can never be undefined

cdk/dcm_floodlight/transform.js
utsabc: Ok but we should validate that `conversionEvent` is defined

cdk/dcm_floodlight/transform.js
utsabc: Tested like some its `traversing`

cdk/dcm_floodlight/transform.js
utsabc: Agreed i we have a check before the function call on what we are passing as value is valid its redundant code

cdk/dcm_floodlight/transform.js
akashrpo: dynamicCustomForm takes care of this

cdk/dcm_floodlight/transform.js
akashrpo: dynamicForm values are always defined

cdk/dcm_floodlight/transform.js
akashrpo: same description applies here as stated above

utsabc: The point I am trying to make is lets treat configs as black box because in future people will be using teraform or some other tools where the configs might not be valid and then we will have to revisit this. Better to make the code resilient rather and depending on correctness of other services. We already have seen such issues in past

cdk/dcm_floodlight/transform.js
utsabc: Same point as below

cdk/dcm_floodlight/transform.js
```diff
@@ -78,7 +78,10 @@ function trackPostMapper(input, mappedPayload, rudderContext) {
   // some() stops execution if at least one condition is passed and returns bool
   event = event.trim().toLowerCase();
   const conversionEventFound = conversionEvents.some(conversionEvent => {
-    if (conversionEvent.eventName.trim().toLowerCase() === event) {
+    if (
+      conversionEvent &&
+      conversionEvent.eventName.trim().toLowerCase() === event
```
utsabc, line 83:
```suggestion
      conversionEvent &&
      conversionEvent.eventName &&
      conversionEvent.eventName.trim().toLowerCase() === event
```

utsabc (approved):

ItsSudip (approved):

PR #1113, opened 2022-04-11
feature(enhancement): Blueshift update group call

sauravlal-rudderstack:
## Description of the change

> Update Group call.

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


v0/destinations/blueshift/transform.js
```diff
@@ -116,58 +113,29 @@ const identifyResponseBuilder = async (message, category, { Config }) => {
...
-      }
-    } else {
-      throw new CustomError(
-        "[Blueshift]:: name and description are required to create an empty list, or List Id is required to add user.",
-        400
-      );
-    }
-  }
+  payload.event = "identify";
+  payload = extractCustomFields(message, payload, ["traits"], ["cookie"]);
```
shrouti1507, line 130: Suggestion: As `cookie` is being mapped from traits only and it is **not** a mandatory field, we can simply move all the properties from traits here and do not need to add it in the mapping json.

v0/destinations/blueshift/data/blueshiftGroupConfig.json
```diff
@@ -1,47 +1,85 @@
...
-    "required": false
+    "required": true
   },
   {
-    "destKey": "identifier_value",
+    "destKey": "customer_id",
     "sourceKeys": "userId",
     "sourceFromGenericMap": true,
-    "required": true
+    "required": false
```
utsabc, line 12:
```suggestion
    "required": true
```
Isn't userId mandatory for mapping?

v0/destinations/blueshift/data/blueshiftGroupConfig.json
```diff
@@ -1,47 +1,85 @@
...
+    "sourceKeys": ["context.idfv", "context.device.id"],
+    "required": false
+  },
+  {
+    "destKey": "device_manufacturer",
+    "sourceKeys": "context.device.manufacturer",
+    "required": false
+  },
+  {
+    "destKey": "os_name",
```
utsabc, line 51: Are these fields populating for each user in group ?

v0/destinations/blueshift/transform.js
```diff
@@ -116,58 +113,29 @@ const identifyResponseBuilder = async (message, category, { Config }) => {
...
+  if (!Config.eventApiKey) {
     throw new CustomError(
-      "[BLUESHIFT] User API Key required for Authentication.",
+      "[BLUESHIFT] event API Key required for Authentication.",
       400
     );
   }
 
-  const payload = constructPayload(message, MAPPING_CONFIG[category.name]);
+  let payload = constructPayload(message, MAPPING_CONFIG[category.name]);
```
utsabc, line 123:
```suggestion
  const payload = constructPayload(message, MAPPING_CONFIG[category.name]);
```

utsabc:
We have removed quite a number of integration tests are we covering all the scenarios with unit test cases?

v0/destinations/blueshift/transform.js
utsabc: In mapping JSON the mapping is sourced from specific fields like `context.device.id` which will be difficult to implement using this function

sauravlal-rudderstack:
> We have removed quite a number of integration tests are we covering all the scenarios with unit test cases?

> Yeah, previously we have lots of conditions so built a test case accordingly, but now we change, so it has no use.

v0/destinations/blueshift/transform.js
utsabc: You cannot reassign `const` but you can modify property of a `const` variable
Example 
```
const payload = { key: value };
payload.key2 = value
```
The above is allowed

```
const payload = { key: value };
payload  = { key2: value2 }
```
The above is not allowed


```
let payload = { key: value };
payload  = { key2: value2 }
```
The above is allowed

v0/destinations/blueshift/data/blueshiftGroupConfig.json
utsabc: So sending a group call to blueshift without `customer_id` works?

utsabc: How is the particular `user` identified then ?

utsabc (approved):

ItsSudip (approved):

utsabc:
Need to check `eu` instead of `EU` in transformer

utsabc (approved):

ItsSudip (approved):

akashrpo (approved):

PR #1130, opened 2022-04-22
feature(new integration): GA4 cloud mode Integration onboarding

akashrpo:
## Description of the change

> 1. GA4 cloud mode Integration onboarding
> 2. We are using the transformer network layer to send event to GA4
> 3. This PR also includes the bug fix for `constructPayload()` util function where support for handling `false` boolean value is added.

## Type of change

## Related issues

> Fix [#1](rudder-server URL encodes the endpoint before sending an event to the destination and GA4 does not validate if the endpoint is URL encoded due to which we do not see the events being reflected in the dashboard. We have a workaround i.e by utilizing the transformer network layer to send the firebase client event to GA4)

### Development

- [ ] Lint rules pass locally

### Code review 


v0/destinations/ga4/config.js
```diff
@@ -0,0 +1,115 @@
+const { getMappingConfig } = require("../../util");
+
+const ENDPOINT = "https://www.google-analytics.com/mp/collect";
+
+const ConfigCategory = {
```
utsabc, line 5: Can we add reference to Rudderstack ecom spec for each of these events

v0/destinations/ga4/data/GA4CommonConfig.json
```diff
@@ -0,0 +1,35 @@
+[
+  {
+    "destKey": "client_id",
+    "sourceKeys": "messageId",
```
utsabc, line 4: lets add support for some custom id as well for this`context.client_id || messageId`

v0/destinations/ga4/data/GA4PostScoreConfig.json
```diff
@@ -0,0 +1,17 @@
...
+    "required": true
+  },
+  {
+    "destKey": "level",
+    "sourceKeys": "properties.level",
+    "required": false
+  },
+  {
+    "destKey": "character",
+    "sourceKeys": "properties.charcharacteracter",
```
utsabc, line 14: There is typo here probably

v0/destinations/ga4/transform.js
```diff
@@ -0,0 +1,387 @@
...
+  };
+  response.params = {
+    api_secret: Config.apiSecret,
+    measurement_id: Config.measurementId
+  };
+  response.body.JSON = rawPayload;
+  return response;
+};
+
+const process = async event => {
```
utsabc, line 358: do we need async here?

v0/destinations/ga4/transform.js
```diff
@@ -0,0 +1,414 @@
...
+    delete rawPayload.client_id;
+  }
+
+  const payload = {};
+  if (eventNameMapping[event.toLowerCase()]) {
+    // GA4 standard events
+    // get event specific parameters
+    switch (event.toLowerCase()) {
+      /* Browsing Section */
+      case "products_searched":
```
utsabc, line 85: Based on this ref:https://www.rudderstack.com/docs/rudderstack-api/api-specification/rudderstack-ecommerce-events-specification/ordering/

We shouldn't consider only snake case event naming here

akashrpo: We are bringing strings to a common ground by trimming and replacing spaces with `_`. [L:40](https://github.com/rudderlabs/rudder-transformer/pull/1130/files/2b1d183dbf6234b61d1a047e1285e3259e74987c#diff-38fd1bc16aadc482899b425f3baf343c7ccfc49e7a9f0b794bae8f79a4ffaf04R40)

v0/destinations/ga4/data/GA4SpendVirtualCurrencyConfig.json
```diff
@@ -0,0 +1 @@
+[]
```
ItsSudip, line 1: Why it is empty? Is it added due to remaining consistent, like for other events we are using a config.json we should use one for this event also?

v0/destinations/ga4/transform.js
```diff
@@ -0,0 +1,414 @@
...
+
+const trackResponseBuilder = async (message, { Config }) => {
+  let event = get(message, "event");
+  if (!event) {
+    throw new CustomError("Event name is required", 400);
+  }
+
+  // trim and replace spaces with "_"
+  event = event.trim().replace(/\s+/g, "_");
+
```
ItsSudip, line 41: Apart from calling lowerCase every time we can make the event lowercase here. WDYT?

v0/destinations/ga4/utils.js
```diff
@@ -0,0 +1,259 @@
...
+    items = [];
+    products.forEach((item, index) => {
+      let element = constructPayload(
+        item,
+        mappingConfig[ConfigCategory.ITEMS.name]
+      );
+      if (!element.item_id && !element.item_name) {
+        throw new CustomError("One of product_id or name is required", 400);
+      }
+
```
ItsSudip, line 228: Don't discard the entire event for a single product.

v0/destinations/ga4/transform.js
```diff
@@ -0,0 +1,414 @@
...
+    case EventType.TRACK:
+      response = await trackResponseBuilder(message, destination);
+      break;
+    default:
+      throw new CustomError(`Message type ${messageType} not supported`, 400);
+  }
+  return response;
+};
+
+module.exports = { process };
```
ItsSudip, line 414: Can't we support router transformation?

ItsSudip:
Suggestion: 
1. Add more unit test cases.
2. Double check all the event names are properly mapped as well as config.json files.

v0/destinations/ga4/transform.js
akashrpo: Due to custom events being case sensitive

v0/destinations/ga4/utils.js
akashrpo: It is a required field. [Ref](https://developers.google.com/analytics/devguides/collection/protocol/ga4/reference/events#begin_checkout_item)

akashrpo:
Test cases are yet to be added after testing each component.

v0/destinations/ga4/transform.js
akashrpo: we are planning to use CDK for GA4 later due to which we decided not to do routerTransform for now.

v0/destinations/ga4/utils.js
ItsSudip: Yes, that I understood. My concern is we are discarding the entire call for a single product being missed the required field. Can't we implement in that way, filer out the wrong products and send the correct products only?

akashrpo: If we do this then the filtration will be done silently without the user being aware of the properties they are trying to send. IMO keeping this gives the user a better idea of what exactly needs to be passed to get proper reporting on the GA4 dashboard. cc @utsabc

v0/destinations/ga4/data/GA4SpendVirtualCurrencyConfig.json
akashrpo: Added. It got left out.

v0/destinations/ga4/data/GA4GroupConfig.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "destKey": "group_id",
+    "sourceKeys": "properties.group_id",
```
ItsSudip, line 4: We can get groupId in the root of the message.

v0/destinations/ga4/data/GA4ItemsConfig.json
```diff
@@ -0,0 +1,56 @@
...
+  },
+  {
+    "destKey": "index",
+    "sourceKeys": "position",
+    "metadata": {
+      "type": "toNumber"
+    },
+    "required": false
+  }
+]
```
ItsSudip, line 56: Can't see anything related to this file except it is mentioned once in the config.js.

akashrpo: we are using it to get items properties and is used in utils.js. `getDestinationItemProperties()`

v0/destinations/ga4/transform.js
utsabc: async doesn't make sense if you are not using promise inside a function. The feature you are mentioning its there in versionedRouter js where we make async processing of events
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function

akashrpo: removed

v0/destinations/ga4/utils.js
utsabc: I agree with @akashrpo on this but !element.item_id will not work when item id is `0` we should use defined and not null

akashrpo: Added this check

ItsSudip (dismissed):

utsabc (approved):

ItsSudip (approved):

PR #1144, opened 2022-04-27
feat: json support for warehouse destinations

Jayachand:
## Description of the change

> Json support for warehouse destinations BQ, Postgres, Redshift, Snowflake
> Assuming destination json path config to be text column with path values separated by commas ex: `testMap.nestedMap, testArray `

https://www.notion.so/rudderstacks/Json-support-in-warehouse-destinations-8a9f76ee58e543b693a88528d26c2494#5f715d8d18194d2c97c03109cffa441e

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


v0/destinations/rs/transform.js
```diff
@@ -21,6 +25,8 @@ function getDataTypeOverride(key, val, options) {
 function process(event) {
   const whSchemaVersion = event.request.query.whSchemaVersion || "v1";
   const whStoreEvent = event.destination.Config.storeFullEvent === true;
+  const destJsonPaths = event.destination?.Config?.jsonPaths || "";
+  console.log(destJsonPaths);
```
saikumarrs, line 29: Forgot to remove this?

Jayachand: done

warehouse/index.js
```diff
@@ -92,6 +95,40 @@ function excludeRudderCreatedTableNames(
...
+  eventType,
+  key,
+  val,
+  output,
+  columnTypes,
+  options,
+  jsonKey = false
+) {
+  const datatype = getDataType(key, val, options, jsonKey);
+  val = jsonKey && isJson(val) ? JSON.stringify(val) : val;
```
saikumarrs, line 109: The JSON check is already made in the above statement.
```suggestion
  val = datatype === "json" ? JSON.stringify(val) : val;
```

warehouse/util.js
```diff
@@ -18,9 +18,31 @@ const isObject = value => {
   );
 };
 
+const isJson = value => {
+  return value != null && (typeof value === "object" || Array.isArray(value));
```
saikumarrs, line 22: Why are arrays considered to be JSON?

saikumarrs (approved):

warehouse/util.js
Jayachand: according to warehouses, arrays can only be stored in json column type. This mapping is about under which column type incoming event columns fit in database

warehouse/index.js
saikumarrs: Can the value of `datatype` be `json` even if `val` is not a JSON?

Jayachand: No, at the time loading data into warehouse like snowflake, it will throw error if correct json string is not passed and column declared as json(variant in DB)

saikumarrs: I'm asking if we can rely on the value of the `datatype` variable so that the original check I proposed can be used. Otherwise, we can skip this.

warehouse/index.js
```diff
@@ -521,6 +565,14 @@ function processWarehouseMessage(message, options) {
   const skipTracksTable = options.integrationOptions.skipTracksTable || false;
   const skipReservedKeywordsEscaping =
     options.integrationOptions.skipReservedKeywordsEscaping || false;
+  // Add json key paths from integration options and destination config
```
psrikanth88, line 568: move these lines into a separate func?

warehouse/index.js
```diff
@@ -92,6 +95,40 @@ function excludeRudderCreatedTableNames(
...
+  utils,
+  eventType,
+  key,
+  val,
+  output,
+  columnTypes,
+  options,
+  jsonKey = false
+) {
+  const datatype = getDataType(key, val, options, jsonKey);
```
psrikanth88, line 108: shouldn't we always set the data type as JSON if its a jsonKey and stringify the value. This is to make sure that if its the first event and we create a JSON column for that field if its explicity passed in config. We do not want to create a date field etc if they have given it as json path. Makes sense?

psrikanth88 (changes requested):

warehouse/index.js
Jayachand: At the transformer stage, it still doesn't know whether it is first time or later. Setting above conflicts with PRD doc of supporting json to only arrays and map. Above scenario would be ideal if we change our approach of realising json type i.e.string,number,map & arrays. Just store everything as string
Note: 
- Postgres, Snowflake & RS json field can have number, string, maps, arrays
- BQ storing it as string always. Parser JSON functions are safe executable i.e. returns null instead of failure

warehouse/index.js
Jayachand: done

warehouse/index.js
Jayachand: Valid json entries will be marked as json. 
Datetime string is not marked as json even though it is declared as jsonKey

Jayachand: To avoid confusion to customers, datestring is considered as string

warehouse/index.js
Jayachand: logic has been changed. Only one time validation is done

saikumarrs (dismissed):

psrikanth88 (dismissed):

sivashanmukh (approved):
LGTM

sivashanmukh (approved):
LGTM

achettyiitr (approved):
LGTM

saikumarrs (approved):

PR #1232, opened 2022-06-07
Bugfix(integration): DCM - add support to take custom variable from properties

akashrpo:
## Description of the change

> add support to take the custom variables from properties
   add page support

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


cdk/dcm_floodlight/transform.js
```diff
@@ -19,14 +19,17 @@ function appendProperties(endpoint, payload) {
...
-function transformCustomVariable(customFloodlightVariable) {
+function transformCustomVariable(customFloodlightVariable, message) {
   const customVariable = {};
   customFloodlightVariable.forEach(item => {
     if (item && !isEmpty(item.from) && !isEmpty(item.to)) {
       // remove u if already there
-      customVariable[
-        `u${item.from.trim().replace(/u/g, "")}`
-      ] = `[${item.to.trim()}]`;
+      if (get(message, `properties.${item.to.trim()}`)) {
```
sanpj2292, line 27:
```suggestion
      const propertyItem = get(message, `properties.${item.to.trim()}`);
      if (propertyItem) {
        // Not sure how important the casting to string is here
        customVariable[`u${item.from.trim().replace(/u/g, "")}`] = `${propertyItem}`;
      }
```
Can we change the implementation to this ?

cdk/dcm_floodlight/transform.js
```diff
@@ -57,7 +60,34 @@ function trackPostMapper(input, mappedPayload, rudderContext) {
...
+    if (!category && !name) {
+      throw new CustomError(
+        "[DCM Floodlight]:: category or name is required for page",
+        400
+      );
+    }
+
+    if (category && name) {
+      message.event = `Viewed ${category} ${name} Page`;
+      message.type = "track";
```
sanpj2292, line 77:
```javascript
message.type = 'track';
```
Can we move this line to Ln 74 ?

cdk/dcm_floodlight/transform.js
```diff
@@ -57,7 +60,34 @@ function trackPostMapper(input, mappedPayload, rudderContext) {
...
+      message.event = `Viewed ${category} Page`;
+      message.type = "track";
+    } else {
+      // named pages
+      message.event = `Viewed ${name} Page`;
+      message.type = "track";
+    }
+  }
+
+  event = getValueFromMessage(message, "event");
```
sanpj2292, line 89: We can simply access `message.event` directly right ?!
Is there is a specific reason to why we are using `getValueFromMessage` function here ?

cdk/dcm_floodlight/transform.js
```diff
@@ -57,7 +60,34 @@ function trackPostMapper(input, mappedPayload, rudderContext) {
 
   const baseEndpoint = "https://ad.doubleclick.net/ddm/activity/";
 
-  let event = getValueFromMessage(message, "event");
+  let event;
+  if (message.type === "page") {
+    const { category } = message.properties;
+    const { name } = message || message.properties;
+
+    if (!category && !name) {
```
utsabc, line 68: Why cant we support unnaamed and non categorised page view events? Something like `Viewed a Page`?

cdk/dcm_floodlight/config.yaml
```diff
@@ -1,11 +1,23 @@
...
+        name: "postMapper"
+      response:
+        endpoint: "{{ rudderContext.endpoint }}"
+        method: GET
+        format: JSON
+        headers:
+          User-Agent: "{{ rudderContext.userAgent }}"
+  page:
+    transformation:
+      mapperPath: ./mapping/track.yaml
```
utsabc, line 18: `track.yaml` should be change to something genric name

cdk/dcm_floodlight/transform.js
akashrpo: Viewed a page won't be mapped to any event in the config. Generally, a user provides the event/action name to a floodlight activity.

cdk/dcm_floodlight/transform.js
akashrpo: Done

cdk/dcm_floodlight/transform.js
akashrpo: Since it's a 1st level child taking it directly should be fine. Incorporated the changes

cdk/dcm_floodlight/transform.js
```diff
@@ -57,7 +58,32 @@ function trackPostMapper(input, mappedPayload, rudderContext) {
 
   const baseEndpoint = "https://ad.doubleclick.net/ddm/activity/";
 
-  let event = getValueFromMessage(message, "event");
+  let event;
+  if (message.type === "page") {
+    const { category } = message.properties;
```
utsabc, line 63: Post mapper is missing js docs plus this can be made more understandable with inline docs

cdk/dcm_floodlight/config.yaml
akashrpo: changed to DCMFloodlightConfig

cdk/dcm_floodlight/transform.js
```diff
@@ -19,14 +18,15 @@ function appendProperties(endpoint, payload) {
...
     if (item && !isEmpty(item.from) && !isEmpty(item.to)) {
       // remove u if already there
-      customVariable[
-        `u${item.from.trim().replace(/u/g, "")}`
-      ] = `[${item.to.trim()}]`;
+      const itemValue = get(message, `properties.${item.to.trim()}`);
+      if (itemValue) {
+        customVariable[`u${item.from.trim().replace(/u/g, "")}`] = itemValue;
+      }
     }
```
utsabc, line 30: We should handle for 0, false , "", null values. The competition is removing in case presence of such values.
Also we should once verify handling of boolean values liek false and true is boolean supported in dcm?

cdk/dcm_floodlight/transform.js
utsabc: One more thing which competition supports is is json style dot notation accessors wrapped as handle bar expressions example {{ context.campaign.name}} how much of an effor is that for supporting that we can use handlebar expressions

saikumarrs: We can re-use the dynamic config module for this.

cdk/dcm_floodlight/transform.js
utsabc: But the user can map `Viewed Page` as a conversion event right?

cdk/dcm_floodlight/transform.js
sanpj2292: We already support handlebars way of implementation in case of mappings as well. You can check it in `[handleMetadataForValue](https://github.com/rudderlabs/rudder-transformer/blob/master/v0/util/index.js#L556-L559)` method. That should give us a lot more flexibility. If you meant something else, let me know

cdk/dcm_floodlight/transform.js
akashrpo: Handled all the cases. Supported types are number and string

cdk/dcm_floodlight/transform.js
akashrpo: added inline comment

cdk/dcm_floodlight/transform.js
akashrpo: It is not required. I've removed it

cdk/dcm_floodlight/transform.js
utsabc: @sanpj2292 can this be done in post mapper? Because this metadata is part of construct payload

cdk/dcm_floodlight/transform.js
```diff
@@ -101,7 +135,10 @@ function trackPostMapper(input, mappedPayload, rudderContext) {
     throw new CustomError("[DCM Floodlight]:: Conversion event not found", 400);
   }
 
-  customFloodlightVariable = transformCustomVariable(customFloodlightVariable);
+  customFloodlightVariable = transformCustomVariable(
```
ItsSudip, line 138: We are considering the properties from the message only so apart from sending the message we can send the properties only. we can provide a check for if properties exist or not if CDK is not doing it in the first place.

cdk/dcm_floodlight/transform.js
akashrpo: Makes sense. Added this enhancement

cdk/dcm_floodlight/transform.js
sanpj2292: @utsabc You can import `handlebars` or `set-value` into `transform.js` & implement it

utsabc: Got it @akashrpo lets do it

cdk/dcm_floodlight/transform.js
utsabc: We are omitting boolean values doesn't dcm support boolean values?

cdk/dcm_floodlight/transform.js
utsabc: @saikumarrs we want this feature in device mode as ewll no technical blockers there?

akashrpo: As discussed added the support to take the values first from `properties` and if not found we are taking from root level `message.*`

ItsSudip (approved):

utsabc (approved):

sanpj2292 (dismissed):

cdk/dcm_floodlight/transform.js
```diff
@@ -1,10 +1,10 @@
...
 const get = require("get-value");
 const {
-  getValueFromMessage,
   removeUndefinedAndNullValues,
   isDefinedAndNotNull
 } = require("rudder-transformer-cdk/build/utils");
+const logger = require("../../logger");
 const { CustomError, getIntegrationsObj, isEmpty } = require("../../v0/util");
 
 // append properties to endpoint
```
utsabc, line 10: You can use `join` here example
```
map().join('operator')
```

utsabc: when you do `return ${key}=${payload[key]}` from map after that you append those in string function

utsabc: Screenshot 2022-06-07 at 8 19 51 PM

cdk/dcm_floodlight/transform.js
```diff
@@ -218,10 +225,10 @@ const postMapper = (input, mappedPayload, rudderContext) => {
   );
 
   let dcmEndpoint = appendProperties(baseEndpoint, mappedPayload);
-  dcmEndpoint = appendProperties(dcmEndpoint, customFloodlightVariable).slice(
-    0,
-    -1
-  );
+  if (!isEmptyObject(customFloodlightVariable)) {
+    dcmEndpoint = dcmEndpoint.concat(";");
+    dcmEndpoint = appendProperties(dcmEndpoint, customFloodlightVariable);
```
utsabc, line 230:
```suggestion
    dcmEndpoint = `${dcmEndpoint};${appendProperties(dcmEndpoint, customFloodlightVariable)}`;
```

utsabc: yeah got it

utsabc (approved):

arnab-p (approved):

PR #1240, opened 2022-06-09
Upgraded "rudder-transformer-cdk" package to v1.4.3

devops-github-rudderstack:


versionedRouter.js
```diff
@@ -124,10 +124,8 @@ async function handleDest(ctx, version, destination) {
         parsedEvent = processDynamicConfig(parsedEvent);
         let respEvents;
         if (isCdkDestination(parsedEvent)) {
-          respEvents = await Executor.execute(
-            parsedEvent,
-            ConfigFactory.getConfig(destination)
-          );
+          const baseConfig = await ConfigFactory.getConfig(destination);
```
sanpj2292, line 127:
```javascript
const executeStartTime = new Date();
  let destHandler;
  let baseConfig;
  // Getting destination handler for non-cdk destination(s)
  if (!isCdkDestination(events[0])) {
    destHandler = getDestHandler(version, destination);
  } else {
    baseConfig = await ConfigFactory.getConfig(destination);
  }
  await Promise.all(
    events.map(async event => {
      try {
        let parsedEvent = event;
        parsedEvent.request = { query: reqParams };
        parsedEvent = processDynamicConfig(parsedEvent);
        let respEvents;
        // More safer this way ?!
        if (isCdkDestination(parsedEvent) && baseConfig) {
          respEvents = await Executor.execute(parsedEvent, baseConfig);
        } else {
          respEvents = await destHandler.process(parsedEvent);
        }
        ...
     )
)
....
```
@arnab-p @saikumarrs @utsabc What do you guys think about this sort of modification ?

saikumarrs: If we are fetching `baseConfig` at the top level, we should validate it right away. There is a bug in the current code at `if (isCdkDestination(parsedEvent) && baseConfig) {`.

saikumarrs:
This is going to be OBE with https://github.com/rudderlabs/rudder-transformer/pull/1338. Hence, closing it.
CC @sanpj2292

PR #1241, opened 2022-06-10
bugfix: Fix for transformerProxy "FORM" format

sanpj2292:
## Description of the change

[CFD](https://www.notion.so/rudderstacks/Events-to-Facebook-Pixel-for-all-free-tier-customers-is-failing-with-error-message-100-p-298429b16a5547219bba14a9f1bd9a62)

The error is happening because of this piece of [source-code](https://github.com/rudderlabs/rudder-transformer/blob/master/adapters/network.js#L197-L200)
To provide more context
```bash
> b = ['{userId: "sineinif"}', '{userId: "somfioemoig"}']
[ '{userId: "sineinif"}', '{userId: "somfioemoig"}' ]
> `${b}`
'{userId: "sineinif"},{userId: "somfioemoig"}'
> JSON.stringify(b)
'["{userId: \\"sineinif\\"}","{userId: \\"somfioemoig\\"}"]'
```
Basically we have `${b}` (something like this) in FORM so it is getting converted to string directly instead of array stringification
Which is why we were getting an exception from Facebook_pixel saying the data has to be an array

## Type of change

### Development

- [ ] Lint rules pass locally

### Code review 


adapters/network.js
```diff
@@ -196,7 +196,11 @@ const proxyRequest = async request => {
     case "FORM":
       data = new URLSearchParams();
       Object.keys(payload).forEach(key => {
-        data.append(`${key}`, `${payload[key]}`);
+        let payloadKeyStr = `${payload[key]}`;
+        if (Array.isArray(payload[key])) {
+          payloadKeyStr = JSON.stringify(payload[key]);
+        }
+        data.append(`${key}`, payloadKeyStr);
```
saikumarrs, line 203: String template literal will fail for object type as well.
```suggestion
        let payloadValStr = `${payload[key]}`;
        // covers both JSON and Array types
        if (typeof payload[key] === 'object') {
          payloadValStr = JSON.stringify(payload[key]);
        }
        data.append(`${key}`, payloadValStr);
```

adapters/network.js
```diff
@@ -196,7 +196,11 @@ const proxyRequest = async request => {
     case "FORM":
```
saikumarrs, line 196: Don't we need a similar fix for Ln 193?

adapters/network.js
sanpj2292: I did the same before committing but checked all the destinations which had `FORM` and `XML` from the unit-testcases available in the repository. Didn't find any where in an object is being sent. I'm ok with this, let me know your thoughts

adapters/network.js
sanpj2292: Checked the destinations which use `XML` the response seemed to have xml formatted string which I think is not a problem. Let me know if you think otherwise

adapters/network.js
utsabc: @sanpj2292 I think we should do that for future proofing. We still dont have any object in FORM values yet but in future we can have

sanpj2292: @utsabc @saikumarrs Done with this change. Can you please check and let me know if you feel this is ok ?

saikumarrs: Can we name the variable better? You can see my previous suggestion for reference.

sanpj2292: Can you check now and confirm ?

__tests__/data/proxy_output.json
```diff
@@ -108,5 +108,18 @@
         "status": 204
       }
     }
+  },
+  {
+    "output": {
```
saikumarrs, line 113: Don't we have a test for testing the destination request payload?
Even if we had this test previously, we wouldn't have caught the error we saw.

sanpj2292: > Even if we had this test previously, we wouldn't have caught the error we saw.

Yes we wouldn't have caught this. 

> Don't we have a test for testing the destination request payload?

Currently no

utsabc (approved):

PR #1244, opened 2022-06-13
Feature: Proxy Test Route

saikumarrs:
## Description of the change

> Added a new route to test the destination proxy feature.

## Type of change

## Related issues

> N/A

### Development

- [ ] Lint rules pass locally

### Code review 


v0/destinations/pardot/util.js
```diff
@@ -100,69 +98,9 @@ const responseHandler = destinationResponse => {
...
-      data = payload;
-      headers = { ...headers, "Content-Type": "application/json" };
-      break;
-    case "XML":
-      data = `${payload}`;
-      headers = { ...headers, "Content-Type": "application/xml" };
-      break;
-    case "FORM":
-      data = new URLSearchParams();
-      data.append("format", "json");
```
sanpj2292, line 137: @saikumarrs 
This addition of `format` key is needed for `Pardot`. Except this, everything else is the same. Kindly please add this back at least till we decide on how to handle such case(s) going forward

saikumarrs: Thanks. Addressed.

sanpj2292:
We can clearly observe that although this is a string[](`['{"a":1, "b": 1}']`) when we directly provide the `JSON.stringify`(ed) output to the `URLSearchParams()`, when it is converted to string, since there are `\\` these slashes are getting encoded but ideally only `\` should be generated(in order to match the router output)

```bash
> a = JSON.stringify(['{"a":1, "b": 1}'])
'["{\\"a\\":1, \\"b\\": 1}"]'
> c = new URLSearchParams()
URLSearchParams {}
> c.append('data', a)
undefined
> c.toString()
'data=%5B%22%7B%5C%22a%5C%22%3A1%2C+%5C%22b%5C%22%3A+1%7D%22%5D'
```

saikumarrs:
> We can clearly observe that although this is a string[](`['{"a":1, "b": 1}']`) when we directly provide the `JSON.stringify`(ed) output to the `URLSearchParams()`, when it is converted to string, since there are `\\` these slashes are getting encoded but ideally only `\` should be generated(in order to match the router output)
> 
> ```shell
> > a = JSON.stringify(['{"a":1, "b": 1}'])
> '["{\\"a\\":1, \\"b\\": 1}"]'
> > c = new URLSearchParams()
> URLSearchParams {}
> > c.append('data', a)
> undefined
> > c.toString()
> 'data=%5B%22%7B%5C%22a%5C%22%3A1%2C+%5C%22b%5C%22%3A+1%7D%22%5D'
> ```

URL decoding the above end result `%5B%22%7B%5C%22a%5C%22%3A1%2C+%5C%22b%5C%22%3A+1%7D%22%5D` gives the following output,
`["{\"a\":1, \"b\": 1}"]`
which is valid as the string is enclosed by double quotes `"` and any double quotes inside the string should be escaped.
What does the router output look like? I'm assuming something along the lines of the input value `['{"a":1, "b": 1}']`

sanpj2292:
> `%5B%22%7B%5C%22a%5C%22%3A1%2C+%5C%22b%5C%22%3A+1%7D%22%5D` gives the following output, `["{\"a\":1, \"b\": 1}"]` which is valid as the string is enclosed by double quotes `"` and any double quotes inside the string should be escaped. What does the router output look like? I'm assuming something along the lines of the input value `['{"a":1, "b": 1}']`

Yes

adapters/network.js
```diff
@@ -161,23 +162,76 @@ const httpPATCH = async (url, data, options) => {
...
+function getFormData(payload, isCustom) {
+  let data = new URLSearchParams();
+  Object.keys(payload).forEach(key => {
+    let payloadValStr = `${payload[key]}`;
+    if (typeof payload[key] === "object") {
+      payloadValStr = stringifyQueryParam(payload[key]);
+    }
+    data.append(key, payloadValStr);
+  });
+  if (!isCustom) {
```
saikumarrs, line 212: As this block should not be executed in production, a better way to handle this is,
`if (isCustom)` as the default value of the flag fails the check thus preventing accidental usages of this method.

saikumarrs, line 212: If these are unavoidable changes needed for comparison, they should be moved to the `proxyTest` controller.

adapters/network.js
```diff
@@ -161,23 +162,76 @@ const httpPATCH = async (url, data, options) => {
...
+ * Obtain FORM payload-format data to send the data to destination
+ *
+ * @param {Object} payload
+ * @returns {String}
+ */
+function getFormData(payload, isCustom) {
+  let data = new URLSearchParams();
+  Object.keys(payload).forEach(key => {
+    let payloadValStr = `${payload[key]}`;
+    if (typeof payload[key] === "object") {
```
saikumarrs, line 207: This check is redundant. It's already being done in the `stringifyQueryParam` method. In fact, the above line is also not required.

adapters/network.js
sanpj2292: The intention is to have the encoding done by default but in-case we need something custom we'd need a way to skip the encoding for the custom-cases. Hence the condition is put that way.

saikumarrs: Do we have an unavoidable use case to skip this action in production or testing? If not, let's remove it.

sanpj2292: For `Pardot` we need to add a custom query parameter so that the request would be successful. Hence for any such custom cases, we'd probably need this kind of a flag

saikumarrs: It's already taken care of for Pardot destination with its own implementation of the `prepareProxyRequest` method.

How do we define this flag? 
Custom is very generic and we are loading multiple responsibilities on it.

sanpj2292:
```javascript
const {
  prepareProxyRequest
} = require("../../../adapters/network");

...

// pardot/util.js
const { endpoint, data, method, params, headers } = prepareProxyRequest(
    request,
    true
  );

  // Modify the data
  const { payloadFormat } = getPayloadData(body);
  if (payloadFormat === "FORM") {
    data.append("format", "json");
  }
```
The `prepareProxyRequest` method returns an object with `data`(key-inside the returned object) as a string
string data-type in js doesn't have `append` method, hence we added a flag so that anyone can now operate on the URLSearchParams's object directly in order to do something very specific for that destination. Let me know if this makes sense

sanpj2292: > How do we define this flag?
Custom is very generic and we are loading multiple responsibilities on it.

Currently I think it is defined as a boolean value and yes, currently the definition is broad.
We can probably turn into options and apply them or think of another way of defining things in order to handle custom things, after probably having the whole picture(I mean completely testing the feature)
That way, we'd know what are exclusive to comparison and what are needed to go to production.

adapters/network.js
sanpj2292: Yes. Made the changes. Can you please check and confirm ?

adapters/network.js
saikumarrs: It's a slippery slope to start with. It's not a good idea for a common utility to host any caller's implementation-specific behaviors.
Once the URLSearchParams object is returned, the caller can transform the data as required. BTW, we can reconstruct a URLSearchParams object from the query string using its constructor.

As far as what we are doing in the custom block, I believe they can be safely operated directly in the payload enumeration block above.

sanpj2292: >BTW, we can reconstruct a URLSearchParams object from the query string using its constructor.

Why would need to again do something that is already done ?


> It's not a good idea for a common utility to host any caller's implementation-specific behaviors.

The common utility is merely considering it rather than doing any computation with it. It wouldn't have any knowledge of how the caller is implementing at all. But I agree that common utility should not have any caller's implementation details
Here we are merely considering if the caller wants to do the things their way or not.


> payload enumeration block above.

Can you point out where exactly it is ?

saikumarrs: Made the needed changes. Resolving this.

controllers/destinationProxy.js
```diff
@@ -0,0 +1,110 @@
...
+        const routerBodyVal = new URLSearchParams(routerDestReqPayload.body);
+        routerDestReqPayload.body = routerBodyVal;
+  
+        const proxyBodyVal = new URLSearchParams();
+        proxyDestReqPayload.body.forEach(function(value, key) {
+          const encodeAsterisk = (x) => x.replace(/\*/g, "%2A");
+          // Router encodes `*` as well
+          proxyBodyVal.append(encodeAsterisk(key), encodeAsterisk(value));
+        });
+        proxyDestReqPayload.body = proxyBodyVal;
```
sanpj2292, line 47: If you see the return of `prepareProxyRequest`, there is no `body` property at all. Why are we making use of an unknown property in comparison ?

controllers/destinationProxy.js
```diff
@@ -0,0 +1,110 @@
...
+        destinationRequestPayload: proxyDestReqPayload,
+      };
+
+      // Special handling required as Go and JavaScript encodes
+      // URL parameters differently
+      const { payloadFormat } = getPayloadData(routerDeliveryPayload.body);
+      if (payloadFormat === "FORM") {
+        // This is to make sure we encode `~` in the data coming from the router
+        // The result coming from the router is already a query parameter string
+        const routerBodyVal = new URLSearchParams(routerDestReqPayload.body);
```
sanpj2292, line 38: if you see data-structure of [routerDestReqPayload](https://github.com/rudderlabs/rudder-server/blob/1728d3ec011f2ea781a3305f29aba1e1ddf3eb67/router/transformer/transformer.go#L41-L52), we don't have `body` property at all in this. So this would be `undefined` which causes exception. Can you tell me what is the motivation behind this change ?

controllers/destinationProxy.js
```diff
@@ -0,0 +1,110 @@
...
+        routerDestReqPayload.body = routerBodyVal;
+  
+        const proxyBodyVal = new URLSearchParams();
+        proxyDestReqPayload.body.forEach(function(value, key) {
+          const encodeAsterisk = (x) => x.replace(/\*/g, "%2A");
+          // Router encodes `*` as well
+          proxyBodyVal.append(encodeAsterisk(key), encodeAsterisk(value));
```
sanpj2292, line 45: Any reason why we're encoding the key ? Because we don't see a similar thing being done while parameters are mapped in server-side

- https://github.com/rudderlabs/rudder-server/blob/1728d3ec011f2ea781a3305f29aba1e1ddf3eb67/router/network.go#L128-L132
- https://github.com/rudderlabs/rudder-server/blob/1728d3ec011f2ea781a3305f29aba1e1ddf3eb67/router/network.go#L150-L159

controllers/destinationProxy.js
saikumarrs: Fixed it.

controllers/destinationProxy.js
saikumarrs: Fixed it.

controllers/destinationProxy.js
saikumarrs: The router's output string is converted into search params object for aligning the `~` encoding automatically. Comparing at the object level gives us an advantage when the order of the entries for some reason is not the same.

sanpj2292 (approved):

PR #1259, opened 2022-06-27
Bugfix(integration):Adobe Analytics Page and Screen Calls expecting event field

shrouti1507:
## Description of the change

The page and screen events do not have `event` field. Adobe analytics was expecting this and was generating error in transformer.

- while adding properties for merchandise events, if `products.propertyName` is mentioned, it was still finding in the root properties, instead of products array. Which is a bug
- here is the [statement](https://github.com/rudderlabs/rudder-transformer/blob/master/v0/destinations/adobe_analytics/transform.js#L300-L310), where we are forEach looping on Object.keys(adobeProdEvent) which adobeProdEvent is a string equal to "custom" so it turns into
c=3.3|u=3.3|s=3.3|t=3.3|o=3.3|m=3.3, which is not the case in device mode. ref: [here](https://github.com/rudderlabs/rudder-sdk-js/blob/e98f9c440aaefeefdcc8e4d8c624969045160519/integrations/AdobeAnalytics/util.js#L515)

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/v0/destinations/adobe_analytics/transform.js
+++ b/v0/destinations/adobe_analytics/transform.js
@@ -435,14 +435,16 @@ const process = async event => {
   const formattedDestination = formatDestinationConfig(destination.Config);
   let payload;
 
+  const messageClone = { ...message };
+  if (messageType === EventType.PAGE || messageType === EventType.SCREEN) {
+    messageClone.event = message.name;
+  }
+
   switch (messageType) {
     case EventType.TRACK:
     case EventType.PAGE:
     case EventType.SCREEN:
-      if (messageType === "page" || messageType === "screen") {
-        message.event = message.name;
-      }
-      payload = handleTrack(message, formattedDestination);
+      payload = handleTrack(messageClone, formattedDestination);
       break;
     default:
       throw new Error("Message type is not supported");
--- a/v0/destinations/adobe_analytics/transform.js
+++ b/v0/destinations/adobe_analytics/transform.js
@@ -439,6 +439,9 @@ const process = async event => {
     case EventType.TRACK:
     case EventType.PAGE:
     case EventType.SCREEN:
+      if (messageType === "page" || messageType === "screen") {
+        message.event = message.name;
+      }
       payload = handleTrack(message, formattedDestination);
       break;
     default:
--- a/v0/destinations/adobe_analytics/transform.js
+++ b/v0/destinations/adobe_analytics/transform.js
@@ -261,7 +261,7 @@ const processTrackEvent = (
   }
 
   // product string section
-  let adobeProdEvent = productMerchEventToAdobeEvent[event.toLowerCase()];
+  const adobeProdEvent = productMerchEventToAdobeEvent[event.toLowerCase()];
   const prodString = [];
   if (adobeProdEvent) {
     const isSingleProdEvent =
@@ -271,6 +271,7 @@ const processTrackEvent = (
         event.toLowerCase() !== "product list viewed") ||
       !Array.isArray(properties.products);
     const productsArr = isSingleProdEvent ? [properties] : properties.products;
+    const adobeProdEventArr = adobeProdEvent.split(",");
 
     productsArr.forEach(value => {
       const category = value.category || "";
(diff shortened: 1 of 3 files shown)


v0/destinations/adobe_analytics/transform.js
```diff
@@ -439,6 +439,9 @@ const process = async event => {
     case EventType.TRACK:
     case EventType.PAGE:
     case EventType.SCREEN:
+      if (messageType === "page" || messageType === "screen") {
+        message.event = message.name;
+      }
```
saikumarrs, line 444: We can do this op before the switch block.

v0/destinations/adobe_analytics/transform.js
```diff
@@ -439,6 +439,9 @@ const process = async event => {
     case EventType.TRACK:
     case EventType.PAGE:
     case EventType.SCREEN:
+      if (messageType === "page" || messageType === "screen") {
+        message.event = message.name;
```
saikumarrs, line 443: Unless we need this updated message object in Ln 452, we should avoid updating the input.
```suggestion
        const messageClone = { ...message };
        messageClone.event = message.name;
```

v0/destinations/adobe_analytics/transform.js
```diff
@@ -439,6 +439,9 @@ const process = async event => {
     case EventType.TRACK:
     case EventType.PAGE:
     case EventType.SCREEN:
+      if (messageType === "page" || messageType === "screen") {
```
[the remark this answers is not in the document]
shrouti1507: addressed

v0/destinations/adobe_analytics/transform.js
shrouti1507: addressed

saikumarrs (dismissed):

v0/destinations/adobe_analytics/transform.js
```diff
@@ -288,10 +288,14 @@ const processTrackEvent = (
         productMerchEventToAdobeEvent[event.toLowerCase()] &&
         productMerchProperties
       ) {
+        adobeProdEvent = adobeProdEvent.toString().split(",");
```
saikumarrs, line 291: There is no need to do this in each iteration of the above `forEach` loop.

saikumarrs: Also, is `toString` necessary?

v0/destinations/adobe_analytics/transform.js
```diff
@@ -288,10 +288,14 @@ const processTrackEvent = (
...
-          if (rudderProp.productMerchProperties.startsWith("products.")) {
+          if (
+            rudderProp.productMerchProperties.startsWith("products.") &&
+            isSingleProdEvent === false
+          ) {
             const key = rudderProp.productMerchProperties.split(".");
-            const v = get(properties, key[1]);
+            const v = get(value, key[1]);
             if (isDefinedAndNotNull(v)) {
               Object.keys(adobeProdEvent).forEach(val => {
```
saikumarrs, line 300: It's an array, `adobeProdEvent.forEach` should work.

v0/destinations/adobe_analytics/transform.js
saikumarrs, line 291: Can we avoid reusing the same variable to represent the event string and its string split array?

saikumarrs (changes requested):
@shrouti1507 please add comments in the code (at least to the sections that are changed)

arnab-p (approved):

v0/destinations/adobe_analytics/transform.js
shrouti1507: addressed

saikumarrs (dismissed):

v0/destinations/adobe_analytics/transform.js
shrouti1507: addressed

saikumarrs (approved):

arnab-p (approved):

PR #1279, opened 2022-07-11
feature(integration): GA4 - add optional parameters for user reports

akashrpo:
## Description of the change

> add optional parameters in order to get user data/reports

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


v0/destinations/ga4/transform.js
```diff
@@ -242,11 +244,27 @@ const responseBuilder = (message, { Config }) => {
...
       message,
       ["properties"],
-      GA4_RESERVED_PARAMETER_EXCLUSION,
+      GA4_RESERVED_PARAMETER_EXCLUSION.concat(GA4_PARAMETERS_EXCLUSION),
       payload
     );
   }
 
+  // take optional engagement_time_msec, session_id parameters for reports
+  if (message.type === "identify" || message.type === "group") {
```
utsabc, line 253: Why can t we put this in the mapping json itself rather than writing extra code to set this?

jjhuff:
Sweet, I think this'll address some of my GA4 problems.
FWIW, Segment seems to default `engagementTime` to `1` for `page` events because that seems to be needed to get decent stats. I can do this with a transformer, but it might be nice to do it here.

Also, any ideas on what to use for `sessionId`?

akashrpo:
Hi @jjhuff, Session ID identifies the session that an event came from. These are automatically collected via gtag.js However, the identifier is not included automatically in events from Measurement Protocol. Reference - https://support.google.com/analytics/answer/9191807?hl=en#zippy=

utsabc (approved):

v0/destinations/ga4/transform.js
```diff
@@ -129,6 +130,15 @@ const responseBuilder = (message, { Config }) => {
         payload
       );
     }
+
```
ItsSudip, line 133: Don't we have to add `GA4_PARAMETERS_EXCLUSION` here in the exclusionFileds?

v0/destinations/ga4/transform.js
```diff
@@ -219,6 +247,15 @@ const responseBuilder = (message, { Config }) => {
       getGA4ExclusionList(mappingConfig[ConfigCategory.GROUP.name]),
       payload
     );
+
```
ItsSudip, line 250: Same here.

v0/destinations/ga4/transform.js
akashrpo: getGA4ExclusionList() is handling it

v0/destinations/ga4/transform.js
akashrpo: getGA4ExclusionList() is handling it

ItsSudip (approved):

PR #1331, opened 2022-08-07
bugfix(integrations): Shopify Identity Resolution Fix

utsabc:
## Description of the change

This pr attempts to address the issue of identity resolution to downstream tools via Shopify connector. This transformation is used to transform Shopify server to server events pushed by webhooks, which currently might not have any user identifier if the user chooses to stay anonymous during the later stages of the journey where these events are generated.

The goal is to use use the `cart_token` available in such events and treat it as an `anonymousId` for these events.
To stitch these same events to the events collected by the tracking script, the tracking script will always check in cookies for the same `cart_token` whenever available the tracking script will shift anonymousId to `cart_token` and make a alias call for the same.

Essentially we user will be tracked by a common identifier for both server to server events and client side events with a common id to resolve the identity resolution issue

Other improvements: 

- We are updating some of our events tracked by server to server api's to more standard naming convention for better resolution at downstream tools

Screenshot 2022-08-12 at 8 47 39 AM
Screenshot 2022-08-12 at 8 49 10 AM

As shown in above screen shot - as both the device_ids in AM (i.e anonymousIds) from front end and s2s remains consistent and as both of them are attached to the userId (Shopify customerId) we are able to stitch the user journey.
We have some unconventional events like fulfillment_create where we are not getting any user identifiers which are outliers but we are still collecting these events for any other downstream analysis. These can easily be filtered using user transform

## Type of change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Related issues

> Fix [#1]() 

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/__tests__/data/shopify_source_output.json
+++ b/__tests__/data/shopify_source_output.json
@@ -106,6 +106,7 @@
       "admin_graphql_api_id": "gid://shopify/Order/4617255092480",
       "app_id": 1354745,
       "buyer_accepts_marketing": false,
+      "cart_token": "7ed2aa8c6e74666534b023e197fffd41",
       "checkout_id": 32397512573184,
       "checkout_token": "7ed2aa8c6e74666534b023e197fffd41",
       "client_details": {
--- a/__tests__/data/shopify_source_output.json
+++ b/__tests__/data/shopify_source_output.json
@@ -385,7 +385,7 @@
       "SHOPIFY": true
     },
     "type": "track",
-    "event": "fulfillments_update",
+    "event": "Fulfillments Update",
     "properties": {
       "admin_graphql_api_id": "gid://shopify/Fulfillment/4124667937024",
       "created_at": "2022-01-05T18:13:02+05:30",
@@ -501,7 +501,7 @@
       "SHOPIFY": true
     },
     "type": "track",
-    "event": "fulfillments_update",
+    "event": "Fulfillments Update",
     "properties": {
       "admin_graphql_api_id": "gid://shopify/Fulfillment/4124667937024",
       "created_at": "2022-01-05T18:13:02+05:30",
--- a/v0/sources/shopify/util.js
+++ b/v0/sources/shopify/util.js
@@ -3,13 +3,17 @@ const {
   CustomError,
   constructPayload,
   extractCustomFields,
-  flattenJson
+  flattenJson,
+  generateUUID
 } = require("../../util");
+const logger = require("../../../logger");
 const {
   lineItemsMappingJSON,
   productMappingJSON,
   LINE_ITEM_EXCLUSION_FIELDS,
-  PRODUCT_MAPPING_EXCLUSION_FIELDS
+  PRODUCT_MAPPING_EXCLUSION_FIELDS,
+  RUDDER_ECOM_MAP,
+  SHOPIFY_TRACK_MAP
 } = require("./config");
 
 /**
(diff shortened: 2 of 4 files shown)


v0/sources/shopify/util.js
```diff
@@ -20,7 +25,9 @@ const {
  */
 const getShopifyTopic = event => {
   const { query_parameters: qParams } = event;
-  console.log("query_params", qParams);
+  logger.info(
+    `[Shopify] Input events: query_params: ${JSON.stringify(qParams)}`
```
saikumarrs, line 29:
```suggestion
    `[Shopify] Input event: query_params: ${JSON.stringify(qParams)}`
```

__tests__/data/shopify_source_output.json
```diff
@@ -501,7 +502,7 @@
       "SHOPIFY": true
     },
     "type": "track",
-    "event": "fulfillments_update",
+    "event": "Fulfillments Update",
```
saikumarrs, line 505: This will be a breaking change, right?

utsabc: Only in terms of event names will address that in change-log

v0/sources/shopify/util.js
utsabc: Addressed

saikumarrs (approved):

v0/sources/shopify/util.js
```diff
@@ -85,9 +92,45 @@ const extractEmailFromPayload = event => {
...
 
+// Hash the id and use it as anonymousId (limiting 256 -> 36 chars)
+const setAnonymousId = message => {
+  switch (message.event) {
+    case SHOPIFY_TRACK_MAP.carts_create:
+    case SHOPIFY_TRACK_MAP.carts_update:
+      message.setProperty(
+        "anonymousId",
+        message.properties?.id
+          ? sha256(message.properties.id).toString().substring(0, 36)
```
koladilip, line 103: uuid v5 does very similar thing right: generating uuid from some string

koladilip (approved):

v0/sources/shopify/util.js
koladilip: This can be taken later too

v0/sources/shopify/util.js
```diff
@@ -85,9 +92,45 @@ const extractEmailFromPayload = event => {
...
+    case SHOPIFY_TRACK_MAP.orders_paid:
+    case SHOPIFY_TRACK_MAP.orders_partially_fullfilled:
+    case RUDDER_ECOM_MAP.checkouts_create:
+    case RUDDER_ECOM_MAP.checkouts_update:
+    case RUDDER_ECOM_MAP.orders_create:
+    case RUDDER_ECOM_MAP.orders_updated:
+      message.setProperty(
+        "anonymousId",
+        message.properties?.cart_token
+          ? sha256(message.properties.cart_token).toString().substring(0, 36)
```
arjunbanker, line 120: is there any guarantee of uniqueness when taking the first 36 chars?

arjunbanker:
Hi there - thanks for adding these improvements! 

Wondering if there is any way to marshall the client anonymous session ID to the server side upon initialization? Or any other way to ensure the anonymousId used by the client and server match up completely? It seems like it should be possible as other services on the Segment platform are able to accomplish it.

The approach taken in this update seems to keep a large class of sessions decoupled. Most e-commerce sessions don't end up with a transaction or logged-in user to alias the server back to the client.

PR #1333, opened 2022-08-08
fix(tfProxy): timeout support for destination requests

sanpj2292:
## Description of the change

Changes in this PR

 - Added support for timeout in destination requests made from transformer proxy
 - [Minor] stat name correction in pardot

## Notion

[Timeout support for destination](https://www.notion.so/rudderstacks/Transformer-Proxy-Server-panics-8fae73bcfc85479299c0301387e19eec)

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


v0/destinations/google_adwords_remarketing_lists/util.js
```diff
@@ -81,12 +101,18 @@ const runTheJob = async (endpoint, headers, method, jobId) => {
...
  */
-const gaAudienceProxyRequest = async request => {
+const gaAudienceProxyRequest = async (request, reqOpts = {}) => {
   const { body, method, params, endpoint } = request;
   const { headers } = request;
   const { customerId, listId } = params;
+  // The reqOpts should be sent as an argument to the functions below
+  // but we might need to do a timeout/3 as there are 3 requests involved
+  const newReqOpts = {
+    timeout: reqOpts.timeout / 3
```
koladilip, line 114: This logic seems to be tricky as sometimes we are calling multiple times using pagination so we won't know how many calls (this may be happening during the process transformation but I am not sure if that is only place such things are implemented). Anyway dividing timeout based number requests is wrong ideally we should give fixed timeout for each call irrespective of number of calls.

sanpj2292: Let's assume that the request to destination's api takes around 1sec and for this destination, we need 3 api calls(1 definitely, the other 2 based on conditions)
Are you saying we should not set the timeout to be 3sec ? Because after we make this change, we need to follow this convention

**Note**: The timeout setting will happen at router and sent as a header to `/proxy`

versionedRouter.js
```diff
@@ -666,6 +668,9 @@ if (transformerProxy) {
         async ctx => {
           const startTime = new Date();
           ctx.set("apiVersion", API_VERSION);
+          const reqOpts = {
+            timeout: ctx.get('RdProxy-Timeout')
+          };
```
saikumarrs, line 673: `handleProxyRequest` already has access to `ctx`. I don't see a need to do this here.

v0/destinations/google_adwords_remarketing_lists/util.js
saikumarrs: I agree that we should not divide the timeout amongst the requests. Given the case that this destination has more than one n/w call in this path, the timeout value should be way higher.

sanpj2292: > the timeout value should be way higher

Yes the idea is timeout should be higher.
Let's assume that in destination docs we have a timeout of almost `1sec` mentioned

If the value is set to be `1sec` and in sending the event if there are `3` network calls, then the router has to wait for `3sec` atleast(but almost always the destination would react within that time), so lower value would mean problem!! (`Agreed`)

If the value set in router is `3sec` and in sending the event if there are `3` network calls, then the timeout in transformer per request would be `3sec` which is a higher value right ?!

Why should a network call have higher timeout value than required ?!

versionedRouter.js
sanpj2292: If you observe keenly, we don't use `ctx` inside the methods that we're writing tests on!
Followed the same pattern. Let me know if there is a disagreement

saikumarrs: I don't think that's the case. Even if it was, doesn't justify it.
Unless there is any other compelling reason, I'd fix this.

v0/destinations/google_adwords_remarketing_lists/util.js
saikumarrs: I see where you are going with this.

For a destination, assuming all the network requests in the proxy processing need the same timeout, `X`, say the maximum number of network requests that can take place is `K`.

I propose specifying two timeout values for a destination in the server config:
- `Proxy_Network_Timeout` -> `X` --- configured to each network client in the proxy path (transformer)
- `Min_Router_Proxy_Timeout` -> at least `K * X` --- configured to network client of the proxy request (router)

In situations where the number of network requests is not easily deducible (pagination, multiple request payloads, etc.) or many requests are involved, leading to a high router timeout, we can define a threshold timeout and set it to the router's network client 'ctx'.
That way, an unusually slow-performing destination will not hamper other events' processing.
Perhaps, we need to apply this strategy in the former case too. A timeout of, say, 5 mins in the router is unrealistic.

versionedRouter.js
sanpj2292: I disagree with the intention here. It doesn't make sense to include the CTX(which is `koa.Context`) in tests as it is more framework and not the ruddestack logic, upon which we'll need to write the tests
Probably it is wrong to send `ctx` as an argument to `handleProxyRequest`, we should may be pass required details only to `handleProxyRequest`

v0/destinations/google_adwords_remarketing_lists/util.js
sanpj2292: I don't see it to be much different from what I had implemented. It is just that one more variable is being maintained with high timeout value(`K*X`).

Why is it wrong to divide, when you know there might be `K` network calls and `X` is probably the timeout needed for each of them ?

v0/destinations/google_adwords_enhanced_conversions/util.js
```diff
@@ -98,24 +100,37 @@ const getConversionActionId = async (method, headers, params) => {
...
+ * @param { { timeout: number } } reqOpts extra options to be set on the http client
  * @returns
  */
-const ProxyRequest = async request => {
+const ProxyRequest = async (request, reqOpts = {}) => {
   const { body, method, endpoint, params } = request;
   const { headers } = request;
+  // The reqOpts should be sent as an argument to the functions below
+  // but we might need to do a timeout/2 as there are 2 requests involved
+  const newReqOpts = {
```
koladilip, line 111: const newReqOpts = { ... reqOpts, timeout: reqOpts.timeout / 2 }
as reqOpts might have more params in the future.

v0/destinations/google_adwords_remarketing_lists/util.js
koladilip: so How can we ensure that **router_sent_timeout_to_proxy**/K will never be less than destination timeout?

versionedRouter.js
koladilip: @sanpj2292 I agree with you sending whole context to handleProxyRequest  makes the function more complex to test so get rid of it.

versionedRouter.js
```diff
@@ -666,6 +668,9 @@ if (transformerProxy) {
...
           await handleProxyRequest(destination, ctx);
```
koladilip, line 674: reqOpts is not being passed.

v0/destinations/google_adwords_remarketing_lists/util.js
sanpj2292: ### Basically let's understand this `Rd-Proxy-Timeout`
This header is being sent from [rudder-server](https://github.com/rudderlabs/rudder-server/blob/master/router/transformer/transformer.go#L290). The timeout is [destination specific timeout config](https://github.com/rudderlabs/rudder-server/blob/master/router/transformer/transformer.go#L253-L255), also please take a look the following configurations in `router.go`
- [config defs](https://github.com/rudderlabs/rudder-server/blob/master/router/router.go#L2098-L2100)
- [Setting up transformer.go](https://github.com/rudderlabs/rudder-server/blob/master/router/router.go#L2170-L2171)

### What value should be set for destination specific timeout in Router ?
We can see that we're sending the destination specific timeout(event delivery timeout) header as `Rd-Proxy-Timeout`
As event delivery sometime involves making additional API calls(may be `K` calls with probably `X`secs timeout each) based on conditions, we can set this destination specific timeout to be `K*X` secs.

In the transformer-side, we would get the timeout header to be `K*X` secs and since K calls are possible, we will divide this timeout by K (`K*X/K`), so that every API call gets `X` secs timeout(set in the http client used by transformer)

### Limitations
- When each API call has unequal timeout, this solution doesn't solve the problem completely. As a workaround we need to set the timeout configuration such that it takes in `K*X_max`
  - K = No of API calls involved in delivery
  - X_max = Maximum timeout for an API call in event delivery

sanpj2292: ### Another Solution to solve different timeout for different API calls problem

Let 's assume there are `3` api calls, with timeouts as X1, X2, X3 (all timeouts in secs only)

- Set the destination specific timeout configuration to be `X1+X2+X3`
   - ```yaml
      Router:
        <DESTINATION>:
            httpTimeout: X1+X2+X3
      ```
- In transformer code, we should probably write in such a way that the timeout is computed before setting all the options for sending one API call
      ```javascript
      // Implementation - 1
      // DESTINATION/util.js
      const reqOptsMap = {
        apiCall_1: {
          timeout: X1
        },
        apiCall_2: {
          timeout: X2
        },
        apiCall_3: {
          timeout: X3
        },
      }
      ...
      // This function is implemented on destination by developer or one who is writing integration
      function destinationProxyRequest() {
        // Just before an api call 1 is made
        const timeout_1 = reqOpts.timeout - sum(all other API calls' timeout)
        await httpSend({...options_1, timeout: timeout_1});
        ...
        // Just before an api call 2 is made
        const timeout_2 = reqOpts.timeout - sum(all other API calls' timeout)
        await httpSend({...options_2, timeout: timeout_2});
        ...
      }
      ```

koladilip: Is the number required calls to destinations maintained somewhere? Also if this number is dynamic(depends on some parameters) then how do we handle the situation.

sanpj2292: > Is the number required calls to destinations maintained somewhere?

No

> Also if this number is dynamic(depends on some parameters) then how do we handle the situation

`Currently` if a destination has 1 api call before sending to destination and 2 api calls based on conditions then we will set the timeout value assuming 3 api calls are possible in rudder-server config

versionedRouter.js
sanpj2292: @koladilip Done. Can you please check and let me know if anything else is required ?

versionedRouter.js
sanpj2292: Done

sanpj2292:
For now, will be closing this PR as a stale one. Once the task of timeouts is being taken up in the coming sprints. Will re-open this PR or open a new one and work on the fix.

PR #1338, opened 2022-08-11
chore(tfProxy): stat tag changed to destType

sanpj2292:
## Description of the change

- To maintain parity with the stats generated from `rudder-server`, we are changing the `destination` tag to `destType`
- The order in which input events come in is same as the order of transformed events(fix made)
- Added a new tag `errorAt` to know where the error is actually occurring
  - This is different from stage
  - `stage` reveals the failure phase occurrence during data processing
  - `errorAt` reveals at which point during the transformation/delivery the failure happens
  - Takes in the below 4 values
     - proc -> Failure during processor transformation
     - rt -> Failure during router transformation
     - batch -> Failure during batch transformation
     - proxy -> Failure during data delivery to destination through transformer

## Type of change

## Related issues

https://www.notion.so/rudderstacks/d5d15ce4be354e3caae11d4ea9f41477?v=7e2d9df8de2f4251a8a80a4cf2c3f662&p=5af8ab9e12bc4247b24515504dd74f9f&pm=c

### Development

- [ ] Lint rules pass locally

### Code review 


v0/util/error.js
```diff
@@ -31,7 +31,12 @@ function ErrorBuilder() {
   };
 
   this.setStatTags = arg => {
-    this.err.statTags = arg;
+    this.err.statTags = {
+      destType: arg.destination || arg.destType,
```
koladilip, line 35: arg.destType || arg.destination would right if we are moving towards destType

koladilip (dismissed):

v0/util/error.js
```diff
@@ -31,7 +31,12 @@ function ErrorBuilder() {
   };
 
   this.setStatTags = arg => {
-    this.err.statTags = arg;
+    this.err.statTags = {
+      destType: arg.destType || arg.destination,
+      stage: arg.stage,
+      scope: arg.scope,
+      meta: arg.meta
+    };
```
saikumarrs, line 39: Add any residual tags too.

```suggestion
    const { destType, destination, ...rest } = arg;
    this.err.statTags = {
      destType: destType || destination,
      ...rest
    };
```

v0/util/error.js
```diff
@@ -31,7 +31,12 @@ function ErrorBuilder() {
   };
 
   this.setStatTags = arg => {
-    this.err.statTags = arg;
+    this.err.statTags = {
+      destType: arg.destType || arg.destination,
```
saikumarrs, line 35: Why still support `destination` even after we fix it across the repo?

sanpj2292: The fix that we have done is more through the calling methods rather than through actual destination's `transform.js` or `util.js`. Hence for now, we have continued to support the earlier semantic

v0/util/error.js
sanpj2292: But do we want to do this change as already existing code is more descriptive of how we want to send the stat tags, what fields we are expecting and all right ?

saikumarrs: The caller would give out that information. If we are concerned about incomplete tags sent from the caller, then we should add validation for the required tags.

v0/util/index.js
```diff
@@ -1313,21 +1313,37 @@ class CustomError extends Error {
...
     };
   }
+  // In previous versions of cdk, we had destination tag
+  // This block is mainly to support it and also for backward for non-cdk destinations
+  if (statTags.destination) {
+    const destType = statTags.destination
+    statTags = Object.keys(statTags).filter(k => k !== 'destination')
+      .reduce((prevMap, k) => ({ ...prevMap, [k]: statTags[k] }), {})
+    statTags.destType = destType;
+  }
```
koladilip, line 1342: if (statTags.destination) {
   statTags.destType = statTags.destination;
   delete statTags.destination;
}

koladilip (dismissed):

v0/util/error.js
saikumarrs: We don't have to do that. It'll only increase our backlog to clean up the `destination` tags later.

Let's fix them all now.

sanpj2292: For non-cdk destinations, this change seems to make sense
For the CDK destination, are you trying to suggest we upgrade the CDK version and clean the tags in CDK as well ?

v0/util/error.js
sanpj2292: Are we thinking of throwing errors if the required tags are not being sent ?

v0/util/index.js
sanpj2292: This is done. Can you check and confirm ?

v0/util/error.js
saikumarrs: Good point. I believe there are no breaking changes with latest CDK lib except for some error message format. If that's the case, let's proceed, otherwise, forget I even brought this up, but ensure to create an task in our backlog.

versionedRouter.js
```diff
@@ -104,54 +104,50 @@ async function handleDest(ctx, version, destination) {
     version,
     ...metaTags
   });
-  const respList = [];
+  // let respList = [];
```
ItsSudip, line 107: We can remove this commented line.

v0/util/error.js
saikumarrs: Not throw errors, but set some defaults maybe.

utsabc:
We are putting a good effort in updating stats, why not pass the namespace as metadata from server (shouldn't be a major change in server) and start collecting stats from transformer itself removes the dependency of passing stats to server and marshalling and capturing there will increase performance as well

sanpj2292:
> We are putting a good effort in updating stats, why not pass the namespace as metadata from server (shouldn't be a major change in server) and start collecting stats from transformer itself removes the dependency of passing stats to server and marshalling and capturing there will increase performance as well

Makes sense. Will take this up in a new PR

v0/util/index.js
```diff
@@ -1322,21 +1322,26 @@ class CustomError extends Error {
...
   if (!statTags) {
     statTags = {
-      destination,
+      destType: destination,
       stage: transformStage,
       scope: TRANSFORMER_METRIC.MEASUREMENT_TYPE.EXCEPTION.SCOPE
     };
   }
+  // In previous versions of cdk, we had destination tag
+  // This block is mainly to support it and also for backward for non-cdk destinations
```
saikumarrs, line 1336: Backward compatibility for non-CDK destinations?

v0/util/index.js
```diff
@@ -1322,21 +1322,26 @@ class CustomError extends Error {
...
   let { statTags } = error;
   if (!statTags) {
     statTags = {
-      destination,
+      destType: destination,
       stage: transformStage,
       scope: TRANSFORMER_METRIC.MEASUREMENT_TYPE.EXCEPTION.SCOPE
     };
   }
+  // In previous versions of cdk, we had destination tag
```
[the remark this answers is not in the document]
sanpj2292: This is to indicate why the block has been added. I'd rather have it until CDK is stable

v0/util/index.js
sanpj2292: This is present as protection in-case I missed changing the `destination` tag to `destType`

saikumarrs: Lol...we should identify all and fix them.

sanpj2292: Fixed them for current destinations inside implemented without CDK

v0/util/error.js
sanpj2292: This is just setting up somethings, if someone doesn't set any values it would have to handled at the outer scope. Again couple of defaults doesn't make sense.

saikumarrs: Ok. How about this?
```
    const { destType, destination, stage, scope, meta, ...rest } = arg;
    this.err.statTags = {
      destType: destType || destination,
      stage,
      scope,
      meta,
      ...rest
    };
```

saikumarrs (dismissed):

koladilip (dismissed):

koladilip (approved):

utsabc (approved):

PR #1399, opened 2022-09-06
feature(new integration): Woopra destination onboarding

anantjain45823:
## Description of the change

> Description here
Woopra Cloud Mode Destination Onboarding
## Type of change
- [✔️] New feature (non-breaking change that adds functionality)

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

- [ ✔️ ] "Ready for review" label attached to the PR and reviewers mentioned in a comment


v0/destinations/woopra/utils.js
```diff
@@ -0,0 +1,30 @@
...
+  return res;
+};
+
+/**
+ * it adds the field requiring some manipulation to the payload
+ * @param {*} message
+ * @param {*} payload
+ */
+const refinePayload = (message, payload) => {
+  const app = message.context?.app?.name + message.context?.app?.build;
```
Gauravudia, line 25: If either of `name` or `build` is null/undefined , `app` value will result to NaN.

v0/destinations/woopra/transform.js
```diff
@@ -0,0 +1,150 @@
...
+  if (!message.name) {
+    throw new CustomError("[ WOOPRA ]:: Page Name can not be empty.", 400);
+  }
+  const payload = constructPayload(
+    message,
+    mappingConfig[ConfigCategories.PAGE.name]
+  );
+  refinePayload(message, payload);
+  projectName =
+    getIntegrationsObj(message, "woopra")?.projectName || projectName;
```
Gauravudia, line 79: Don't modify config properties, create new variable

v0/destinations/woopra/utils.js
```diff
@@ -0,0 +1,30 @@
...
+  const browser = mergeObjectValues(getBrowserInfo(message.context.userAgent));
+  set(payload, "app", app);
+  set(payload, "browser", browser);
```
Gauravudia, line 28: Don't modify payload here. Create a new variable , do the changes and return that variable here.
In caller function use the response return by `refinePayload`

v0/destinations/woopra/utils.js
anantjain45823: Yeah it might be the case but while assigning the payload I am already removing those values usinhg `removeUndefinedAndNullValues()`.

v0/destinations/woopra/data/WoopraIdentifyConfig.json
```diff
@@ -0,0 +1,282 @@
...
+                        "miss",
+                        "MISS"
+                    ],
+                    "destVal": "Miss"
+                }
+            ]
+        }
+    },
+    {
+        "sourceKeys": "createdAt",
```
shrouti1507, line 47: Can you please clarify why we have chosen createdAt rather than originalTimestamp?

v0/destinations/woopra/data/WoopraIdentifyConfig.json
```diff
@@ -0,0 +1,282 @@
...
+                    "sourceVal": [
+                        "customer",
+                        "Customer",
+                        "CUSTOMER"
+                    ],
+                    "destVal": "Customer"
+                }
+            ]
+        }
+    },
```
shrouti1507, line 272: This looks like inspired from Mautic. 1) Woopra has not explicitly mentioned this fields, 2) we are using strict multimap, which means it is supposed to throw error if any other value is sent. As woopra doesn't have any ENUMS for these fields, it doesn't make sense to restrict the values. Lets remove these from here. and take the traits object and convert every key to cv_key-name and send out.

v0/destinations/woopra/data/WoopraIdentifyConfig.json
```diff
@@ -0,0 +1,282 @@
...
+                        "No",
+                        "NO",
+                        "N",
+                        "n"
+                    ],
+                    "destVal": "no"
+                }
+            ]
+        }
+    },
```
shrouti1507, line 164: This as well. Please follow the next comment.

v0/destinations/woopra/data/WoopraTrackConfig.json
```diff
@@ -0,0 +1,83 @@
...
+        "destKey": "ce_product_id"
+      },
+      {
+        "sourceKeys": "properties.rating",
+        "destKey": "ce_rating"
+      },
+      {
+        "sourceKeys": "properties.review_body",
+        "destKey": "ce_review_body"
+      },
```
shrouti1507, line 62: same here. if woopra has not mentioned any of these. Lets not include it here. Rest of the properties look generic, they can remain. Lets filter out the generic properties and add ce prefix to all keys in properties and send out

shrouti1507: As I see from the documentation, we can send user parameters via track call as well. We need to add them in the payload as well.

shrouti1507: Lets also add the https://www.rudderstack.com/docs/event-spec/standard-events/common-fields/#contextual-fields app, device, os and screen from here, as a session variable. Starting with `cs_< key name >`

v0/destinations/woopra/utils.js
Gauravudia: if we only have `message.context.app.name` then ideally we should send this .
but due to string concatenation of null/undefined 
const app = message.context?.app?.name + message.context?.app?.build;
`app` will be remove from final payload.

v0/destinations/woopra/data/WoopraTrackConfig.json
anantjain45823: for app device and os we are already mapping directly i.e. cs_, ce_, cv_ is not required and for screen they have not mentioned anything and if we are passing its not showing anywhere.

v0/destinations/woopra/data/WoopraPageConfig.json
```diff
@@ -0,0 +1,91 @@
+[
+    {
```
shrouti1507, line 2: This is the same mapping from identify. Rather here, map the fields mentioned [here](https://rudderlabs.slack.com/archives/C03THKDK7GF/p1662619819297989?thread_ts=1662574807.867089&cid=C03THKDK7GF). Some of them are available in generic fields as well.

If we have provision to merge user data with these, then lets use the constructPayload function twice for the two mapping JSONS and merge the payload generated in a single final payload.

v0/destinations/woopra/data/WoopraTrackConfig.json
```diff
@@ -0,0 +1,99 @@
+[
+  {
```
shrouti1507, line 2: This also looks like same, as identify. Lets rename that file as common field mapping or user field mapping and call that same file for multiple functions. We do not need redundant files.

v0/destinations/woopra/data/WoopraPageConfig.json
anantjain45823: Means to map some already defined (at Woopra ) events to ours and take a custom one as well just in case ?

v0/util/data/GenericFieldMapping.json
```diff
@@ -82,7 +82,6 @@
   ],
 
   "state":["traits.state", "context.traits.state"],
-  "zipcode":["traits.zipcode", "context.traits.zipcode", "traits.zip", "context.traits.zip"],
```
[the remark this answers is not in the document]
anantjain45823: Actually I noticed it was repeated.

v0/destinations/woopra/utils.js
```diff
@@ -0,0 +1,95 @@
...
+ * It adds the field (cookie,app and browser) requiring some manipulation to the payload
+ * and also the custom fields provided
+ * @param {*} message
+ * @param {*} payload
+ * @param {List} prefix=ce for properties and cv for traits or both as list
+ */
+const refinePayload = (message, payload, prefix) => {
+  let app;
+  if (message.context?.app?.name && message.context?.app?.build) {
+    app = message.context?.app?.name + message.context?.app?.build;
```
shrouti1507, line 70: app build we need not to add here. We do not do it in other destinations

v0/destinations/woopra/utils.js
```diff
@@ -0,0 +1,95 @@
...
+    set(payload, "app", app);
```
shrouti1507, line 71: We should be able to add it in mapping then

v0/util/data/GenericFieldMapping.json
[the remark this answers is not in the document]
shrouti1507: ok.

v0/destinations/woopra/utils.js
anantjain45823: Yes

v0/destinations/woopra/transform.js
```diff
@@ -0,0 +1,145 @@
...
+const trackResponseBuilder = async (message, projectName) => {
+  const endpoint = `${BASE_URL}/ce`;
+  if (!message.event) {
+    throw new CustomError("[ WOOPRA ]:: Event Name can not be empty.", 400);
+  }
+  const payload = constructPayload(
+    message,
+    mappingConfig[ConfigCategories.TRACK.name]
+  );
+  const prefix = ["ce", "cv"];
```
shrouti1507, line 53: We do not need to add both ce and cv here. Just call the identifyResponseBuilder once from here. The user fields will be available. Do not call the responseBuilder function from inside this function. Keep it outside for all. and return the actual object that you have created

v0/destinations/woopra/transform.js
```diff
@@ -0,0 +1,145 @@
...
+const pageResponseBuilder = async (message, projectName) => {
+  const endpoint = `${BASE_URL}/ce`;
+  if (!message.name) {
+    throw new CustomError("[ WOOPRA ]:: Page Name can not be empty.", 400);
+  }
+  const payload = constructPayload(
+    message,
+    mappingConfig[ConfigCategories.PAGE.name]
+  );
+  const prefix = ["ce", "cv"];
```
shrouti1507, line 69: same thing here as well. Call the identifyResponseBuilder and get the visitor properties back.

v0/destinations/woopra/config.js
```diff
@@ -0,0 +1,53 @@
+const { getMappingConfig } = require("../../util");
+
+const BASE_URL = "https://www.woopra.com/track";
+const ConfigCategories = {
+  TRACK: {
+    type: "track",
+    name: "WoopraTrackConfig"
```
shrouti1507, line 7: You can even add the prefix values here as well. And can import from the object for each event.

v0/destinations/woopra/utils.js
anantjain45823: Made changes for the same.

v0/destinations/woopra/data/WoopraPageConfig.json
shrouti1507: - The fields that are stated in woopra [doc](https://docs.woopra.com/reference/track-ce#pageviews) for page call, are already expected fields from [rudderstack page call properties](https://www.rudderstack.com/docs/event-spec/standard-events/page/#properties). 
- So, for page call, let us create a dedicated mapping JSON, where these values will be mapped from properties  and context.page object as I showed here. If any field beyond these comes in properties, lets send it out with proper prefix.

Screenshot 2022-09-08 at 8 27 04 PM

v0/destinations/woopra/data/WoopraConfig.json
```diff
@@ -0,0 +1,99 @@
...
+        "sourceKeys": ["traits.age","context.traits.age"],
+        "destKey": "cv_age"
+    },
+    {
+        "sourceKeys": "userIdOnly",
+        "destKey": "cv_id",
+        "sourceFromGenericMap":true
+    },
+    {
+        "sourceKeys": "timestamp",
```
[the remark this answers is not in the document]
anantjain45823: Inside generic mapping of timestamp we have OriginalTimestamp at the most priority.

v0/destinations/woopra/config.js
anantjain45823: And pass that to the common function . Yeah got it

v0/destinations/woopra/transform.js
anantjain45823: Building a common builder as identify builder name will raise confusions and calling that from all the three call functions.

v0/destinations/woopra/transform.js
anantjain45823: Have changed the cost name to extractedprojectname

v0/destinations/woopra/utils.js
anantjain45823: we can do it either wise but I opted for passing the payload as Param because otherwise we have to take another variable and then performing merging of two payloads which will again take space and time.

v0/destinations/woopra/data/WoopraPageConfig.json
anantjain45823: for Page call we can use either of the below 
const eventName = type === "page" ? "Loaded a Page" : "Loaded a Screen";
This one is for Mixpanel.

OR

{
    "destKey": "cv_event",
    "sourceKeys": ["event", "properties.event", "name", "properties.name"],
    "required": true,
    "metadata": {
      "defaultValue": "Web Page Viewed",
      "template": "Web Page Viewed: {{value}}"
    }
  }

we can use this as well, found under Clevertap.

v0/destinations/woopra/transform.js
```diff
@@ -0,0 +1,172 @@
...
+    ConfigCategories.PAGE.genericFields
+  );
+  const commonPayload = payload;
+  const pagePayload = constructPayload(
+    message,
+    mappingConfig[ConfigCategories.PAGE.name]
+  );
+  const mergedPayload = { ...commonPayload, ...pagePayload };
+
+  mergedPayload.event = getEvent(message);
```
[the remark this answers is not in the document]
shrouti1507: this is addressed

v0/destinations/woopra/transform.js
```diff
@@ -0,0 +1,172 @@
...
+  payload.event = message.event;
+  return responseBuilder(payload, endpoint, method, extractedProjectName);
+};
+const pageResponseBuilder = async (message, projectName) => {
+  const endpoint = `${BASE_URL}/ce`;
+  const {
+    method,
+    payload,
+    extractedProjectName
+  } = await commonPayloadGenerator(
```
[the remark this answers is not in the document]
anantjain45823: Updated

v0/destinations/woopra/transform.js
```diff
@@ -0,0 +1,160 @@
...
+  const pagePayload = constructPayload(
+    message,
+    mappingConfig[ConfigCategories.PAGE.name]
+  );
+  const mergedPayload = { ...commonPayload, ...pagePayload };
+
+  mergedPayload.event = getEvent(message);
+  return responseBuilder(mergedPayload, endpoint, method, extractedProjectName);
+};
+const process = async event => {
```
shrouti1507, line 91: again we do not need process function to be async as there is no axios

shrouti1507: this is addressed

Gauravudia (approved):

shrouti1507 (approved):

PR #1415, opened 2022-09-13
respond with retriable request error code in case of 5xx

Jayachand:
## Description of the change

> retry user transformation requests from server when control plane is down or in unresponsive state
> response code should be 809 at request level too

https://www.notion.so/rudderstacks/Mitigate-effects-of-control-plane-downtime-on-rudder-transformer-a7ed0e0ca3c84b9184a23a2c75b40729

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


util/utils.js
```diff
@@ -1,15 +1,21 @@
...
 class RespStatusError extends Error {
   constructor(message, statusCode) {
     super(message);
     this.statusCode = statusCode || 400;
   }
 }
 
+class RetryRequestError extends RespStatusError {
+  constructor(message) {
+    super(message, 809);
```
koladilip, line 11: 809? what is the reason behind using this as status code?

Jayachand: Server team needed an unique request code (not used by anyone). Randomly chosen

versionedRouter.js
```diff
@@ -374,6 +374,7 @@ if (startDestTransformer) {
         { processSessions }
       );
 
+      let ctxStatusCode = 200;
```
atzoum, line 377: Backend config is also queried while performing [tracking plan validations](https://github.com/rudderlabs/rudder-transformer/blob/a2d565f082c00058a25f841b91a0d3b27526076f/versionedRouter.js#L188), shouldn't we perform similar adaptations there as well? [This call](https://github.com/rudderlabs/rudder-transformer/blob/a2d565f082c00058a25f841b91a0d3b27526076f/versionedRouter.js#L201) might fail too, right?

saikumarrs (dismissed):

util/utils.js
koladilip: Add comment so that it will be clear in the future

versionedRouter.js
Jayachand: tracking plan validations are soft validations. If any error is thrown other than schema validation error, event will be forwarded. There would be no dropping of events.

atzoum: Can you please further explain the above statement? What error code will rudder-server receive in case a schema validation fails due to unavailability of the control plane? We wouldn't want to drop events due to such an event, correct?

Jayachand: That's correct. if control-plane is not available, response code would be 200.

atzoum: so transformer treats cp errors like the tracking plan validation is fine? Doesn't sound like the right thing to do, does it? What is the rationale behind it?

koladilip: > That's correct. if control-plane is not available, response code would be 200.

Shouldn't this be 500 so that request will be retried?

atzoum: @koladilip to provide you with some context, we cannot block the processor's pipeline. In processor we do not support retries, we need to process all events of a single batch at once.
Now, in the special case of CP unavailability we have decided that we should retry indefinitely (block the pipeline) until CP is up again, so that we don't have to mark jobs as failed, and have the customer asking us to do an event replay.

Jayachand: Discussed internally. Moving to hard limitation.
@atzoum Implemented the same error handling for tracking plan api as well.

util/customTransforrmationsStore-v1.js
```diff
@@ -54,16 +48,15 @@ async function getLibraryCodeV1(versionId) {
     version: 1
   };
   try {
+    const url = `${getLibrariesUrl}?versionId=${versionId}`;
     const startTime = new Date();
-    const response = await fetchWithProxy(
-      `${getLibrariesUrl}?versionId=${versionId}`
-    );
+    const response = await fetchWithProxy(url);
```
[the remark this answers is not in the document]
Jayachand: @atzoum Added the change

saikumarrs (approved):

atzoum (approved):

PR #1419, opened 2022-09-14
(Feature) Added the support for User Session in Amplitude

1abhishekpandey:
## Description of the change

> Added user session support for Amplitude destination.
> Previously sessionId is supported for Amplitude if this param is passed at the root. So, now the first preference is given to the `sessionId` on the root and then to `context.sessionId` and then to the  (to have the backward compatibility) of the payload, if both sessionId fields are empty then `-1` will be returned as usual.
> So, if the sessionId passed on the root of the payload is accepted, then strings and numbers were accepted, but if sessionId is being accepted from the context then only numbers will be accepted
> In `ALIAS` event we are not passing session param because Alias API of AM doesn't provide support for it.

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


ItsSudip (approved):

__tests__/data/am_output.json
```diff
@@ -3593,5 +3593,113 @@
...
+                    "user_properties": {
+                        "anonymousId": "5d205961641ee6c5",
+                        "id": "User Android",
+                        "userId": "User Android"
+                    },
+                    "event_type": "Viewed Screen event Screen",
+                    "user_id": "User Android",
+                    "device_brand": "Google",
+                    "time": 1662393883250,
+                    "session_id": "1662393792",
```
utsabc, line 3634: This is Long defined by Amplitude docs. We are sending string
Screenshot 2022-09-15 at 1 14 29 PM

utsabc (changes requested):

__tests__/data/am_output.json
desusai7: @utsabc , we are handling it now, can you please take a look again !

utsabc: Sure

utsabc (approved):

v0/destinations/am/transform.js
```diff
@@ -84,15 +84,27 @@ const aliasEndpoint = destConfig => {
...
-  const sessionId = payload.session_id;
-  if (sessionId) {
-    if (typeof sessionId === "string") {
-      return sessionId.substr(sessionId.lastIndexOf(":") + 1, sessionId.length);
-    }
-    return sessionId;
+function handleSessionIdUnderRoot(message) {
+  const sessionId = get(message, "session_id");
+  if (typeof sessionId === "string") {
+    return sessionId.substr(sessionId.lastIndexOf(":") + 1, sessionId.length);
```
ItsSudip, line 90: Can't we simply provide 0 apart from `sessionId.lastIndexOf(":") + 1`?

desusai7: So, the reason why we are starting from the `lastIndexOf(":)`, is because when the sessionId in the payload is in the format `"User:123456"`,  we want to consider only `123456`

ItsSudip (approved):

PR #1429, opened 2022-09-20
refactor(algolia): error handling with builder for better alerting

sanpj2292:
## Description of the change

A simple error was being categorised with scope `exception`, resulting in repeated alerts. Something like this, could be a problem for other such error which could be categorised with a different `meta` like `badParam` to reduce the number of alerts triggered. Hence through this PR, we are refactoring such errors in `algolia`

"error_reason","count"
"""Missing required value from \""properties.index\""""",2515

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/v0/destinations/algolia/transform.js
+++ b/v0/destinations/algolia/transform.js
@@ -1,7 +1,8 @@
 const set = require("set-value");
 const { EventType } = require("../../../constants");
+const { TRANSFORMER_METRIC } = require("../../util/constant");
+const ErrorBuilder = require("../../util/error");
 const {
-  CustomError,
   getValueFromMessage,
   constructPayload,
   defaultRequestConfig,
@@ -13,7 +14,12 @@ const {
   handleRtTfSingleEventError
 } = require("../../util/index");
 
-const { ENDPOINT, MAX_BATCH_SIZE, trackMapping } = require("./config");
+const {
+  ENDPOINT,
+  MAX_BATCH_SIZE,
+  trackMapping,
+  DESTINATION
+} = require("./config");
 
 const {
   genericpayloadValidator,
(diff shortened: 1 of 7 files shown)


v0/destinations/algolia/transform.js
```diff
@@ -22,10 +28,20 @@ const {
   clickPayloadValidator
 } = require("./util");
 
+const DestTfError = new ErrorBuilder()
```
saikumarrs, line 31: We shouldn't be reusing the ErrorBuilder instance. Perhaps a better way to achieve the same is:

```
const DefStatTags = {
      destType: DESTINATION,
      stage: TRANSFORMER_METRIC.TRANSFORMER_STAGE.TRANSFORM,
      scope: TRANSFORMER_METRIC.MEASUREMENT_TYPE.TRANSFORMATION.SCOPE,
    };
.
.
.
// some point in the code,
// throw an actual error object
// either overriding defaults or adding extra info
throw new ErrorBuilder()
  .setStatTags(DefStatTags)
  .setMeta(
    TRANSFORMER_METRIC.MEASUREMENT_TYPE.TRANSFORMATION.META.BAD_PARAM
  )
  .build();

```

v0/destinations/algolia/transform.js
```diff
@@ -92,17 +126,31 @@ const trackResponseBuilder = (message, { Config }) => {
 const process = event => {
   const { message, destination } = event;
   if (!message.type) {
-    throw new CustomError(
```
saikumarrs, line 95: The error object properties are different than before.
```
// prev
{
  message: "message Type is not present. Aborting message.",
  response: {
    status: 400
  }
}

// current
{
  message: "message Type is not present. Aborting message.",
  status: 400,
  statTags: {
  // the tags we set
  }
}
```

Are we okay with it?

saikumarrs (changes requested):

v0/destinations/algolia/transform.js
sanpj2292: By okay, if you meant that this change would cause some kind of error on server. That is not the case

saikumarrs: Either of these formats must fail due to the data contract breach, right?

sanpj2292: In the `processRouterDest` method's try-catch block(inside the `.map`) we do something like this.

```javascript
// Ref: https://github.com/rudderlabs/rudder-transformer/blob/master/v0/util/index.js#L1589-L1612
const errRespEvent = handleRtTfSingleEventError(
          input,
          error,
          "ALGOLIA"
        );
```

This will make sure that, we adhere to the contract. Let me know if it makes sense

PR #1444, opened 2022-09-26
fix: facebook pixel proxy tests

sanpj2292:
## Description of the change

We have identified an issue with the proxy request mocks for `facebook_pixel` destination. Through this PR we are trying to fix it.
Also we have added a new `errorCode` and `subCode` combination and also added a test-case for the same

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


saikumarrs (approved):

v0/destinations/facebook_pixel/network-handler.js
```diff
@@ -59,6 +59,13 @@ const errorDetailsMap = {
         scope: TRANSFORMER_METRIC.MEASUREMENT_TYPE.API.SCOPE,
         meta: TRANSFORMER_METRIC.MEASUREMENT_TYPE.API.META.ABORTABLE
       }
+    },
+    33: {
+      status: 500,
+      statTags: {
+        scope: TRANSFORMER_METRIC.MEASUREMENT_TYPE.API.SCOPE,
+        meta: TRANSFORMER_METRIC.MEASUREMENT_TYPE.API.META.RETRYABLE
+      }
```
ItsSudip, line 68: According to the documentation we need to update the access token before retrying the event. So if we send a retryable error from here with 500 responses, it will continue retrying but will not succeed until the access token is updated. Apart from that if we return 400 and abort the event with a proper message like what might be the issue, that will be correct. WDYT?

saikumarrs (approved):

v0/destinations/facebook_pixel/network-handler.js
sanpj2292: The idea behind sending this `retryable` meta is that, we would want to keep this in-play until someone actually changes the token in configuration so that when the event retries it is successful. But if it makes sense, we can return `400` for this error and then make a decision on how to comprehensively handle this case.

ItsSudip: With the retriable approach, the event will be retried until or unless the access token gets updated. But events will not be discarded. Now with the second approach events will get discarded and the customer has to update the access token and retry the event as well if possible. With this in mind, the first approach looks safer. So let's stick with that.

ItsSudip (dismissed):

v0/destinations/facebook_pixel/network-handler.js
sanpj2292: I think configuration problem should probably be aborted as if not corrected, it could potentially cause bottlenecks in the system. So probably will change this to an abortable kind of case for now. Does this make sense ?

ItsSudip: Yeah we can do that as well.

v0/destinations/facebook_pixel/network-handler.js
```diff
@@ -59,14 +59,14 @@ const errorDetailsMap = {
...
-        meta: TRANSFORMER_METRIC.MEASUREMENT_TYPE.API.META.RETRYABLE
-      }
     }
+    // 33: {
+    //   status: 500,
+    //   statTags: {
+    //     scope: TRANSFORMER_METRIC.MEASUREMENT_TYPE.API.SCOPE,
+    //     meta: TRANSFORMER_METRIC.MEASUREMENT_TYPE.API.META.RETRYABLE
+    //   }
+    // }
```
ItsSudip, line 69: Can you add some comments regarding this commented code?

saikumarrs: @sanpj2292, how about we still keep this but just qualify it as abortable?

sanpj2292: @saikumarrs By default we're actually using `abortable` only. That is the reason I commented this. You can verify the same in `defaultStatTags` variable

saikumarrs: Since this is a valid and known status code, no harm in keeping it. Hence, my suggestion.

sanpj2292: Changed it

saikumarrs (approved):

koladilip (approved):

PR #1448, opened 2022-09-28
feat: integrate cdk v2

saikumarrs:
## Description of the change

Integrated CDK 2.0.

## Type of change

## Related issues

N/A

### Development

- [x] Lint rules pass locally

### Code review 


koladilip (approved):

cdk/v2/utils.js
```diff
@@ -37,8 +43,39 @@ function getPlatformBindingsPaths() {
...
+  let errorInfo = err;
+  if (err instanceof WorkflowExecutionError) {
+    logger.error(
+      "Error occurred during workflow step execution: ",
+      err.stepName
+    );
+    errorInfo = {
+      message: err.message,
+      status: err.status,
+      destinationResponse: err.error.destinationResponse,
```
koladilip, line 57: err.error could be undefined.

saikumarrs: Fixed it.

cdk/v2/utils.js
```diff
@@ -0,0 +1,81 @@
+const path = require("path");
+const fs = require("fs");
+const {
+  WorkflowExecutionError
+} = require("rudder-workflow-engine/build/errors");
```
sanpj2292, line 5: We should expose the modules or functions in a better way. Just a note @koladilip (even I missed this with prev version of CDK 🙂)

saikumarrs: I agree. It's a matter of publishing the package from the `build` directory. Let me make that change as part of the next release.

versionedRouter.js
```diff
@@ -90,6 +93,38 @@ const functionsEnabled = () => {
...
+
+    // Dump the raw error
+    logger.error(err);
+
+    return {
+      metadata: parsedEvent.metadata,
+      statusCode: errObj.status,
+      error: errObj.message || "Error occurred while processing the payload",
+      statTags: {
+        errorAt: TRANSFORMER_METRIC.ERROR_AT.PROC,
```
sanpj2292, line 121: It is not always going to be a `PROC` only as in CDK V2, we  inherently support have support for router transformation type of logic as well.
May be we need to accept one more argument like `errorAt` and pass it explicitly from where this method is called.

Something like below should solve our problem. If there is a more intuitive solution in mind, even that works
```javascript
async function handleCdkV2(destName, parsedEvent, errorAt) {
...
    statTags: {
        errorAt,
        ...
    }
}

// handleDest - processor transformation
    respEvents = await handleCdkV2(destination, parsedEvent, TRANSFORMER_METRIC.ERROR_AT.PROC);

// routerHandleDest - router transformation
   respEvents = await handleCdkV2(destination, parsedEvent, , TRANSFORMER_METRIC.ERROR_AT.RT);
```

cdk/v2/utils.js
koladilip: We can just use the main field in package.json
https://stackoverflow.com/questions/22512992/how-to-use-the-main-parameter-in-package-json#22513200

versionedRouter.js
saikumarrs: I added it as a default value which can get overridden if the stat tags (that user sent) contain it.
I've incorporated your suggestion.

cdk/v2/utils.js
saikumarrs: Yes. We already have defined it. However, not all the types (ex: `error.ts`) are exported from `index.ts`.
Hence, I could fix this issue for some of the imports.
We can resolve others here once we update the package in the next release.

saikumarrs: https://github.com/rudderlabs/rudder-workflow-engine/pull/14

versionedRouter.js
sanpj2292: Rather than having a default we should thrown an error if `flowType` is not defined, as this is going to be at `versionedRouter` level we should be good with even throwing an error.

cdk/v2/handler.js
```diff
@@ -0,0 +1,29 @@
+const { WorkflowUtils, WorkflowEngine } = require("rudder-workflow-engine");
+const {
+  getRootPathForDestination,
+  getWorkflowPath,
+  getPlatformBindingsPaths
+} = require("./utils");
+
+const destinationWorkflowEngineMap = new Map();
+
+function getWorkflowEngine(destName) {
```
koladilip, line 10: destination can have more than one workflow so how do we handle that?
We can come up some standard way of naming these.

versionedRouter.js
saikumarrs: Throw an error and fail the event?

At the end of the day, it's just another meta information in the metrics. If we realize any bug after the fact, we can always go and fix it.
Metrics is an auxiliary functionality. It shouldn't interfere with real processing any day.

versionedRouter.js
```diff
@@ -90,6 +93,38 @@ const functionsEnabled = () => {
   return areFunctionsEnabled === 1;
 };
 
+async function handleCdkV2(destName, parsedEvent, flowType) {
+  try {
+    const workflowEngine = getWorkflowEngine(destName);
```
koladilip, line 98: we need to use flowType to get specific workflow

koladilip (commented):
Looks good overall small minor comments.

versionedRouter.js
sanpj2292: It's not going to interfere the real processing actually. How does that happen is ?

We are going to call this handler for CDK V2 at versionedRouter level only and since while adding this handler elsewhere and testing it, an error will be thrown while dev testing only, which is when the fix also would happen. Any additional protection can be provided by adding unit-test cases

It also makes sense to may be to not throw an error but we don't have a context of the caller to know at which handler did the error actually occur. Without that context, it would be very difficult to debug. So instead have a default `errorAt` tag known as `UNKNOWN` may be to indicate that the error occurred from an unknown caller. This is more apt than having `PROC` as the default for `errorAt` tag

sanpj2292: `errorAt` currently serves as a pointer to the endpoint(more like handler of that endpoint) at which the error has occurred. I'm just making sure of the definition and how it is being used for alerting

saikumarrs: Added `UNKNOWN` type.

versionedRouter.js
sanpj2292: @saikumarrs Based on [this](https://github.com/rudderlabs/rudder-transformer/pull/1448#discussion_r984245041) what does `flowType` mean here ?

cdk/v2/handler.js
saikumarrs: Addressed.

versionedRouter.js
saikumarrs: Addressed.

saikumarrs: @sanpj2292 `flowType` === `errorAt`. I just reused what we already have.
I didn't want to use `errorAt` naming as it restricts its scope to just errors.

In the future, we should replace `errorAt` with `flowType` everywhere.

sanpj2292: Yeah, makes sense

sanpj2292 (approved):

PR #1469, opened 2022-10-10
feat(error reporting): integrate bugsnag

saikumarrs:
## Description of the change

Integrated Bugsnag error collection tool to notify handled and unhandled errors.
Presently, only the `versionedRouter.js` module is instrumented.

Note: Only the destination transformations and data delivery endpoints are instrumented for now.

## Type of change

## Related issues

N/A

### Development

- [x] Lint rules pass locally

### Code review 


util/errorNotifier/bugsnag.js
```diff
@@ -0,0 +1,40 @@
...
+function notify(err, context, metadata) {
+  if (context) Bugsnag.setContext(context);
+  if (metadata) Bugsnag.addMetadata("metadata", metadata);
+
+  Bugsnag.notify(err);
+
+  // Reset context and metadata that
+  // could've been set before
+  if (context) Bugsnag.setContext("");
+  if (metadata) Bugsnag.clearMetadata("metadata");
```
koladilip, line 34:
```suggestion
  Bugsnag.notify(error, function (event) {
    event.context = context;
    event.addMetadata("metadata", metadata);
  });
```

koladilip: Ref: https://docs.bugsnag.com/platforms/javascript/reporting-handled-errors/

versionedRouter.js
```diff
@@ -123,10 +150,32 @@ async function handleCdkV2(destName, parsedEvent, flowType) {
...
+      ...getReqMetadata()
+    });
+
+    return resp;
   }
 }
 
 async function handleDest(ctx, version, destination) {
+  const getReqMetadata = event => {
+    try {
```
koladilip, line 166: can we do this instead of wrapping in try catch?
```js
return {
         destType: destination,
         destinationId: event?.destination?.ID,
         destName: event?.destination?.Name,
         metadata: event?.metadata
       };
```

versionedRouter.js
```diff
@@ -650,6 +705,15 @@ if (transformerTestModeEnabled) {
 }
 
 async function handleSource(ctx, version, source) {
+  const getReqMetadata = () => {
+    try {
+      return { srcType: source };
```
koladilip, line 710: what are we expecting to go wrong here?

versionedRouter.js
```diff
@@ -718,6 +788,15 @@ if (startSourceTransformer) {
 }
 
 async function handleProxyRequest(destination, ctx) {
+  const getReqMetadata = () => {
+    try {
+      return { destType: destination };
```
koladilip, line 793: do we need try catch here?

versionedRouter.js
```diff
@@ -826,6 +910,22 @@ router.get("/features", ctx => {
 });
 
 const batchHandler = ctx => {
+  const getReqMetadata = destEvents => {
+    try {
+      const reqBody = ctx.request.body;
+      const firstEvent = destEvents[0];
+      return {
+        destType: reqBody.destType,
```
koladilip, line 918: can we use reqBody?.destType like these to avoid try catch?

versionedRouter.js
```diff
@@ -209,6 +258,12 @@ async function handleDest(ctx, version, destination) {
             ...errObj.statTags
           }
         };
+        errNotificationClient.notify(error, "Destination Transformation", {
+          ...resp,
+          ...getCommonMetadata(ctx),
+          ...getReqMetadata(event)
+        });
+        return resp;
```
utsabc, line 266: Same query as above

versionedRouter.js
saikumarrs: I wrapped all the logic related to error notification in try-catches even though unnecessary for most of the cases as I didn't want anything going wrong to affect actual processing. For the same reason, the instrumentation is usually the last thing in the catch blocks.

koladilip (approved):

utsabc (approved):

versionedRouter.js
```diff
@@ -871,6 +976,16 @@ router.post("/batch", ctx => {
 });
 
 const fileUpload = async ctx => {
+  const getReqMetadata = () => {
```
sanpj2292, line 979: Why do we need a callback here ? we already have `destType` in Ln 989, which can be used where `getReqMetadata` method is called right ?

sanpj2292: Where there is a need of firstEvent kind of thing, it probably makes sense to have a callback. But in methods like `handleProxyRequest`, `pollStatus`, `fileUpload` etc,. since the variable `destType` is directly available, it looks unnecessary.

saikumarrs: I made the implementation generic across all the endpoint handlers. All the metadata should be served from here.

koladilip (approved):

versionedRouter.js
```diff
@@ -99,9 +100,35 @@ const functionsEnabled = () => {
...
+  };
+}
+
+async function handleCdkV2(destType, parsedEvent, flowType) {
+  const getReqMetadata = () => {
+    try {
+      return {
+        flowType,
+        destType,
+        destinationId: parsedEvent.destination.ID,
```
koladilip, line 120: as we discussed change to parsedEvent.destination?.ID

utsabc (approved):

versionedRouter.js
saikumarrs: Addressed.

koladilip (approved):

versionedRouter.js
```diff
@@ -214,9 +191,9 @@ async function handleDest(ctx, version, destination) {
     try {
       return {
         destType: destination,
-        destinationId: event.destination.ID,
-        destName: event.destination.Name,
-        metadata: event.metadata
+        destinationId: event?.destination?.ID,
+        destName: event?.destination?.Name,
+        metadata: event?.metadata
```
utsabc, line 196: First level optional chaining required?

saikumarrs: Yes. We are being very conservative.

koladilip (approved):

utsabc (approved):

PR #1484, opened 2022-10-14
feat(marketo): adding cache eviction and dynamic ttl support

ujjwal-ab:
## Description of the change

> Adding Cache eviction in case of authentication failure due to invalid/expired token and dynamic `ttl` support on the basis of the age of the token received in the request

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


versionedRouter.js
```diff
@@ -18,10 +19,10 @@ const {
   isHttpStatusSuccess,
   getErrorRespEvents,
   isCdkDestination,
-  getErrorStatusCode
-} = require("./v0/util");
+  getErrorStatusCode,
+  getDestHandler
+} = v0Utils;
```
saikumarrs, line 24: Rather than destructuring, directly use the utils object to invoke the methods.

v0/util/index.js
```diff
@@ -1732,6 +1732,37 @@ const flattenMultilevelPayload = payload => {
...
+ * Obtain the authCache instance used to store the access token information to send/get information to/from destination
+ * @param {string} destType destination name
+ * @returns {Cache | undefined} The instance of "v0/util/cache.js"
+ */
+const getDestAuthCacheInstance = destType => {
+  const destInf = getDestHandler("v0", destType);
+  if (destInf) {
+    return destInf?.authCache;
+  }
+  return {};
```
saikumarrs, line 1763:
```suggestion
  return destInf?.authCache || {};
```

v0/util/index.js
```diff
@@ -1732,6 +1732,37 @@ const flattenMultilevelPayload = payload => {
...
+ * @returns
+ *  The transform.js instance used for destination transformation
+ */
+const getDestHandler = (_version, dest) => {
+  if (DestHandlerMap.hasOwnProperty(dest)) {
+    // eslint-disable-next-line import/no-dynamic-require, global-require
+    return require(`../destinations/${DestHandlerMap[dest]}/transform`);
+  }
+  // eslint-disable-next-line import/no-dynamic-require, global-require
+  return require(`../destinations/${dest}/transform`);
```
saikumarrs, line 1750:
```suggestion
  const destName = DestHandlerMap[dest] || dest;
  // eslint-disable-next-line import/no-dynamic-require, global-require
  return require(`../destinations/${destName}/transform`);
```

v0/util/index.js
```diff
@@ -1732,6 +1732,37 @@ const flattenMultilevelPayload = payload => {
...
+/**
+ * Gets the destintion's transform.js file used for transformation
+ * **Note**: The transform.js file is imported from
+ *  `v0/destinations/${dest}/transform`
+ * @param {*} _version -> version for the transfor
+ * @param {*} dest destination name
+ * @returns
+ *  The transform.js instance used for destination transformation
+ */
+const getDestHandler = (_version, dest) => {
```
saikumarrs, line 1744: There is no need for version here as it's a v0 util method.

v0/util/cache.js
```diff
@@ -17,12 +17,20 @@ class Cache {
 
     return storeFunction().then(result => {
       // store in cache if the value is valid, else skip
+      if (result && typeof result === "object") {
+        this.cache.set(key, result.token, result.age);
+        return result.token;
+      }
```
saikumarrs, line 23: This logic should not be specific to tokens.

v0/destinations/marketo/transform.js
```diff
@@ -547,7 +559,7 @@ const processEvent = async (message, destination, token) => {
   }
 
   // wrap response for router processing
-  return responseWrapper(response);
+  return responseWrapper(response, destination.ID);
```
saikumarrs, line 562: The `responseWrapper` function takes authKey as the second argument.

v0/destinations/marketo/transform.js
```diff
@@ -499,11 +507,15 @@ const processTrack = async (message, formattedDestination, token) => {
...
-const responseWrapper = response => {
+const responseWrapper = (response, authKey) => {
   const resp = defaultRequestConfig();
   resp.endpoint = response.endPoint;
   resp.method = defaultPostRequestConfig.requestMethod;
-  resp.headers = { ...response.headers, "Content-Type": "application/json" };
+  resp.headers = {
+    ...response.headers,
+    "Content-Type": "application/json",
+    "x-cache-key": authKey
```
saikumarrs, line 517: What's the need for this?

versionedRouter.js
```diff
@@ -784,8 +778,9 @@ if (startSourceTransformer) {
   });
 }
 
-async function handleProxyRequest(destination, ctx) {
+const handleProxyRequest = async (destination, ctx) => {
   const destinationRequest = ctx.request.body;
+  const authCache = v0Utils.getDestAuthCacheInstance(destination);
```
saikumarrs, line 783: This handler shouldn't worry about the auth cache, as it's not a standard for all the proxy destinations (including the future).

We can take two approaches:
1. Handle the cache within the response handler.
2. Set the callback function to the response handler that handles post-response transformation actions (in this case, cache eviction based on the error type deduction).

saikumarrs (changes requested):

saikumarrs:
@ujjwal-ab @sanpj2292 Please update the PR description with relevant details.

v0/destinations/marketo/transform.js
sanpj2292: You need to basically know what is the key used to store information in cache. And in proxy or response handler for that matter you wouldn't have access to the key. Hence we are sending the information as part of transformed payload. Please let me know if it makes sense

v0/destinations/marketo/transform.js
sanpj2292: Yes, here `destination.ID` is the `authKey`

v0/util/cache.js
ujjwal-ab: Yes, it can be used for any data not specifically tokens. Will update the namings to avoid confusion.

versionedRouter.js
sanpj2292: We don't want to make changes in other parts unless it is absolutely necessary

saikumarrs: Let's not have two different ways of accessing the utils in the versioned router, at least. It sets a bad precedent for future developers.

v0/util/index.js
ujjwal-ab: Updated

v0/util/index.js
ujjwal-ab: Updated

v0/util/cache.js
```diff
@@ -17,12 +17,20 @@ class Cache {
 
     return storeFunction().then(result => {
       // store in cache if the value is valid, else skip
+      if (result && typeof result === "object") {
```
saikumarrs, line 20: This approach fails if we were to store objects instead of primitive values in the cache, it'd fail.
Update the signature of the store function to take two values, namely, `value` and `age` (optional).

ujjwal-ab: Okay, but, why would it fail with the current implementation? As we can store objects also as a value here.

saikumarrs: It'd fail if the `result` is an objection and doesn't follow the structure as expected in Ln 20-22.

PR #1488, opened 2022-10-14
feat: added multitopic support for kafka

utsabc:
## Description of the change

Added support for multiple topic routing for kafka destination

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


v0/destinations/kafka/transform.js
```diff
@@ -18,19 +19,39 @@ function batch(destEvents) {
...
+ * @param {*} input
+ * @returns {*} metadata
+ */
+function processMetadata(input) {
+  const { metadata, inputEvent } = input;
+  const clonedMetadata = cloneDeep(metadata);
+  const integrationsObj = getIntegrationsObj(inputEvent.message, "kafka");
+  const { topic } = integrationsObj || {};
+  if (topic) {
+    clonedMetadata.rudderId = `${clonedMetadata.rudderId}<<>>${topic}`;
```
saikumarrs, line 52: `<<>>` is there any significance to this separator?

utsabc: its for keeping parity no significant meaning
original rudderId is stored in this format `anonymousId<<>>userId`

v0/destinations/kafka/transform.js
```diff
@@ -18,19 +19,39 @@ function batch(destEvents) {
...
+/**
+ * This functions takes event matadata and updates it based on the transformed and raw paylaod
+ * @param {*} input
+ * @returns {*} metadata
+ */
+function processMetadata(input) {
+  const { metadata, inputEvent } = input;
+  const clonedMetadata = cloneDeep(metadata);
+  const integrationsObj = getIntegrationsObj(inputEvent.message, "kafka");
+  const { topic } = integrationsObj || {};
```
krishna2020, line 50: As we don't have hold of destConfig in processorTransform, we can't set default topic here. So at the server end, do the error signify the failure is due to topic missing for the event ?

v0/destinations/kafka/transform.js
```diff
@@ -18,19 +19,39 @@ function batch(destEvents) {
 }
 
 function process(event) {
-  const integrationsObj = getIntegrationsObj(event.message, "kafka");
-  const { schemaId } = integrationsObj || {};
+  const { message } = event;
+  const integrationsObj = getIntegrationsObj(message, "kafka");
```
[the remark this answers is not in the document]
krishna2020: For that matter, we can't even consider exposing eventName/eventType mapping in UI.

v0/destinations/kafka/transform.js
```diff
@@ -18,19 +19,39 @@ function batch(destEvents) {
 }
 
 function process(event) {
-  const integrationsObj = getIntegrationsObj(event.message, "kafka");
-  const { schemaId } = integrationsObj || {};
+  const { message } = event;
```
krishna2020, line 22: Post passing the destConfig in event, we can have the entire handling here:
If conf.MultiTopicSupport=false, take destConf.Topic
Else, getIntegrationsObj/event mapping based to identify the topic & fallback to default destConf.Topic

v0/destinations/kafka/transform.js
```diff
@@ -1,3 +1,4 @@
+const cloneDeep = require("lodash/cloneDeep");
 const { getIntegrationsObj } = require("../../util");
 
 function batch(destEvents) {
```
krishna2020, line 4: batch logic needs to be handled separately - internally it's not calling process(event)

v0/destinations/kafka/transform.js
```diff
@@ -1,36 +1,74 @@
...
-    batchedRequest.push(event.message);
-  });
-  respList.push({
-    batchedRequest,
-    metadata,
-    destination: destEvents[0].destination
-  });
+
+  // Grouping the events by topic
+  const groupedEvents = destEvents.reduce((acc, event) => {
```
koladilip, line 9: is there lodash groupBy functionality we can leverage ?

v0/destinations/kafka/transform.js
```diff
@@ -1,36 +1,74 @@
...
 
   return respList;
 }
 
 function process(event) {
-  const integrationsObj = getIntegrationsObj(event.message, "kafka");
+  const { message, destination } = event;
+  const integrationsObj = getIntegrationsObj(message, "kafka");
   const { schemaId } = integrationsObj || {};
+  const topic = integrationsObj?.topic || destination.Config.topic;
```
koladilip, line 41: as discussed, if topic is empty then throw 400 error

v0/destinations/kafka/transform.js
```diff
@@ -1,36 +1,74 @@
...
+    }
+    return acc;
+  }, {});
+
+  // Creating a batched request for each topic
+  // we are grouping the events based on topics
+  // and creating a batched request for each topic
+  // example: input events = [{event1,topic1},{event2,topic1},{event3,topic2}]
+  // out from transformer:  {batchedRequest:[{event1},{event2}]}, {batchedRequest:[{event3}]} (2 multilexed responses)
+  for (const [events] of Object.entries(groupedEvents)) {
```
koladilip, line 24: Object.values is sufficient right as we are not using any keys

versionedRouter.js
```diff
@@ -240,7 +240,13 @@ async function handleDest(ctx, version, destination) {
 
             return {
               output: { ...ev, userId },
-              metadata: event.metadata,
+              metadata: destHandler.processMetadata
+                ? destHandler.processMetadata({
+                    metadata: event.metadata,
+                    inputEvent: parsedEvent,
+                    outPutEvent: ev
```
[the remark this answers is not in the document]
utsabc: This is addressed

v0/destinations/kafka/transform.js
utsabc: Handled

v0/destinations/kafka/transform.js
utsabc: Yes addressed

versionedRouter.js
```diff
@@ -295,7 +295,13 @@ async function handleDest(ctx, version, destination) {
...
             return {
               output: { ...ev, userId },
-              metadata: event.metadata,
+              metadata: destHandler.processMetadata
+                ? destHandler.processMetadata({
+                    metadata: event.metadata,
+                    inputEvent: parsedEvent,
+                    outputEvent: ev
+                  })
+                : event.metadata,
```
saikumarrs, line 304: Why can't we manipulate the `metadata` within the `process` method itself? We have access to all the data there.

v0/destinations/kafka/transform.js
```diff
@@ -1,36 +1,83 @@
...
-    batchedRequest.push(event.message);
-  });
-  respList.push({
-    batchedRequest,
-    metadata,
-    destination: destEvents[0].destination
+
+  // Grouping the events by topic
+  const groupedEvents = groupBy(destEvents, event => {
+    return event.message.topic;
```
saikumarrs, line 14: Here the topic is taken from the message where it is picked from the integrations object in the `process` function. Is that a mistake? If no, what's the reason?

__tests__/data/kafka_batch_input.json
```diff
@@ -22,7 +22,8 @@
         "request_ip": "[::1]",
         "anonymousId": "anon-id-new",
         "originalTimestamp": "2021-07-08T02:45:11.329+05:30"
-      }
+      },
```
krishna2020, line 25: Add/update a test case with 2 different topics

v0/destinations/kafka/transform.js
```diff
@@ -1,36 +1,83 @@
...
+  // example: input events = [{event1,topic1},{event2,topic1},{event3,topic2}]
+  // out from transformer:  {batchedRequest:[{event1},{event2}]}, {batchedRequest:[{event3}]} (2 multilexed responses)
+  for (const events of Object.values(groupedEvents)) {
+    const response = {};
+    response.batchedRequest = events.map(event => event.message);
+    response.metadata = events.map(event => event.metadata);
+    response.destination = events[0].destination;
+    respList.push(response);
+  }
+
```
krishna2020, line 29: If respList has size as 2, will server store these 2 as different jobId's for router ?

utsabc: So it will be handled like this
Resp1 = [e1,e2,e3], Resp2=[e4,e5]
No these each events metadata will tracked in rt table (jobid is a property in metadata)

in produce if batching is enabled resp1 will be processed if it fails [e1,e2,e3] all jobs fail

krishna2020: Ok each event will have jobId in rt_jobs table. But during produce, Resp1 & Resp2 will be treated as two separate requests for delivery.

v0/destinations/kafka/transform.js
koladilip: This is batch api @saikumarrs so by then it is already processed so topic is now available in message.

v0/destinations/kafka/transform.js
koladilip: so even in case of batching we won't be creating new jobID in rt_jobs right?

v0/destinations/kafka/transform.js
```diff
@@ -1,36 +1,83 @@
...
   return respList;
 }
 
 function process(event) {
-  const integrationsObj = getIntegrationsObj(event.message, "kafka");
+  const { message, destination } = event;
+  const integrationsObj = getIntegrationsObj(message, "kafka");
   const { schemaId } = integrationsObj || {};
+  const topic = integrationsObj?.topic || destination.Config.topic;
```
koladilip, line 37: is destination.Config.topic going to be dynamic field like {{}} this?

v0/destinations/kafka/transform.js
utsabc: No we are handling this as discussed

versionedRouter.js
utsabc: No it will break the contact between server and transformer if you check carefully then we have to pass updated `metadata` in `events` array which will break the structure of `output`

__tests__/data/kafka_batch_input.json
utsabc: Updated

versionedRouter.js
saikumarrs: How does it break the data contract? You'd send `metadata` for each event processed, and the value of `rudderId` is only updated now.

v0/destinations/kafka/transform.js
utsabc: That would be handled before this call is made at versionedRouter

v0/destinations/kafka/transform.js
utsabc: Correct

versionedRouter.js
utsabc: Currently we don't sent metadata as part of the processed output from `process` function

utsabc:
```
 output: { ...ev, userId }
```
each of the events `ev`s would then contain `metadata`

v0/destinations/kafka/transform.js
utsabc: Addressed

versionedRouter.js
saikumarrs: We don't have to return anything from the process function. As you have the same metadata object reference in the function, updating it would be sufficient.

utsabc: That would lead to some level of abstraction and unwanted hiding of context. Then `process` function implicitly has a side effect i.e it does something more (mutates) state of an object which it isn't supposed to

saikumarrs: I see your point.

v0/destinations/kafka/transform.js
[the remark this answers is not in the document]
krishna2020: This will be resolved with https://github.com/rudderlabs/rudder-server/pull/2625#discussion_r1007865366

krishna2020 (approved):
We still have one more feature to implement on taking eventName/eventType wise topic mapping from FE configuration.

koladilip (approved):

krishna2020 (approved):

koladilip (approved):

saikumarrs (approved):

PR #1496, opened 2022-10-18
fix: posthog group update

sauravlal-rudderstack:
## Description of the change

> posthog group update

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


v0/destinations/posthog/transform.js
```diff
@@ -88,11 +89,24 @@ const responseBuilderSimple = (message, category, destination) => {
...
+  // This will add the attributes $groups, which will associate the group with the user.
+  if (category.type === CONFIG_CATEGORIES.GROUP.type && payload.properties) {
+    const groupType = get(payload, "properties.$group_type");
+    const groupKey = get(payload, "properties.$group_key");
+    if (groupType && groupKey) {
+      payload.properties.$groups = {
+        [groupType]: groupKey
+      };
+    }
+  }
```
ItsSudip, line 109:
```suggestion
if (category.type === CONFIG_CATEGORIES.GROUP.type) {
    // This is to ensure groupType delete from $group_set, as it is properly mapped
    // in 'properties.$group_type'.
    if (destination.Config.useV2Group) {
      delete payload?.properties?.$group_set?.groupType;
    }
    // This will add the attributes $groups, which will associate the group with the user.
    if (payload.properties) {
      const groupType = get(payload, "properties.$group_type");
      const groupKey = get(payload, "properties.$group_key");
      if (groupType && groupKey) {
        payload.properties.$groups = {
          [groupType]: groupKey
        };
      }
    }
  }
```

ItsSudip (changes requested):

v0/destinations/posthog/transform.js
sauravlal-rudderstack: updated

v0/destinations/posthog/transform.js
```diff
@@ -88,10 +89,22 @@ const responseBuilderSimple = (message, category, destination) => {
...
+    }
+    // This will add the attributes $groups, which will associate the group with the user.
+    if (payload.properties) {
+      const groupType = get(payload, "properties.$group_type");
+      const groupKey = get(payload, "properties.$group_key");
+      if (groupType && groupKey) {
+        payload.properties.$groups = {
+          [groupType]: groupKey
+        };
+      }
```
ItsSudip, line 106: As both of these are required fields we should make `groupId` and `traits.groupType` required in config file.

sauravlal-rudderstack: Posthog doesn't throw an error through API if we are not passing the groupId and groupType. And also it successfully shows events without these.

ItsSudip: What is the purpose of sending group calls then? Customers will get confused as they will see successful group calls but are unable to add users to the group right?

sauravlal-rudderstack: Can't we mention this in docs, that Posthog doesn't throw errors, Is it good that we should throw errors as Posthog doesn't do that? 
We can clearly mention in docs that `groupKey` and `groupType` is required for creating/updating a group. It will not throw an error if the user does not pass it. I'm not sure about the use case without these fields, but we also need to think about backward compatibility. LMKYourThought?

ItsSudip: If we make the call without `groupKey` does the event show up in the dashboard?

sauravlal-rudderstack: Yes, it will show.

utsabc (approved):

PR #1519, opened 2022-10-27
feat(marketo): add dynamic ttl & cache eviction support

ujjwal-ab:
## Description of the change

Adding Cache eviction in case of authentication failure due to invalid/expired token and dynamic ttl support on the basis of the age of the token received in the request
## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


versionedRouter.js
```diff
@@ -878,10 +879,11 @@ async function handleProxyRequest(destination, ctx) {
     return {};
   };
 
-  const destinationRequest = ctx.request.body;
+  const { metadata, destinationRequest } = ctx.request.body;
```
saikumarrs, line 882: This would be a breaking change. We need to make these changes backward compatible with previous server versions.
CC @sanpj2292

versionedRouter.js
```diff
@@ -878,10 +879,11 @@ async function handleProxyRequest(destination, ctx) {
...
   const destNetworkHandler = networkHandlerFactory.getNetworkHandler(
     destination
   );
+  const authCache = v0Utils.getDestAuthCacheInstance(destination);
```
saikumarrs, line 886: https://github.com/rudderlabs/rudder-transformer/pull/1484#discussion_r1001360629

versionedRouter.js
```diff
@@ -905,7 +907,9 @@ async function handleProxyRequest(destination, ctx) {
     });
     response = destNetworkHandler.responseHandler(
       processedProxyResponse,
-      destination
+      destination,
+      metadata.destInfo,
```
saikumarrs, line 911: Even if we were to take this approach, destructuring the metadata should happen within the response handler.

v0/util/cache.js
```diff
@@ -17,12 +17,21 @@ class Cache {
...
       // store in cache if the value is valid, else skip
+      if (result && typeof result === "object") {
+        if (result.value && result.age) {
+          return result.value;
+        }
+      }
       if (result) {
         this.cache.set(key, result);
       }
       return result;
```
saikumarrs, line 28: The logic is flawed.

```suggestion
      let retVal = result;
      if (result) {
        if (typeof result === "object" and "value" in result and "age" in result) {
          this.cache.set(key, result.value, result.age);
          retVal = result.value;
        } else {
          this.cache.set(key result);
        }
      }
      return retVal;
```

saikumarrs (changes requested):
Partial review.

versionedRouter.js
sanpj2292: @ujjwal-ab Let's send the metadata information as part of the destinationRequest itself, so that the backward compatibility is maintained

versionedRouter.js
sanpj2292: This can be moved to be inside of the responseHandler itself. Can you once verify ? @ujjwal-ab

versionedRouter.js
```diff
@@ -915,7 +915,7 @@ async function handleProxyRequest(destination, ctx) {
       destination
     });
     response = destNetworkHandler.responseHandler(
-      processedProxyResponse,
+      { ...processedProxyResponse, metadata },
```
saikumarrs, line 918: To avoid collisions, we should not merge metadata into the proxy response.

saikumarrs: As it is internal, we don't have to worry about backward compatibility.

v0/util/cache.js
```diff
@@ -17,12 +17,22 @@ class Cache {
 
     return storeFunction().then(result => {
       // store in cache if the value is valid, else skip
+      let retVal = result;
       if (result) {
-        this.cache.set(key, result);
+        if (typeof result === "object" && result.value && result.age) {
```
saikumarrs, line 22: This is incorrect. It'll fail for cases where `value` is falsy (`false`, `0`, "", etc.)

versionedRouter.js
saikumarrs: You can send `metadata` (`reqMetadata`) as a third optional argument to this function.

v0/destinations/marketo/util.js
```diff
@@ -104,6 +111,18 @@ const marketoResponseHandler = (destResponse, sourceMessage, stage) => {
...
     // marketo application level failure
     if (response && !response.success) {
+      if (response.errors && destInfo) {
+        const { authKey } = destInfo;
+        if (
+          response.errors.some(errorObj => {
+            return errorObj.code === "601" || errorObj.code === "602";
+          })
+        ) {
+          if (authCache && authKey) {
```
saikumarrs, line 121: Do this check early.

v0/destinations/marketo/util.js
```diff
@@ -77,7 +78,13 @@ const marketoApplicationErrorHandler = (
   }
 };
 
-const marketoResponseHandler = (destResponse, sourceMessage, stage) => {
+const marketoResponseHandler = (
+  destResponse,
+  sourceMessage,
+  stage,
+  destInfo,
```
saikumarrs, line 85: We should call it `destReqMetadata` or simply `reqMetadata`.

v0/destinations/marketo/util.js
```diff
@@ -104,6 +111,18 @@ const marketoResponseHandler = (destResponse, sourceMessage, stage) => {
     }
     // marketo application level failure
     if (response && !response.success) {
+      if (response.errors && destInfo) {
```
saikumarrs, line 114: Add some documentation for this snippet of code.

v0/destinations/marketo/transform.js
```diff
@@ -611,6 +619,7 @@ const processRouterDest = async inputs => {
   const respList = await Promise.all(
     inputs.map(async input => {
       try {
+        input.metadata.destInfo = { authKey: input.destination.ID };
```
saikumarrs, line 622: This would be messy, as @utsabc noted [here](https://github.com/rudderlabs/rudder-transformer/pull/1488#discussion_r1007847457).
We should probably wait for this PR to get merged.

v0/destinations/marketo/transform.js
```diff
@@ -70,11 +70,13 @@ const getAuthToken = async formattedDestination => {
     const data = marketoResponseHandler(
       clientResponse,
       "During fetching auth token",
-      TRANSFORMER_METRIC.TRANSFORMER_STAGE.TRANSFORM
+      TRANSFORMER_METRIC.TRANSFORMER_STAGE.TRANSFORM,
+      authCache,
+      formattedDestination.ID
```
saikumarrs, line 75: This is the token fetch request. It should not lead to cache eviction, as the token is definitely not present in the cache.

v0/destinations/marketo/transform.js
```diff
@@ -197,7 +203,9 @@ const lookupLeadUsingId = async (
     const data = marketoResponseHandler(
       clientResponse,
       "[Marketo Transformer]: During lead look up using userId",
-      TRANSFORMER_METRIC.TRANSFORMER_STAGE.TRANSFORM
+      TRANSFORMER_METRIC.TRANSFORMER_STAGE.TRANSFORM,
+      authCache,
+      formattedDestination.ID
```
saikumarrs, line 208: Rather than scattering the auth cache key in multiple places in the code, we can abstract it as a function. So, there is only one place we need to change if we ever wish to change the key value.

saikumarrs (changes requested):

versionedRouter.js
sanpj2292: Why are you saying there will be collisions ?

saikumarrs: We can't be sure that `processedProxyResponse` would not contain a field named `metadata`. It's a preemptive measure.

v0/destinations/marketo/util.js
sanpj2292: It is `destInfo` that is being sent from `responseHandler` function. I don't think there is a need to change the name

versionedRouter.js
sanpj2292: Then we can put another named key `rudderJobMetadata` instead of `metadata` and proceed with the same implementation(as in the changes)
```javascript
response = destNetworkHandler.responseHandler(
      processedProxyResponse,
      { ...processedProxyResponse, rudderJobMetadata: metadata },
      ...);
```

v0/destinations/marketo/util.js
saikumarrs: I don't want to debate further on the naming, but in general, it's metadata IMO, which should only be destructured at the last node of usage.

sanpj2292: > which should only be destructured at the last node of usage.

Yes. As I found out about `marketoResponseHandler` being used for parsing intermediate API calls. @ujjwal-ab can you make relevant changes ?

ujjwal-ab: We will have to destructure the metadata in the `responseHandler` itself as we are using the same `marketoResponseHandler` function for cache eviction for internal lookup call failure and proxy both so we will have to send the same params in both cases as we are passing the `authCache` and `authKey` in case of internal lookups.

saikumarrs: Yeah, you can wrap up that data in other places accordingly. The idea is to stick to a generic format for passing the data across the layers.

ujjwal-ab: Should we make the internal lookups format in the proxy format?

saikumarrs: Yes, you can wrap it under the veil of metadata.

v0/destinations/marketo/transform.js
ujjwal-ab: Can these parameters be abstracted as the `destinationId` is not accessible globally here?

saikumarrs: Yeah, I'd send the whole input event to that new function and let it figure out and return the auth cache key.

ujjwal-ab: Updated it.

v0/destinations/marketo/transform.js
saikumarrs: @ujjwal-ab can you address this?

ujjwal-ab: Ya, it's addressed.

v0/util/cache.js
```diff
@@ -17,12 +17,26 @@ class Cache {
 
     return storeFunction().then(result => {
```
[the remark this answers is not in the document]
ujjwal-ab: In that case, it will fail. But it is structured in that way only like if the value is not in the cache then this function is being used with some implementations to get the value and store it.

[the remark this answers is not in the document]
ujjwal-ab: Updated it.

v0/util/cache.js
```diff
@@ -9,19 +9,28 @@ class Cache {
...
+  async get(key, storeFunction) {
     const value = this.cache.get(key);
     if (value) {
       return Promise.resolve(value);
     }
 
-    return storeFunction().then(result => {
-      // store in cache if the value is valid, else skip
-      if (result) {
+    const result = await storeFunction();
```
koladilip, line 18:
```suggestion
    const result = storeFunction ? await storeFunction() : undefined;
```

ujjwal-ab: Updated

koladilip (approved):

saikumarrs (approved):

versionedRouter.js
```diff
@@ -425,6 +425,10 @@ async function routerHandleDest(ctx) {
   try {
     const { input } = ctx.request.body;
     destType = ctx.request.body.destType;
+    const routerDestHandler = getDestHandler("v0", destType);
+    input[0].metadata = routerDestHandler.processMetadataForRouter
+      ? routerDestHandler.processMetadataForRouter(input[0])
+      : input[0].metadata;
```
utsabc, line 431: This should be called for all the processed events. Once the router transformation is complete for all the transformed event present in `listOutput` each metadata should be updated.

ujjwal-ab: Updated it.

saikumarrs (approved):

sanpj2292 (approved):

utsabc (approved):

PR #1536, opened 2022-11-01
fix: fixing errors caught by bugsnag

ItsSudip:
## Description of the change

Fixing errors caught in bugsnag.

## Type of change
- [X] Bug fix (non-breaking change that fixes an issue)

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/v0/destinations/facebook_pixel/transform.js
+++ b/v0/destinations/facebook_pixel/transform.js
@@ -588,6 +588,17 @@ const processEvent = (message, destination) => {
           })
           .build();
       }
+      if (typeof message.event !== "string") {
+        throw new ErrorBuilder()
+          .setMessage("event name should be string")
+          .setStatus(400)
+          .setStatTags({
+            ...statTags,
+            meta:
+              TRANSFORMER_METRIC.MEASUREMENT_TYPE.TRANSFORMATION.META.BAD_EVENT
+          })
+          .build();
+      }
       standard = eventsToEvents;
       if (standard) {
         standardTo = standard.reduce((filtered, standards) => {
--- a/v0/destinations/facebook_pixel/utils.js
+++ b/v0/destinations/facebook_pixel/utils.js
@@ -238,7 +238,7 @@ const deduceFbcParam = message => {
   try {
     parseUrl = new URL(url);
   } catch {
-    throw new CustomError("url is not in correct format", 400);
+    return undefined;
   }
   const paramsList = new URLSearchParams(parseUrl.search);
   const fbclid = paramsList.get("fbclid");
--- a/v0/destinations/facebook_pixel/utils.js
+++ b/v0/destinations/facebook_pixel/utils.js
@@ -32,16 +32,16 @@ const formatRevenue = revenue => {
 };
 
 /**
-*
-* @param {*} message Rudder Payload
-* @param {*} defaultValue product / product_group
-* @param {*} categoryToContent [ { from: 'clothing', to: 'product' } ]
-*
-* We will be mapping properties.category to user provided content else taking the default value as per ecomm spec
-* If category is clothing it will be set to ["product"]
-* @return Content Type array as defined in:
-* - https://developers.facebook.com/docs/facebook-pixel/reference/#object-properties
-*/
+ *
+ * @param {*} message Rudder Payload
+ * @param {*} defaultValue product / product_group
+ * @param {*} categoryToContent [ { from: 'clothing', to: 'product' } ]
+ *
+ * We will be mapping properties.category to user provided content else taking the default value as per ecomm spec
+ * If category is clothing it will be set to ["product"]
+ * @return Content Type array as defined in:
+ * - https://developers.facebook.com/docs/facebook-pixel/reference/#object-properties
+ */
 const getContentType = (message, defaultValue, categoryToContent) => {
   const { integrations } = message;
   if (
@@ -214,31 +214,35 @@ const transformedPayloadData = (
   return customData;
 };
 
-
 /**
- * 
- * @param {*} message 
+ *
+ * @param {*} message
  * @returns fbc parameter which is a combined string of the parameters below
- * 
+ *
  * version : "fb" (default)
- * 
+ *
  * subdomainIndex : 1 ( recommended by facebook, as well as our JS SDK sets cookies on the main domain, i.e "facebook.com")
- * 
+ *
  * creationTime : mapped to originalTimestamp converted in miliseconds
- * 
+ *
  * fbclid : deduced query paramter from context.page.url
- * 
+ *
  * ref: https://developers.facebook.com/docs/marketing-api/conversions-api/parameters/fbp-and-fbc#fbc
  */
-const deduceFbcParam = (message) => {
-
+const deduceFbcParam = message => {
   const url = message.context?.page?.url;
   if (!url) {
     return undefined;
   }
-  let parseUrl = new URL(url)
-  let paramsList = new URLSearchParams(parseUrl.search);
-  const fbclid = paramsList.get('fbclid')
+  const URLRegex = new RegExp(
+    "^(?:http(s)?:\\/\\/)?[\\w.-]+(?:\\.[\\w\\.-]+)+[\\w\\-\\._~:/?#[\\]@!\\$&'\\(\\)\\*\\+,;=.]+$"
+  );
+  if (!url.match(URLRegex)) {
+    throw new CustomError("url is not in correct format", 400);
+  }
+  const parseUrl = new URL(url);
+  const paramsList = new URLSearchParams(parseUrl.search);
+  const fbclid = paramsList.get("fbclid");
 
   if (!fbclid) {
     return undefined;
--- a/v0/destinations/facebook_pixel/utils.js
+++ b/v0/destinations/facebook_pixel/utils.js
@@ -234,13 +234,12 @@ const deduceFbcParam = message => {
   if (!url) {
     return undefined;
   }
-  const URLRegex = new RegExp(
-    "^(?:http(s)?:\\/\\/)?[\\w.-]+(?:\\.[\\w\\.-]+)+[\\w\\-\\._~:/?#[\\]@!\\$&'\\(\\)\\*\\+,;=.]+$"
-  );
-  if (!url.match(URLRegex)) {
+  let parseUrl;
+  try {
+    parseUrl = new URL(url);
+  } catch {
     throw new CustomError("url is not in correct format", 400);
   }
-  const parseUrl = new URL(url);
   const paramsList = new URLSearchParams(parseUrl.search);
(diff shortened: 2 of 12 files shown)


v0/destinations/facebook_pixel/utils.js
```diff
@@ -214,31 +214,35 @@ const transformedPayloadData = (
...
   const url = message.context?.page?.url;
   if (!url) {
     return undefined;
   }
-  let parseUrl = new URL(url)
-  let paramsList = new URLSearchParams(parseUrl.search);
-  const fbclid = paramsList.get('fbclid')
+  const URLRegex = new RegExp(
+    "^(?:http(s)?:\\/\\/)?[\\w.-]+(?:\\.[\\w\\.-]+)+[\\w\\-\\._~:/?#[\\]@!\\$&'\\(\\)\\*\\+,;=.]+$"
+  );
```
utsabc, line 239: A brief about what does this regex test positive to can help. Check `kustomer/util.js` as reference

v0/util/index.js
```diff
@@ -1546,7 +1546,10 @@ const isOAuthSupported = (destination, destHandler) => {
 
 function isAppleFamily(platform) {
   const appleOsNames = ["ios", "watchos", "ipados", "tvos"];
-  return appleOsNames.includes(platform?.toLowerCase());
+  if (platform) {
+    return appleOsNames.includes(platform?.toLowerCase());
+  }
+  throw new CustomError("Platform is not defined", 400);
```
utsabc, line 1552: This function is having a side effect the responsibility of validating the existence of platform should be by caller

v0/destinations/facebook_pixel/utils.js
```diff
@@ -214,31 +214,35 @@ const transformedPayloadData = (
...
-
+const deduceFbcParam = message => {
   const url = message.context?.page?.url;
   if (!url) {
     return undefined;
   }
-  let parseUrl = new URL(url)
-  let paramsList = new URLSearchParams(parseUrl.search);
-  const fbclid = paramsList.get('fbclid')
+  const URLRegex = new RegExp(
```
saikumarrs, line 237: Rather than taking this pain, we can just handle the error from URL class.

v0/destinations/zendesk/transform.js
```diff
@@ -151,6 +151,9 @@ async function getUserId(message, headers, type) {
       ? get(message, "context.traits")
       : getFieldValueFromMessage(message, "traits");
   const userEmail = traits.email;
```
saikumarrs, line 153: The issue here is that traits can be undefined.

v0/util/index.js
```diff
@@ -1546,7 +1546,10 @@ const isOAuthSupported = (destination, destHandler) => {
 
 function isAppleFamily(platform) {
   const appleOsNames = ["ios", "watchos", "ipados", "tvos"];
-  return appleOsNames.includes(platform?.toLowerCase());
+  if (platform) {
```
saikumarrs, line 1549: platform must be a string for this to be successful.

saikumarrs (changes requested):

v0/destinations/facebook_pixel/utils.js
ItsSudip: Removed this part. Handled through URL class.

v0/util/index.js
ItsSudip: We are checking if platform is string or not.

v0/destinations/facebook_pixel/utils.js
ItsSudip: Updated.

v0/util/index.js
```diff
@@ -1546,7 +1546,10 @@ const isOAuthSupported = (destination, destHandler) => {
 
 function isAppleFamily(platform) {
   const appleOsNames = ["ios", "watchos", "ipados", "tvos"];
-  return appleOsNames.includes(platform?.toLowerCase());
+  if (platform && typeof platform === "string") {
+    return appleOsNames.includes(platform?.toLowerCase());
+  }
+  throw new CustomError("Platform is not defined or is not a string", 400);
```
saikumarrs, line 1552: Why throw error? Returning `false` is enough, right?

v0/util/index.js
```diff
@@ -1546,7 +1546,10 @@ const isOAuthSupported = (destination, destHandler) => {
 
 function isAppleFamily(platform) {
   const appleOsNames = ["ios", "watchos", "ipados", "tvos"];
-  return appleOsNames.includes(platform?.toLowerCase());
+  if (platform && typeof platform === "string") {
```
saikumarrs, line 1549:
```suggestion
  if (typeof platform === "string") {
```

v0/destinations/zendesk/transform.js
```diff
@@ -150,7 +150,10 @@ async function getUserId(message, headers, type) {
     type === "group"
       ? get(message, "context.traits")
       : getFieldValueFromMessage(message, "traits");
-  const userEmail = traits.email;
+  const userEmail = traits?.email;
+  if (!userEmail) {
+    throw new CustomError("Email is required for group call", 400);
```
saikumarrs, line 155:
```suggestion
    throw new CustomError("Email ID is required for "group" events", 400);
```

saikumarrs: Also, can you double-check that this flow will always be hit for all the group events?

v0/destinations/ga4/transform.js
```diff
@@ -41,6 +41,11 @@ const responseBuilder = (message, { Config }) => {
   }
 
   // trim and replace spaces with "_"
+  if (typeof event !== "string") {
```
saikumarrs, line 44: Shouldn't this validation be done very early in the execution?

v0/destinations/ga/transform.js
```diff
@@ -934,7 +934,7 @@ function processSingleMessage(message, destination) {
           : ConfigCategory.NON_ECOM;
         category.hitType = "event";
         customParams.ni =
-          message?.properties.nonInteraction !== undefined
+          message?.properties?.nonInteraction !== undefined
             ? message.properties.nonInteraction
             : 1;
```
saikumarrs, line 939:
```suggestion
          isDefinedAndNotNull(message?.properties?.nonInteraction)
            ? message.properties.nonInteraction
            : 1;
```

saikumarrs: Maybe you need to import that utility method.

v0/destinations/facebook_pixel/utils.js
```diff
@@ -214,31 +214,34 @@ const transformedPayloadData = (
...
     return undefined;
   }
-  let parseUrl = new URL(url)
-  let paramsList = new URLSearchParams(parseUrl.search);
-  const fbclid = paramsList.get('fbclid')
+  let parseUrl;
+  try {
+    parseUrl = new URL(url);
+  } catch {
+    throw new CustomError("url is not in correct format", 400);
```
saikumarrs, line 241: Why throw error? From that logic, it doesn't seem like it's required that URL is valid.

v0/destinations/facebook_pixel/transform.js
```diff
@@ -588,6 +588,17 @@ const processEvent = (message, destination) => {
           })
           .build();
       }
+      if (typeof message.event !== "string") {
```
saikumarrs, line 591: We should do this validation very early in the execution.

v0/destinations/ga4/transform.js
```diff
@@ -41,6 +41,11 @@ const responseBuilder = (message, { Config }) => {
...
+    throw new CustomError(
```
saikumarrs, line 45: We should use ErrorBuilder and set appropriate stat tags.

saikumarrs (changes requested):

v0/util/index.js
saikumarrs: This is an anti-pattern. Since this is just validating the input value's presence, type-checking the input can go in here.

utsabc: The function name is `isAppleFamily` which gives an impression that this function returns a boolean value for a supplied value if it meets the condition of mentioned in function name `verb`

If we would also want to check the validity of the supplied value we should change function name to `validateAppleFamily` a verb which share most context.

v0/destinations/ga/transform.js
utsabc: We don't need to check the validity of `message` (i.e root level)

v0/util/index.js
saikumarrs: Yes, any value like "null", "undefined", 2, etc, will fail the apple family check and should return `false`. We shouldn't throw an error. The validation is considered part of the check itself.

v0/destinations/zendesk/transform.js
ItsSudip: We need the email for getting zendeskId which is needed for both the identify and group events.

v0/destinations/ga4/transform.js
ItsSudip: After getting the type of event we are doing these checks first.

v0/destinations/facebook_pixel/transform.js
ItsSudip: It is inside the scope of the switch after finding type of the event.

v0/destinations/ga4/transform.js
ItsSudip: There are lot of customErrors, we can make this change later.

v0/destinations/ga4/transform.js
saikumarrs: I see this is needed for all the event types. So, it makes sense to move to the highest level.

v0/destinations/facebook_pixel/transform.js
saikumarrs: Needs some refactoring but it's fine for now.

v0/destinations/ga4/transform.js
ItsSudip: I think responsebuilder is the best place for this check as for every events we are calling responsebuilder function. And message.event will not available for every events, apart from track for other events we are adding message.event. So it would be better if we do this check in responsebuilder.

saikumarrs (approved):

utsabc (approved):

v0/util/index.js
```diff
@@ -1546,7 +1546,10 @@ const isOAuthSupported = (destination, destHandler) => {
 
 function isAppleFamily(platform) {
   const appleOsNames = ["ios", "watchos", "ipados", "tvos"];
-  return appleOsNames.includes(platform?.toLowerCase());
+  if (typeof platform === "string") {
```
krishna2020, line 1549: Is there a need to support another data type (like array) with valid platform value coming in ?

PR #1539, opened 2022-11-02
feat(Zendesk): update identify call

sauravlal15:
## Description of the change

> In Zendesk for identify call, it will take an extra attribute i.e. primaryEmail from `traits`, and set this primaryEmail as the primary email for users in Zendesk. If that email doesn't exist, then it will add an email and sets it as a primary.
## Type of change
- [ ✅] New feature (non-breaking change that adds functionality)

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


v0/destinations/zendesk/transform.js
```diff
@@ -348,14 +408,27 @@ async function processIdentify(message, destinationConfig, headers) {
...
+  const primaryEmail = get(traits, "primaryEmail");
+  if (userId && primaryEmail) {
+    const userIdentityId = await getUserIdentityId(
+      userId,
+      primaryEmail,
+      headers
+    );
+    if (userIdentityId) {
+      returnList.push(
+        responseBuilderToSetPrimaryAccount(userIdentityId, userId, headers)
```
ItsSudip, line 421: Don't we need to verify the emails?

ItsSudip (commented):
We have to add unit test cases for this change.

v0/destinations/zendesk/transform.js
sauravlal15: No, I've checked that, without verification also it lets me add it as primary.

v0/destinations/zendesk/transform.js
```diff
@@ -347,15 +435,30 @@ async function processIdentify(message, destinationConfig, headers) {
...
+      primaryEmail,
+      headers,
+      destinationConfig
+    );
+    if (userIdentityId) {
+      returnList.push(
+        responseBuilderToSetPrimaryAccount(userIdentityId, userId, headers)
+      );
+    }
+    delete traits.primaryEmail;
```
ItsSudip, line 452: We are already deleting it in line no 419, right?

v0/destinations/zendesk/transform.js
```diff
@@ -45,6 +46,86 @@ function responseBuilder(message, headers, payload, endpoint) {
...
+const getUserIdentityId = async (
+  userId,
+  primaryEmail,
+  headers,
+  destinationConfig
+) => {
+  const url = `${endPoint}users/${userId}/identities`;
+  const config = { headers };
+  try {
+    const resp = await axios.get(url, config);
```
ItsSudip, line 93: Use httpGET function here.

v0/destinations/zendesk/transform.js
```diff
@@ -45,6 +46,86 @@ function responseBuilder(message, headers, payload, endpoint) {
...
+    if (!resp || !resp.data || resp.data.count === 0) {
```
ItsSudip, line 94: If the count=0 we don't need to through error we can create the user.

v0/destinations/zendesk/transform.js
```diff
@@ -45,6 +46,86 @@ function responseBuilder(message, headers, payload, endpoint) {
...
+      }
+    };
+    if (destinationConfig.createUsersAsVerified)
+      payloadBody.identity = { ...payloadBody.identity, verified: true };
+    const response = await axios.post(createIdentityUrl, payloadBody, config);
+    if (!response || !response.data || !response.data.identity) {
+      logger.debug("Failed in creating Identity.");
+      return undefined;
+    }
+    return response.data.identity.id;
```
ItsSudip, line 122:
```suggestion
    return response.data.identity?.id;
```

v0/destinations/zendesk/transform.js
sauravlal15: count === 0 means that there is no identity of the user, but in our flow for setting the email as primary, there must be 1 email of the user which means the count is always greater than zero, otherwise it's unable to fetch the user identities.

v0/destinations/zendesk/transform.js
sauravlal15: Updated

v0/destinations/zendesk/transform.js
sauravlal15: Updated

v0/destinations/zendesk/transform.js
sauravlal15: Done

v0/destinations/zendesk/transform.js
ItsSudip: Yes, we can create the user and add the email to that user.

sauravlal15: Okay, got it, will update.

sauravlal15: Actually, this case will never come, this is just for one more API response validation, As for updating the primary email, we need userId and we are taking userId only from the email, so that means the email should always be there and the count is greater than 0.

ItsSudip: What if the customer provides `primaryEmail` with the first API call?

sauravlal15: updated

v0/destinations/zendesk/transform.js
```diff
@@ -350,15 +453,41 @@ async function processIdentify(message, destinationConfig, headers) {
...
+    // and it will set it as primary while creating the user.
+    const primaryEmailPayload = {
+      type: "email",
+      value: primaryEmail
+    };
+    if (payload.user.identities && Array.isArray(payload.user.identities)) {
+      payload.user.identities.push(primaryEmailPayload);
+    } else {
+      payload.user.identities = [primaryEmailPayload];
+    }
```
ItsSudip, line 481: This scenario will never occur as userId is a required field if you see line number 430.

v0/destinations/zendesk/transform.js
```diff
@@ -45,6 +47,100 @@ function responseBuilder(message, headers, payload, endpoint) {
...
+    if (
+      res.success === false ||
+      !res.response ||
+      !res.response.data ||
+      res.response.data.count === 0
+    ) {
+      logger.debug("Failed in fetching Identity details");
+      return undefined;
+    }
+    const identities = get(res.response.data, "identities");
```
ItsSudip, line 104:
```suggestion
    const { identities } = res.response?.data;
```

sauravlal15: updated

v0/destinations/zendesk/transform.js
```diff
@@ -45,6 +47,100 @@ function responseBuilder(message, headers, payload, endpoint) {
...
 }
 
+/**
+ * Returns the payload for updating primary email of users.
+ * @param {*} userIdentityId -> userIdentity Id
+ * @param {*} userId -> userId of users
+ * @param {*} headers -> Authorizations for API's call
+ * @returns
+ */
+const responseBuilderToSetPrimaryAccount = (
```
ujjwal-ab, line 57: Can't we use the common response builder instead of creating another here?

sauravlal15: for both the builder function the common part is headers, other than that all were different. So, I've separated it for getting its clear use.

ujjwal-ab (approved):

ItsSudip (approved):

PR #1537, opened 2022-11-02
feature: batch user deletion support for mixpanel and clevertap

mihir-4116:
## Description of the change

- as of now for existing destination if we want to delete users then we are doing it one by one. like for every user we are making user deletion request to destination which is impecting the performance of the api
- in this pr for mixpanel and clevertap destination we are making user deletion request in batch since they are supporting batch user deletion
- also, for a perticular user if we don't found required key(ex **userId**) then we will drop that perticular user and proceed for remaining users

## Type of change
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ✅] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally
- [ ] The code changed/added as part of this pull request has been covered with tests
- [ ] All tests related to the changed code pass in development

### Code review 

- [ ]  This pull request has a descriptive title and information useful to a reviewer. There may be a screenshot or screencast attached
- [ ✅] "Ready for review" label attached to the PR and reviewers mentioned in a comment
- [ ] Changes have been reviewed by at least one other engineer
- [ ] Issue from task tracker has a link to this pull request

PR #1551, opened 2022-11-05
feat: move deleteusers tests to a separate file

sanpj2292:
## Description of the change

The intention through this PR is to move the `deleteUsers` tests to a different file rather than have it in `proxy` tests. In the same way, it is seen that we have different responses and `axios` is being used in different ways. To mock all of them in a generic we have implemented a generic axios mock client through which `deleteUsers` test-cases are being executed

**Note**: This generic axios mock client is being currently used by only `GA`. All other destinations(for `deleteUsers`) are using the earlier mock client(`mockedAxiosClient`)

In brief the changes are:
- implement general axios mock client
- test of the general mock axios client
- implement unit tests using the general mock axios client
- new way of defining the http responses

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

- [x] "Ready for review" label attached to the PR and reviewers mentioned in a comment


__tests__/deleteUsers.test.js
```diff
@@ -0,0 +1,57 @@
+const name = "DeleteUsers";
```
krishna2020, line 1: Can you also check on having unit test cases for deleteUsers around batching implementation ?
Ref: https://github.com/rudderlabs/rudder-transformer/pull/1537/files#diff-3107f146043efa1c7ff3cb2ba1ce699080ab88d2e4befc7a370e3bdede0bd1f4
Batching logic (the way data is chunked) should be moved to a separate function & have unit test case around it.

sanpj2292: I see from most of the implementations that we are using lodash's `chunk` method to batch the requests. If batching logic is what we are more concerned with then we'd be testing lodash's `chunk` method actually. Don't think it'd make sense. But if we want to verify if the mappings of userId are being done correctly, we can think on those lines. Not sure which one of these two is being referred to here, can you let me know @krishna2020 ?

krishna2020: I meant the first one only & agree that we shouldn't be concerned as lodash's chunk is getting used & there's no additional custom logic around batching. At least for Mixpanel & Clevertap, no need at the moment.

__tests__/deleteUsers.test.js
```diff
@@ -0,0 +1,57 @@
...
+  describe(`${name} Tests: ${destination}`, () => {
+    beforeAll(() => {
+      let axiosResponses;
+      try {
+        axiosResponses = require(`./data/${destination}_deleteUsers_response.json`);
+      } catch (error) {
+       // Do nothing
+       logger.error(`Error while reading ${destination}_deleteUsers_response.json: ${error}`)
+      }
+      if (Array.isArray(axiosResponses)) {
```
krishna2020, line 34: the major part to test in delete flow is around forming the right req body & if that itself is mocked, what purpose are we trying to achieve for deleteUsers flow ?

krishna2020: I see the need of having unit tests for this module.
We can use ga_deleteUsers_response.json to assert the reqBody generation in userDeletionHandler. And continue using ga_deleteUsers_response.json to mock for complete component test as happening now too.

sanpj2292: That was the case even previously & not in a right way. When no mock response is provided, the `mockedAxiosClient` used to return `{}`, which when processed would return `200` for destinations other than `GA`.

**Purpose**:
1. With this PR, we have provided a mechanism where in we have a way to mention the mock responses separately
2. Also we have implemented a generic axios mock client which can be used for any other requirements as well. Which I feel is a lot more extensible
3. Separating it from `proxy` tests made more sense as we might have other destinations which send a different response to deletion as opposed to event-delivery(as we see in `GA`)

__tests__/genAxios.test.js
```diff
@@ -0,0 +1,125 @@
...
+const mockMethod1 = async () => {
+  // some random code execution before axios.get
+  const resp1 = await axios.get("http:///wwww.example.com/1");
+  // some random code execution before axios
+  const resp2 = await axios({
+    url: "http://www.example.com/2",
+    method: "post"
+  });
+  // some random code executed before axios.post
+  const resp3 = await axios.post("http://www.example.com/3")
```
saikumarrs, line 81: If I were to interchange these calls, will this test case still pass?

__tests__/deleteUsers.test.js
saikumarrs: Moreover, this approach will fail if the requests are made parallelly.

__tests__/genAxios.test.js
sanpj2292: No they would not

__tests__/genAxios.test.js
```diff
@@ -0,0 +1,125 @@
...
+  return [
+    resp1,
+    resp2,
+    resp3
+  ];
+}
+const mockMethod2 = async () => {
```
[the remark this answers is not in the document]
sanpj2292: Are you pointing to the naming or do you want to know why are there two mock methods ?

__tests__/deleteUsers.test.js
sanpj2292: >  assert the reqBody generation in userDeletionHandler

The test-cases for the same has been added. Can you once check and confirm if that suffices ?

__tests__/genAxios.test.js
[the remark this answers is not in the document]
saikumarrs: The naming itself.

__tests__/genAxios.test.js
saikumarrs: That's a very fragile state to be. We should change it probably.
How about serving the response by matching the request configuration (URL, data, params, headers, etc.)?

__tests__/deleteUsers.test.js
saikumarrs: Although the new changes you made succeeded in fulfilling the objective, I'm not too sure that's the right way. 
We cannot ensure all the user deletion handlers do not change or add something to the return value of prepare request function while making the request.

We should intercept the requests to Axios instead, if feasible.

__tests__/genAxios.test.js
sanpj2292: Don't think it is fragile. When the order of requests is changed, that is like a logical change happening & tests should be able to recognise it IMHO.

Serving response by matching the request configuration is one of the hardest mocking techniques. But they can be easily achieved via using libraries like `supertest`, `nock` etc,.
The limitation I see is that, the generic implementation using such libraries would be a little trickier to implement, as each destination's test-case needs to have it's own implementation of http request mocker

Even in `rudder-server` the mock server is not much different. Please let me know if anything I described is not clear

sanpj2292: >How about serving the response by matching the request configuration (URL, data, params, headers, etc.)?

We had implemented a mock axios client with a similar intention but I see limitation wrt generalisation

__tests__/deleteUsers.test.js
sanpj2292: >We should intercept the requests to Axios instead

Can you elaborate ?

__tests__/genAxios.test.js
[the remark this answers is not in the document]
sanpj2292: They are just names intented to mock basically some methods which contain several axios calls.

v0/destinations/ga/deleteUsers.js
```diff
@@ -28,37 +28,64 @@ const userDeletionHandler = async (userAttributes, config, rudderDestInfo) => {
...
+/**
+ * This function will help to delete the users one by one from the userAttributes array.
+ *
+ * @param {*} userAttributes Array of objects with userId, emaail and phone
+ * @param {*} config Destination.Config provided in dashboard
+ * @param {Record<string, any> | undefined} rudderDestInfo contains information about the authorisation details to successfully send deletion request
+ * @returns {Array<{ body: any, headers: any }>}
+ */
+const userDeletionHandler = async (userAttributes, config, rudderDestInfo) => {
+  const userDeleteRequests = prepareDeleteRequest(
```
krishna2020, line 75: can we make prepareDeleteRequest too to parallel processing ?
 await Promise.all(
    userAttributes.map(async user => {const request = await prepareDeleteRequest(user, config, rudderDestInfo); return request;})
    );

sanpj2292: If that is the intended change, what is the intended difference between `userDeletionHandler` and `prepareDeleteRequest` ?

My understanding was that `prepareDeleteRequest` is like the request formation or preparation stage.
The actual requests(using http client like axios) will be made in `userDeletionHandler`
Can you elaborate on this ?

__tests__/genAxios.test.js
saikumarrs: The mocked client needn't be generic. I believe we can feed the expected inputs and outputs.

sanpj2292: No matter what specifics are present in the input, we care more about output that we get. So, I don't think this way we'll have any problem as such.
If this still is not satisfactory, we might need to take the approach taken for mocking axios client in [google_adwords_offline_conversion.test.js](https://github.com/rudderlabs/rudder-transformer/blob/master/__tests__/google_adwords_offline_conversions.test.js#L41-L146)

__tests__/deleteUsers.test.js
sanpj2292: Please look at the code mentioned in the conversation link
https://github.com/rudderlabs/rudder-transformer/pull/1551/files#r1018180720

koladilip (commented):
Partial review only

__tests__/deleteUsers.test.js
```diff
@@ -0,0 +1,113 @@
...
+        destination: currentVal,
+        method: delUsers?.prepareDeleteRequest
+      });
+    } catch (error) {
+      // Do nothing
+    }
+    return acc;
+  }, [])
+  .map(v => {
+    const { inputJson, outputJson, method, destination } = v;
```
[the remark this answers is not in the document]
sanpj2292: Done

v0/destinations/ga/deleteUsers.js
krishna2020: As we don't have batching for UA, other optimization is to do as much parallel processing as possible - so `prepareDeleteRequest` is also one good candidate

sanpj2292: The async version on `prepareDeleteRequest` is performing a little poorly than the current implemented one

__tests__/deleteUsers.test.js
sanpj2292: The fixes has been made in this PR. Can you take a look @saikumarrs ?
https://github.com/rudderlabs/rudder-transformer/pull/1563

saikumarrs: LGTM

saikumarrs (approved):

koladilip (approved):

__tests__/deleteUsers/README.md
```diff
@@ -0,0 +1,52 @@
+# DeleteUsers Tests
+
+All the tests data for deleteUsers are to be present in __tests__/data/deleteUsers/${destination}/
+
+### Files and their significance
+  - __`handler_input.json`__ - Input data for  `handleDeletionOfUsers` function in `versionedRouter.js`(alias for `_deleteUsers_proxy_input.json`)
+  - __`handler_output.json`__ - Output of `handleDeletionOfUsers` function in `versionedRouter.js`(alias for `_deleteUsers_proxy_output.json`)
+  - __`prepare_req_input.json`__ - Input data for `prepareDeleteRequest` function exported from the destination's `deleteUsers.js` (An example can be seen in the case of `ga`)
```
krishna2020, line 8: Delete prepare*.json files & add note on `reqParams` in `nw_client_data.json` getting validated

sanpj2292: I have made changes, can you verify ?

krishna2020 (approved):

PR #1553, opened 2022-11-07
feat(destination): support custom mapping with client_id for ga4 destination

mihir-4116:
## Description of the change
slack thread : https://rudderstack.slack.com/archives/C01E4PLB135/p1666963827304709

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

- [ ✅] "Ready for review" label attached to the PR and reviewers mentioned in a comment


akashrpo:
Generally, we shouldn't change existing test cases. Let's keep a test case where client_id is taken from messageId.

__tests__/data/ga4.json
```diff
@@ -10813,7 +10813,7 @@
         "JSON": {
           "timestamp_micros": 1650950229000000,
           "non_personalized_ads": true,
-          "client_id": "ec5481b6-a926-4d2e-b293-0b3a77c4d3be",
+          "client_id": "3f034872-5e28-45a1-9eda-ce22a3e36d1a",
```
akashrpo, line 10816: Generally, we shouldn't change existing test cases. Let's keep a test case where client_id is taken from messageId.

akashrpo: also, the description should match the intent of the test case.

__tests__/data/ga4.json
```diff
@@ -10723,7 +10720,6 @@
           "device": {
             "adTrackingEnabled": "false",
             "advertisingId": "T0T0T072-5e28-45a1-9eda-ce22a3e36d1a",
-            "id": "3f034872-5e28-45a1-9eda-ce22a3e36d1a",
```
akashrpo, line 10726: Let's not remove existing test cases instead adding new ones for testing specifically for client_id would be recommended.

v0/destinations/ga4/transform.js
```diff
@@ -34,6 +34,33 @@ const {
   GA4_PARAMETERS_EXCLUSION
 } = require("./utils");
 
+/**
+ * Returns GA4 client_id
+ * @param {*} message
+ * @param {*} Config
+ * @returns
+ */
+const getGa4ClientId = (message, Config) => {
```
akashrpo, line 43:
```suggestion
const getGA4ClientId = (message, Config) => {
```

v0/destinations/ga4/transform.js
```diff
@@ -34,6 +34,33 @@ const {
...
+  // first we will search from webapp
+  if (Config.clientIdFieldIdentifier) {
+    const { clientIdFieldIdentifier } = Config;
+    return get(message, clientIdFieldIdentifier);
+  }
+  // if we don't fount it then we will fallback to default search
```
akashrpo, line 49:
```suggestion
  // if we don't find it from the config then we will fall back to the default search
```

v0/destinations/ga4/transform.js
mihir-4116: updated

v0/destinations/ga4/transform.js
mihir-4116: updated

__tests__/data/ga4.json
mihir-4116: @akashrpo this test cases i changed because it's failing because of existing changes, so to fix that i changed it. if we keep it as it is then it will fail

v0/destinations/ga4/transform.js
```diff
@@ -34,6 +34,33 @@ const {
...
+/**
+ * Returns GA4 client_id
+ * @param {*} message
+ * @param {*} Config
+ * @returns
+ */
+const getGA4ClientId = (message, Config) => {
+  // first we will search from webapp
+  if (Config.clientIdFieldIdentifier) {
+    const { clientIdFieldIdentifier } = Config;
```
akashrpo, line 46: we can do object destructuring in the function parameter itself.

mihir-4116: updated

__tests__/data/ga4.json
```diff
@@ -10962,5 +10958,182 @@
         "scope": "exception"
       }
     }
+  },
+  {
+    "description": "Configure client_id (properties.client_id)",
```
akashrpo, line 10963:
```suggestion
    "description": "extract client_id from the path defined in the webapp config i.e (properties.client_id)",
```

v0/destinations/ga4/transform.js
```diff
@@ -34,6 +34,32 @@ const {
...
+/**
+ * Returns GA4 client_id
+ * @param {*} message
+ * @param {*} clientIdFieldIdentifier
+ * @returns
+ */
+const getGA4ClientId = (message, { clientIdFieldIdentifier }) => {
+  // first we will search from webapp
+  if (clientIdFieldIdentifier) {
+    return get(message, clientIdFieldIdentifier);
```
akashrpo, line 46: Do we need to fallback to default values if the value is not found here? If not we need to throw proper error here

mihir-4116: @akashrpo yes, we need to fall back to default search since `clientIdFieldIdentifier` is optional from webapp.

__tests__/data/ga4.json
mihir-4116: updated

v0/destinations/ga4/transform.js
akashrpo: well we have a return statement here in that case it will throw error instead

akashrpo: Let's add a test case for this too

mihir-4116: updated

akashrpo: we are not yet taking the values from the fallback types

__tests__/data/ga4.json
```diff
@@ -10962,5 +10958,307 @@
...
+          "generateLeadValueTrait": "value",
+          "generateLeadCurrencyTrait": "currency",
+          "clientIdFieldIdentifier": "properties.client_id"
+        },
+        "Enabled": true
+      }
+    },
+    "output": {
+      "statusCode": 400,
+      "error": "[Google Analytics 4]: ga4ClientId, anonymousId, context.device.id or messageId must be provided",
```
akashrpo, line 11256: This test case is not correct as we need to take `client_id` from fallback values. Here, we have `"anonymousId": "ea5cfab2-3961-4d8a-8187-3d1858c99090",` defined in the input payload

mihir-4116: updated

v0/destinations/ga4/transform.js
mihir-4116: updated

v0/destinations/ga4/transform.js
```diff
@@ -34,6 +34,34 @@ const {
...
+    clientId = get(message, clientIdFieldIdentifier);
+  }
+  // if we don't find it from the config then we will fall back to the default search
+  if (!clientId) {
+    clientId =
+      getDestinationExternalID(message, "ga4ClientId") ||
+      get(message, "anonymousId") ||
+      get(message, "messageId");
+  }
+  return clientId;
```
akashrpo, line 56:
```suggestion
  return (
    get(message, clientIdFieldIdentifier) ||
    getDestinationExternalID(message, "ga4ClientId") ||
    get(message, "anonymousId") ||
    get(message, "messageId")
  );
```

mihir-4116: @akashrpo this change will fail test cases, and in order it to work i need to pass clientIdIdentifier to every test case

akashrpo (approved):

v0/destinations/ga4/transform.js
```diff
@@ -63,13 +91,10 @@ const responseBuilder = (message, { Config }) => {
...
       // GA4 uses it as an identifier to distinguish site visitors.
-      rawPayload.client_id =
-        getDestinationExternalID(message, "ga4ClientId") ||
-        get(message, "anonymousId") ||
-        get(message, "messageId");
+      rawPayload.client_id = getGA4ClientId(message, Config);
       if (!isDefinedAndNotNull(rawPayload.client_id)) {
         throw new CustomError(
-          "ga4ClientId, anonymousId or messageId must be provided",
+          `[Google Analytics 4]: ${Config.clientIdFieldIdentifier}, ga4ClientId, anonymousId or messageId must be provided`,
```
ItsSudip, line 97: If the custoem is not setting `config.clientIdFieldIdentifier` it will show undefined. Can we use some generic term here?

ItsSudip (changes requested):

v0/destinations/ga4/transform.js
mihir-4116: @ItsSudip i already discussed this with @akashrpo , this error is just for transformer test cases because in real this error will never be thrown because at last we are depending on `messageId` or `anonymousId`. transformer will never throw this error since we wanted to cover all the test cases that's this error is there (also this error is suggested by @akashrpo )

ItsSudip (approved):

PR #1599, opened 2022-11-23
fix(destination): PostHog: set timestamp in right place of the payload

timgl:
## Description of the change

The PostHog transformer doesn't set the timestamp in the right place, causing all events to be set to the current timestamp.

## Type of change

## Related issues

N/A

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/v0/destinations/posthog/transform.js
+++ b/v0/destinations/posthog/transform.js
@@ -83,6 +83,9 @@ const responseBuilderSimple = (message, category, destination) => {
     // fail-safety for developer error
     throw new CustomError(ErrorMessage.FailedToConstructPayload, 400);
   }
+  if(!payload.timestamp && isDefinedAndNotNull(payload.properties['$timestamp'])) {
+    payload.timestamp = payload.properties['$timestamp']
+  }
 
   payload.properties = {
     ...generatePropertyDefination(message),


sauravlal15:
Hi @timgl, I checked this PR, can you give some more data that helps to understand how events were taking the current timestamp? 
Basically, for setting the timestamp with an event, we need to pass it as `timestamp` or `originalTimestamp`. If these fields were missing, then RudderStack sends the current Timestamp with the events.

timgl:
> Hi @timgl, I checked this PR, can you give some more data that helps to understand how events were taking the current timestamp?

They weren't, as they were set within the "properties" field, but they need to be set one level above, so:

```json
{
  "distinct_id": "your distinct id",
  "timestamp": "2022-11-11T12:00:00.0000Z",
  "properties": {"key": "value"}
}
```

sauravlal15:
> > Hi @timgl, I checked this PR, can you give some more data that helps to understand how events were taking the current timestamp?
> 
> They weren't, as they were set within the "properties" field, but they need to be set one level above, so:
> 
> ```json
> {
>   "distinct_id": "your distinct id",
>   "timestamp": "2022-11-11T12:00:00.0000Z",
>   "properties": {"key": "value"}
> }
> ```

Yes, they need to check the event and the timestamp, here is the RudderStack [docs](https://www.rudderstack.com/docs/api/http-api/), as a reference.

timgl:
> Yes, they need to check the event and the timestamp, here is the RudderStack [docs](https://www.rudderstack.com/docs/api/http-api/), as a reference.

I was using BigQuery ETL to send events and they didn't have the timestamp set correctly. Are you saying I didn't configure that BigQuery destination correctly?

sauravlal15:
> I was using BigQuery ETL to send events and they didn't have the timestamp set correctly. Are you saying I didn't configure that BigQuery destination correctly?
Hey, can you send the timestamp at the root level instead of sending in the properties, we are mapping everything from properties to posthog properties so the timestamp is going inside properties. So can you send the timestamp at the root level of the payload?

timgl:
> Hey, can you send the timestamp at the root level instead of sending in the properties, we are mapping everything from properties to posthog properties so the timestamp is going inside properties. So can you send the timestamp at the root level of the payload?

That is what this PR is trying to fix, as I don't think this is possible when setting up the destination in Rudderstack?
image

sauravlal15:
> > Hey, can you send the timestamp at the root level instead of sending in the properties, we are mapping everything from properties to posthog properties so the timestamp is going inside properties. So can you send the timestamp at the root level of the payload?
> 
> That is what this PR is trying to fix, as I don't think this is possible when setting up the destination in Rudderstack? 

Hey, as per our payload structure we can take timestamps from the root level of the message. And if you want to use it from properties then we have to add in a mapping file not from here. 
CC: @ItsSudip

timgl:
> Hey, as per our payload structure we can take timestamps from the root level of the message. And if you want to use it from properties then we have to add in a mapping file not from here.

Would you like me to do this or can you handle this?

sauravlal15:
Hey @timgl, what's the update on this?

timgl:
I implemented it in transformations for now but it'd still be good to fix this for everyone else I think?

Also, I'm still having issues with the [reverse ETL](https://rudderstack.slack.com/archives/C01E4PLB135/p1670271500990999) but struggling to figure out the solution, so would be awesome to get a hand there.

timgl:
@ItsSudip Could you bump a fix for this of some kind? I'm still using the hacky transformation to work around this.

ItsSudip:
Hey, @timgl we can absorb this change. Only one suggestion in spite of `$timestamp` can we expect `timestamp` in properties?

timgl:
The timestamp is coming from the rudderstack side so I'm not entirely sure! I don't think we ever use `$timestamp`

mihir-4116:
@timgl created ew pr against develop as per our release rules. this change will get released with next transformer release
pr : https://github.com/rudderlabs/rudder-transformer/pull/2507

PR #1658, opened 2022-12-13
refactor: use new error types

saikumarrs:
## Description of the change

Replaced old error types with new in the regulation API utils.

## Type of change

## Related issues

N/A

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/src/v0/util/regulation-api.js
+++ b/src/v0/util/regulation-api.js
@@ -1,4 +1,4 @@
-const ErrorBuilder = require("./error");
+const { PlatformError } = require("./errorTypes");
 
 class RegulationApiUtils {
   /**
@@ -8,10 +8,7 @@ class RegulationApiUtils {
    */
   static executeCommonValidations(userAttributes) {
     if (!Array.isArray(userAttributes)) {
-      throw new ErrorBuilder()
-        .setMessage("userAttributes is not an array")
-        .setStatus(400)
-        .build();
+      throw new PlatformError("userAttributes is not an array");
     }
   }
 }
--- a/src/v0/util/regulation-api.js
+++ b/src/v0/util/regulation-api.js
@@ -1,4 +1,4 @@
-const { InstrumentationError } = require("./errorTypes");
+const { PlatformError } = require("./errorTypes");
 
 class RegulationApiUtils {
   /**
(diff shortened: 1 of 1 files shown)


src/v0/util/regulation-api.js
```diff
@@ -8,10 +8,7 @@ class RegulationApiUtils {
    */
   static executeCommonValidations(userAttributes) {
     if (!Array.isArray(userAttributes)) {
-      throw new ErrorBuilder()
-        .setMessage("userAttributes is not an array")
-        .setStatus(400)
-        .build();
+      throw new PlatformError("userAttributes is not an array");
```
utsabc, line 11:
```suggestion
      throw new PlatformError("[Regulation API] userAttributes is not an array");
```

koladilip: @utsabc we are not adding this to message as context will be added by server but since is this not a destination, what metadata does server add for this one?

koladilip (approved):

src/v0/util/regulation-api.js
saikumarrs: User deletion is for destinations. Hence, the destination type stat tag is added at the versioned router level in the transformer.

saikumarrs: Also, I just realized that it's actually an `InstrumentationError` rather than `PlatformError`. I'll fix this.

utsabc: Why `InstrumentationError` this request is coming from server

utsabc: This is platform only

saikumarrs: @utsabc, the context of user deletion is already captured at the stat tags level, `feature: "userDeletion"`. So, this won't be required.

saikumarrs: However, sending stat tags isn't part of the data contract for the regulation API. So, once we add it, this will be addressed.

saikumarrs: > Why `InstrumentationError` this request is coming from server

@utsabc, users would send this data, right?

utsabc: No server is forming this request: https://github.com/rudderlabs/rudder-server/blob/37f9bb37a943a37d4d7fee6e69fbcf164ba5c181/regulation-worker/internal/delete/api/api.go#L170

saikumarrs: Addressed.

utsabc (approved):

PR #1570, opened 2022-11-14
feat: transformer refactor

utsabc:
## Description of the change

This PR works towards refactoring transformer and into specific modules and bring MVC architecture for better organization of codebase

- Removed Unnecessary Data files which are not used today

## Type of change
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally
- [ ] The code changed/added as part of this pull request has been covered with tests
- [ ] All tests related to the changed code pass in development

### Code review 

- [ ]  This pull request has a descriptive title and information useful to a reviewer. There may be a screenshot or screencast attached
- [ ] "Ready for review" label attached to the PR and reviewers mentioned in a comment
- [ ] Changes have been reviewed by at least one other engineer
- [ ] Issue from task tracker has a link to this pull request

PR #1710, opened 2022-12-18
chore(refactor): refactor work for transformer

utsabc:
## Description of the change

breaking down versionedRouterjs to following components
- services
- controller
- routes
 

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


utsabc:
Continued work from the following [PR](https://github.com/rudderlabs/rudder-transformer/pull/1570)

tsconfig.json
```diff
@@ -0,0 +1,105 @@
...
+    // "noPropertyAccessFromIndexSignature": true,       /* Enforces using indexed accessors for keys declared using an indexed type. */
+    // "allowUnusedLabels": true,                        /* Disable error reporting for unused labels. */
+    // "allowUnreachableCode": true,                     /* Disable error reporting for unreachable code. */
+
+    /* Completeness */
+    // "skipDefaultLibCheck": true,                      /* Skip type checking .d.ts files that are included with TypeScript. */
+    "skipLibCheck": true                                 /* Skip type checking all .d.ts files. */
+  },
+  "exclude": ["dist", "node_modules", "benchmark", "__mocks__", "__tests__", "tsconfig.json"],
+  "include": ["./**/*.json", "./**/*.ts", "./**/*.js", "services/serviceSelectorts"]
```
saikumarrs, line 104: Since we are only bothered with files in `src`,
```suggestion
  // "exclude": [],
  "include": ["./src/**/*.json", "./src/**/*.ts", "./src/**/*.js", "services/serviceSelectorts"]
```

saikumarrs, line 104: What is "services/serviceSelectorts"?

package.json
```diff
@@ -37,6 +37,9 @@
     "@aws-sdk/lib-storage": "^3.142.0",
     "@bugsnag/js": "^7.18.0",
     "@ndhoule/extend": "^2.0.0",
+    "@types/koa": "^2.13.5",
+    "@types/koa-bodyparser": "^4.3.10",
+    "@types/koa-router": "^7.4.4",
```
saikumarrs, line 42: Shouldn't they go into `devDependencies`?

src/util/dynamicConfigInjector.ts
```diff
@@ -0,0 +1,77 @@
+import cloneDeep from "lodash/cloneDeep";
+import { ProcessorRequest, RouterRequestData } from "../types/index";
+
+/* eslint-disable no-param-reassign */
+const get = require("get-value");
+const unset = require("unset-value");
+
+export class DynamicConfigInjectionService {
```
saikumarrs, line 8: A service under `util`?

src/routes/utility.router.ts
```diff
@@ -0,0 +1,15 @@
...
+import MetricsController from "../controllers/metrics.controller";
+import ProfileController from "../controllers/profile.controller";
+import UtilityController from "../controllers/utility.controller";
+
+const router = new Router();
+
+router.post("/heapdump", ProfileController.profile);
+router.get("/metrics", MetricsController.exportMetric);
+router.get("/health", UtilityController.healthStats);
+router.get("/transformerBuildVersion", UtilityController.buildVersion);
```
saikumarrs, line 11: Let's change it to just `buildVersion`. If we have to be backward compatible, keep both but mark the current one as deprecated.

src/routes/trackingPlan.router.ts
```diff
@@ -0,0 +1,8 @@
+import Router from "koa-router";
+import TrackingPlanController from "../controllers/trackingPlan.controller";
+
+const router = new Router();
+
+router.post("/v0/validate", TrackingPlanController.validateTrackingPlan);
```
saikumarrs, line 6:
```suggestion
router.post("/:version/validate", TrackingPlanController.validateTrackingPlan);
```

src/routes/bulkUpload.router.ts
```diff
@@ -0,0 +1,15 @@
...
+  getWarnJobStatus
+} from "../controllers/bulkUpload.controller";
+
+const router = new Router();
+
+router.post("/fileUpload", fileUpload);
+router.post("/pollStatus", pollStatus);
+router.post("/getFailedJobs", getFailedJobStatus);
+router.post("/getWarningJobs", getWarnJobStatus);
+export const sourceRoutes = router.routes();
```
saikumarrs, line 15: Probably a typo to name it `sourceRoutes`.

src/routes/utility.router.ts
```diff
@@ -0,0 +1,15 @@
+import Router from "koa-router";
```
saikumarrs, line 1: The term "utility" doesn't sound right. I'd instead suggest "miscellaneous" until we find the right buckets for each of these routes.

src/controllers/utility.controller.ts
```diff
@@ -0,0 +1,36 @@
+import { Context } from "koa";
+import fs from "fs";
+
+export default class UtilityController {
```
saikumarrs, line 4: Let's write a parallel service module to move the response logic there.

src/controllers/destination.controller.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  public static batchProcess(ctx: Context) {
+    let requestMetadata = MiscService.getRequestMetadata(ctx);
+    const routerRequest = ctx.request.body as RouterRequest;
+    const destination = routerRequest.destType;
+    let events = routerRequest.input;
+    const integrationService = ServiceSelector.getDestinationService(events);
+    const destinationHandler = ServiceSelector.getDestHandler(
+      destination,
+      "V0"
```
saikumarrs, line 68: Let's not hardcode this way.

src/controllers/destination.controller.ts
```diff
@@ -0,0 +1,84 @@
...
+      ctx
+    ) as ProcessorRequest[];
+    const resplist = await integrationService.processorRoutine(
+      events,
+      destination,
+      destinationHandler,
+      requestMetadata
+    );
+    ctx.body = resplist;
+    MiscService.transformerPostProcessor(ctx);
```
saikumarrs, line 31: Theoretically, it's not a service's responsibility to update the `ctx` object.

src/util/serviceSelector.ts
```diff
@@ -0,0 +1,106 @@
+import {  ProcessorRequest, RouterRequestData } from "../types/index";
+import { MiscService } from "../services/misc.service";
+import { INTEGRATION_SERVICE } from "../routes/utils/constants";
+import CDKV1ServiceDestination from "../services/destination/cdkV1Integration.destination.service";
+import CDKV2ServiceDestination from "../services/destination/cdkV2Integration.destination.service";
+import IntegrationServiceDestination from "../interfaces/IntegrationDestinationService";
+import NativeIntegrationServiceDestination from "../services/destination/nativentegration.destination.service";
+import IntegrationServiceSource from "../interfaces/IntegrationSourceService";
+import NativeIntegrationServiceSource from "../services/source/nativeIntegration.source.service";
```
saikumarrs, line 9: The naming is a bit odd. It should rather be `<implementation><module>Service` (ex: `CDKV1DestinationService`, `NativeSourceService` etc.).

src/util/serviceSelector.ts
```diff
@@ -0,0 +1,106 @@
...
+
+export class ServiceSelector {
+  private static sourceHandlerMap: Map<string, any> = new Map();
+  private static destHandlerMap: Map<string, any> = new Map();
+  private static serviceMap: Map<string, any> = new Map();
```
saikumarrs, line 14: How about a multi-level map to hold all these together?

saikumarrs (commented):
Partial review 1

A couple of common suggestions:
- error handling should be implemented for all the public controller functions
- Bugsnag notification should be integrated for all the routes (under our purview).
- A lot of the logic from controllers should be moved to their corresponding service modules.

src/controllers/destination.controller.ts
```diff
@@ -0,0 +1,84 @@
...
+    let requestMetadata = MiscService.getRequestMetadata(ctx);
+    const routerRequest = ctx.request.body as RouterRequest;
+    const destination = routerRequest.destType;
+    let events = routerRequest.input;
+    const integrationService = ServiceSelector.getDestinationService(events);
+    const destinationHandler = ServiceSelector.getDestHandler(
+      destination,
+      "v0"
+    );
+    events = PreTransformationServiceDestination.preProcess(
```
koladilip, line 45: Can we make this generic method like PreTransformationServiceDestination.preProcess<RouterRequestData>

src/controllers/destination.controller.ts
```diff
@@ -0,0 +1,84 @@
...
+    return ctx;
+  }
+
+  public static batchProcess(ctx: Context) {
+    let requestMetadata = MiscService.getRequestMetadata(ctx);
+    const routerRequest = ctx.request.body as RouterRequest;
+    const destination = routerRequest.destType;
+    let events = routerRequest.input;
+    const integrationService = ServiceSelector.getDestinationService(events);
+    const destinationHandler = ServiceSelector.getDestHandler(
```
koladilip, line 66: ServiceSelector.getDestHandler is generic or only for native (v0)?

src/controllers/profile.controller.ts
```diff
@@ -0,0 +1,15 @@
...
+import { UserTransfromServiceResponse } from "../types/index";
+import profile from "../services/profile.service";
+export default class ProfileController {
+  public static async profile(ctx: Context) {
+    let requestMetadata = MiscService.getRequestMetadata(ctx);
+    const format = ctx.request.query.format as string;
+    const { credBucketDetails } = ctx.request.body as any;
+    const response = await profile(credBucketDetails, format);
+    ctx.body = response.body;
+    MiscService.transformerPostProcessor(ctx, response.status);
```
koladilip, line 12: transformerPostProcessor-> postProcess or something in that direction function names should action/verbs

src/controllers/regulation.controller.ts
```diff
@@ -0,0 +1,105 @@
...
+        );
+      }
+      return {};
+    };
+
+    const { body } = ctx.request;
+    const respList = [];
+    const rudderDestInfo = getRudderDestInfo();
+    let response;
+    await Promise.all(
```
koladilip, line 54: Avoid such logic as it is hard to guarantee order

src/controllers/userTransform.controller.ts
```diff
@@ -0,0 +1,78 @@
...
+import { compileUserLibrary } from "../util/ivmFactory";
+import UserTransformService from "../services/userTransform/userTransform.service";
+import logger from "../logger";
+import { setupUserTransformHandler } from "../util/customTransformer";
+
+export default class UserTransformController {
+  public static async userTransform(ctx: Context) {
+    let requestMetadata = MiscService.getRequestMetadata(ctx);
+    const events = ctx.request.body as ProcessorRequest[];
+    const processedRespone: UserTransfromServiceResponse = await UserTransformService.userTransformRoutine(
```
koladilip, line 13: UserTransformService.userTransformRoutine -> UserTransformService.transform (function should be verb/action)

src/interfaces/IntegrationDestinationService.ts
```diff
@@ -0,0 +1,41 @@
...
+    Metadata,
+    ProcessorRequest,
+    ProcessorResponse,
+    RouterRequestData,
+    RouterResponse,
+    TransformedEvent
+  } from "../types/index";
+  
+  export default interface IntegrationDestinationService {
+    processorRoutine(
```
koladilip, line 12: make sure names follow function naming conventions

src/routes/utility.router.ts
koladilip: Internally service only right, we can do it for now, why maintain both

src/services/profile.service.ts
```diff
@@ -100,17 +100,17 @@ function promisifiedRead(readable) {
...
   try {
     const supportedCloudProvidersForDumpStorage = ["aws"];
-    const credBucketDetails = ctx.request.body;
     const shouldGenerateLocally = !credBucketDetails.sendTo;
     console.log("Before Heapsnapshot converted into a readable stream");
     let fileName = "";
     let format = "pb.gz";
     let profile;
-    if (ctx.request.query.format && ctx.request.query.format === "v8") {
+    if (format === "v8") {
```
koladilip, line 113: format defined here then why need if condition?

src/services/userTransform/userTransform.service.ts
```diff
@@ -0,0 +1,172 @@
...
+          destinationType: eventsToProcess[0]?.metadata.destinationType,
+          messageIds
+        };
+
+        const metaTags =
+          eventsToProcess.length && eventsToProcess[0].metadata
+            ? getMetadata(eventsToProcess[0].metadata)
+            : {};
+
+        if (transformationVersionId) {
```
koladilip, line 54: handle not first using `if (!transformationVersionId) {` then you can avoid nesting for other case, more readable.

src/util/dynamicConfigInjector.ts
```diff
@@ -0,0 +1,77 @@
...
+          return value
+        } */
+    return value;
+  }
+
+  private static configureVal(
+    value: any,
+    event: ProcessorRequest | RouterRequestData
+  ) {
+    if (value) {
```
koladilip, line 46: handle `if(!value)` first to avoid nesting for other case

src/util/serviceSelector.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  private static isCdkV2Destination(destinationDefinitionConfig: Object) {
+    return !!destinationDefinitionConfig?.["cdkV2Enabled"];
+  }
+
+  private static fetchCachedService(serviceType: string) {
+    let service: any;
+    if (this.serviceMap.has(serviceType)) {
+      service = this.serviceMap.get(serviceType);
```
koladilip, line 27: `return  this.serviceMap.get(serviceType);` to avoid else case and nesting for other case

src/routes/utility.router.ts
saikumarrs: I'm not sure which all places this is used. Hence, was suggesting to support both for backward compatibility.

src/util/dynamicConfigInjector.ts
utsabc: This is basically the service selector earlier we discussed this could be an utility hence placed it here @koladilip @krishna2020 should this be mode under service dir?

src/controllers/utility.controller.ts
utsabc: There isn't much logic to move to a service you have anything specific in mind

saikumarrs: Even if the logic is simpler, it's better to follow the convention and move it to a service. Future developers can make mistakes and extend functionality here itself.

src/controllers/destination.controller.ts
utsabc: Actually this is a post processing layer not sure what to call this

saikumarrs: This can be a utility method within the controllers.

src/util/serviceSelector.ts
saikumarrs: It improves extensibility. Also, aligns with the tags structure we are now coming up with.

src/controllers/bulkUpload.controller.ts
```diff
@@ -0,0 +1,188 @@
...
+  // cluster, namespace, etc information
+  // from the request
+  return {
+    namespace: "Unknown",
+    cluster: "Unknown"
+  };
+};
+
+export const fileUpload = async ctx => {
+  const getReqMetadata = () => {
```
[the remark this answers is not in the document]
utsabc: I haven't refactored this piece of code, I have one doubt regard the functions inside another functions (during run time) its interpreted only during the execution of the parent function to save on the performance. But if we declare the common function do we loose that benefit?

tsconfig.json
utsabc: Addressed

utsabc: This is removed

package.json
utsabc: Refactoed

src/routes/utility.router.ts
utsabc: Refactored

src/routes/bulkUpload.router.ts
utsabc: Addressed

src/routes/utility.router.ts
utsabc: Addressed

src/controllers/utility.controller.ts
utsabc: Done

src/controllers/destination.controller.ts
utsabc: Fixed

src/controllers/destination.controller.ts
utsabc: Done

src/controllers/bulkUpload.controller.ts
```diff
@@ -0,0 +1,188 @@
...
+  };
+
+  const { destType } = ctx.request.body;
+  const destFileUploadHandler = getPollStatusHandler(
+    "v0",
+    destType.toLowerCase()
+  );
+  let response;
+  if (!destFileUploadHandler || !destFileUploadHandler.processPolling) {
+    ctx.status = 404;
```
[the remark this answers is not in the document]
utsabc: On second thought reverted  for now to verify server compatibility

src/controllers/bulkUpload.controller.ts
```diff
@@ -0,0 +1,188 @@
...
+    ctx.body = `${destType} doesn't support bulk upload`;
+    return null;
+  }
+  try {
+    response = await destFileUploadHandler.processPolling(ctx.request.body);
+  } catch (error) {
+    response = {
+      statusCode: error.response ? error.response.status : 400,
```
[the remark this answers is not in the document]
utsabc: Fixed

src/controllers/bulkUpload.controller.ts
```diff
@@ -0,0 +1,188 @@
...
+      ...response,
+      ...getCommonMetadata(ctx),
+      ...getReqMetadata()
+    });
+  }
+  ctx.body = response;
+  return ctx.body;
+};
+
+export const getFailedJobStatus = async ctx => {
```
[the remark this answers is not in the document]
utsabc: Yes we can refactor this I haven't refactored much on these files as we discussed this whole flow needs some design discussion

src/controllers/delivery.controller.ts
```diff
@@ -0,0 +1,53 @@
+import { Context } from "koa";
+import { MiscService } from "../services/misc.service";
+import { TransformedEvent } from "../types/index";
+import { ServiceSelector } from "../util/serviceSelector";
+import networkHandlerFactory from "../adapters/networkHandlerFactory";
```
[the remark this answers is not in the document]
utsabc: these are stome static utils

src/controllers/destination.controller.ts
utsabc: Can you elaborate a bit more?

src/controllers/destination.controller.ts
utsabc: Its native flow only for transformation. As of now its need in other flow for metadata processing after transformation

src/controllers/regulation.controller.ts
utsabc: This is still to be refactored didn't touch much of regulation flow will take it up

src/services/profile.service.ts
utsabc: Fixed minor bug

src/util/serviceSelector.ts
utsabc: We would want to store the services if not present right?

src/util/dynamicConfigInjector.ts
utsabc: fixed

src/services/userTransform/userTransform.service.ts
utsabc: fixed

src/controllers/destination.controller.ts
```diff
@@ -1,10 +1,17 @@
...
+import {
+  ProcessorRequest,
+  RouterRequestData,
+  RouterRequest
+} from "../types/index";
 import { ServiceSelector } from "../util/serviceSelector";
+import ControllerUtility from "./util";
 
 export default class DestinationController {
+  private static DEFAULT_VERSION = "v0";
```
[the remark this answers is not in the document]
utsabc: For router transform we are not sending it

src/util/dynamicConfigInjector.ts
utsabc: Oh yeah for this one yes I confused it with serviceSerlector

src/controllers/destination.controller.ts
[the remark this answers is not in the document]
saikumarrs: We can use the default value only for that.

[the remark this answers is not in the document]
utsabc: Ok you mean we declare this within the function?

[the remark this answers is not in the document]
saikumarrs: No, declare outside, but use the default only for that.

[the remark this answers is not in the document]
utsabc: Yeah I am using only at `destinationTransformAtRouter`

sanpj2292:
>this method's implementation is request/context specific. Hence, it cannot be extracted as a utility. The only thing common across all these functions is the method name itself.
However, after we establish consistency in the request structure across endpoints, it makes sense to.

What about having this as a common middleware and binding it to the request ?

```javascript

koaRouter.post("/v0/my/endpoint", (ctx, next) => { 
 ctx.req.getReqMetadata = () => { ... }
 next();
}, MyEndPointHandler);

// MyEndpointHandler
// inside the specific method
// call using ctx.req.getReqMetadata() & make use of it wherever required

```
If there are some common things, we can atleast use the same middleware for a couple of requests

src/interfaces/IntegrationDestinationService.ts
```diff
@@ -0,0 +1,41 @@
...
+      events: ProcessorRequest[],
+      destinationType: string,
+      destHandler: any,
+      requestMetadata: Object
+    ): Promise<ProcessorResponse[]>;
+  
+    routerRoutine(
```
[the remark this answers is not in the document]
utsabc: Yes we can @koladilip wdyt?

src/types/index.ts
```diff
@@ -0,0 +1,191 @@
...
+  status: number;
+  message: string;
+  destinationResponse: Object;
+  statTags: Object;
+  authErrorCategory?: string;
+};
+
+type RudderMessage = {
+  userId?: string;
+  anonymousId: string;
```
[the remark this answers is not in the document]
utsabc: Yes but how do we define that either one of them is mandatory?

src/types/index.ts
```diff
@@ -0,0 +1,191 @@
...
+  params?: Record<string, unknown>;
+  body?: {
+    JSON?: Record<string, unknown>;
+    JSON_ARRAY?: Record<string, unknown>;
+    XML?: Record<string, unknown>;
+    FORM?: Record<string, unknown>;
+  };
+  files?: Record<string, unknown>;
+  metadata?: Metadata;
+  [key: string]: any;
```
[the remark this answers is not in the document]
utsabc: This is kept to allow anything extra which we are not considering now. We can limit this by making type more strict

src/types/index.ts
```diff
@@ -0,0 +1,191 @@
+type TransformedEvent = {
+  version: string;
+  type: string;
+  method: string;
```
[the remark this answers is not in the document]
utsabc: Make sense

src/types/index.ts
```diff
@@ -0,0 +1,191 @@
+type TransformedEvent = {
```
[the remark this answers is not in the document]
utsabc: Ok

src/types/index.ts
```diff
@@ -0,0 +1,191 @@
...
+    JSON_ARRAY?: Record<string, unknown>;
+    XML?: Record<string, unknown>;
+    FORM?: Record<string, unknown>;
+  };
+  files?: Record<string, unknown>;
+  metadata?: Metadata;
+  [key: string]: any;
+};
+
+type Metadata = {
```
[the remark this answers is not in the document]
utsabc: This is server metadata for a job

src/types/index.ts
```diff
@@ -0,0 +1,191 @@
...
+  ID: string;
+  Config: Object;
+};
+
+type DestinationDefinition = {
+  ID: string;
+  Name: string;
+  DisplayName: string;
+  Config: Object;
+  ResponseRules: Object;
```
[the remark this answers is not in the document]
utsabc: Yes

src/types/index.ts
```diff
@@ -0,0 +1,191 @@
...
+  messageIds: string[];
+  rudderId: string;
+  receivedAt: string;
+  eventName: string;
+  eventType: string;
+  sourceDefinitionId: string;
+  destinationDefinitionId: string;
+};
+
+type UserTransformation = {
```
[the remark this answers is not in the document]
utsabc: Input

src/routes/misc.router.ts
```diff
@@ -13,4 +13,4 @@ router.get("/buildVersion", MiscController.buildVersion);
 router.get("/version", MiscController.version);
 router.get("/features", MiscController.features);
 
-export const utilityRoutes = router.routes();
+export const miscRoutes = router.routes();
```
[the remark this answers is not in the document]
utsabc: ok

src/controllers/userTransform.controller.ts
```diff
@@ -0,0 +1,80 @@
...
+import MiscService from "../services/misc.service";
+import { ProcessorRequest, UserTransfromServiceResponse } from "../types/index";
+import { compileUserLibrary } from "../util/ivmFactory";
+import UserTransformService from "../services/userTransform/userTransform.service";
+import logger from "../logger";
+import { setupUserTransformHandler } from "../util/customTransformer";
+import ControllerUtility from "./util";
+
+export default class UserTransformController {
+  public static async userTransform(ctx: Context) {
```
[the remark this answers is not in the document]
utsabc: But without dir segregation during import it would be confusing to import and find

src/controllers/userTransform.controller.ts
```diff
@@ -0,0 +1,80 @@
...
+      const res = await compileUserLibrary(code);
+      ctx.body = res;
+    } catch (error) {
+      ctx.body = { error: error.message };
+      ctx.status = 400;
+    }
+    return ctx;
+  }
+
+  public static async testUsertransformSethandle(ctx: Context) {
```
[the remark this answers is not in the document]
utsabc: This is used before saving a user transformation to `test` validity for it

src/types/index.ts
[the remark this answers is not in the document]
saikumarrs: Let's remove this. Any such extra considerations would be evident that way.

src/types/index.ts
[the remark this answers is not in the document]
saikumarrs: Let's add that info to the type. It sounds like the user transformation function itself.

src/interfaces/IntegrationDestinationService.ts
```diff
@@ -0,0 +1,41 @@
...
+    DeliveryResponse,
+    Metadata,
+    ProcessorRequest,
+    ProcessorResponse,
+    RouterRequestData,
+    RouterResponse,
+    TransformedEvent
+  } from "../types/index";
+  
+  export default interface IntegrationDestinationService {
```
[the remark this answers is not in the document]
utsabc: Addressed

src/util/serviceSelector.ts
```diff
@@ -0,0 +1,106 @@
...
+    if (this.sourceHandlerMap.get(source)) {
+      sourceHandler = this.destHandlerMap.get(source);
+    } else {
+      sourceHandler = MiscService.getSourceHandler(source, version);
+      this.sourceHandlerMap.set(source, sourceHandler);
+    }
+    return sourceHandler;
+  }
+
+  public static getNativeIntegrationServiceDest(): IntegrationDestinationService {
```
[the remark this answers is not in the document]
utsabc: Addressed

src/util/serviceSelector.ts
```diff
@@ -0,0 +1,106 @@
...
+    return this.fetchCachedService(INTEGRATION_SERVICE.NATIVE_DEST);
+  }
+
+  public static getNativeIntegrationServiceSource(): IntegrationSourceService {
```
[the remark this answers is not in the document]
utsabc: Addressed

src/types/index.ts
```diff
@@ -0,0 +1,191 @@
...
+  Name: string;
+  DestinationDefinition: DestinationDefinition;
+  Config: Object;
+  Enabled: boolean;
+  WorkspaceID: string;
+  Transformations: UserTransformation[];
+  RevisionID: string;
+};
+
+type Library = {
```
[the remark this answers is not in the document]
utsabc: Addressed

src/types/index.ts
```diff
@@ -0,0 +1,191 @@
...
+  VersionID: string;
+};
+
+type ProcessorRequest = {
```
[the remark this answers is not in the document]
utsabc: Addressed

src/types/index.ts
```diff
@@ -0,0 +1,191 @@
...
+  message: string;
+  destinationResponse: Object;
+  statTags: Object;
+  authErrorCategory?: string;
+};
+
+type RudderMessage = {
+  userId?: string;
+  anonymousId: string;
+  type: string;
```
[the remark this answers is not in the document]
utsabc: Addressed

src/types/index.ts
```diff
@@ -0,0 +1,191 @@
...
+  channel: string;
```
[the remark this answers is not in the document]
utsabc: Sometimes this is destinations specific hence keeping this open ended (need to be discussed with new intg team)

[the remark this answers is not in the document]
saikumarrs: The possible values are `web`, `mobile`, and `server`.

src/controllers/userTransform.controller.ts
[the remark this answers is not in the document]
saikumarrs: I meant to fix the method name `userTransform` as the class name already establishes the context.

src/util/utils.js
```diff
@@ -23,8 +23,13 @@ const responseStatusHandler = (status, entity, id, url) => {
   }
 };
 
+const getIntegrationVersion = () => {
```
saikumarrs, line 26: It doesn't have to be a function, just a constant.

src/types/index.ts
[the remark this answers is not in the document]
utsabc: For `pinterest` we are taking some other values as well from `channel` though channel parameter is an optional source from which its taken 
Screenshot 2023-01-02 at 11 10 42 AM

src/types/index.ts
```diff
@@ -14,7 +14,6 @@ type TransformedEvent = {
   };
   files?: Record<string, unknown>;
   metadata?: Metadata;
-  [key: string]: any;
 };
 
 type Metadata = {
```
saikumarrs, line 19: From all the type and field names, replace `transform` with `transformation`, as the former is a verb and the latter is a noun.

src/controllers/userTransform.controller.ts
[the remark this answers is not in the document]
utsabc: Screenshot 2023-01-02 at 11 13 20 AM
Currently this is the structure I kept if I remove `user` it might become confusing while user is looking at just file name

saikumarrs (dismissed):

src/types/index.ts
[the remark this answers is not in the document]
saikumarrs: That seems like a bad implementation. The `channel` field is not supposed to take other values as it is part of the HTTP event spec.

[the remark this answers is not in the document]
utsabc: Agreed, just wanted to confirm with new-intg team before making breaking changes

src/controllers/userTransform.controller.ts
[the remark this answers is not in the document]
saikumarrs: I'm not asking to remove "user" from the file name but from the method name inside it, `userTransform` to `transform`. However, looks like you've already addressed that.

saikumarrs:
> > this method's implementation is request/context specific. Hence, it cannot be extracted as a utility. The only thing common across all these functions is the method name itself.
> > However, after we establish consistency in the request structure across endpoints, it makes sense to.
> 
> What about having this as a common middleware and binding it to the request ?
> 
> ```js
> koaRouter.post("/v0/my/endpoint", (ctx, next) => { 
>  ctx.req.getReqMetadata = () => { ... }
>  next();
> }, MyEndPointHandler);
> 
> // MyEndpointHandler
> // inside the specific method
> // call using ctx.req.getReqMetadata() & make use of it wherever required
> ```
> 
> If there are some common things, we can atleast use the same middleware for a couple of requests

How is this a common middleware? We'd have to define `getReqMetadata` for each endpoints, right? Moreover, I'd avoid tampering with the `ctx.req` object.

src/types/index.ts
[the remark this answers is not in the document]
saikumarrs: We can try this,
https://stackoverflow.com/a/57528831/17305433

src/types/index.ts
```diff
@@ -0,0 +1,213 @@
...
+  userId?: string;
+  anonymousId: string;
+  type: MessageType;
+  channel: string;
+  context: Object;
+  originalTimestamp: Date;
+  sentAt: Date;
+  event?: string;
+  integrations?: Object;
+  messageId?: string;
```
saikumarrs, line 158: I don't think this is optional.

src/types/index.ts
```diff
@@ -0,0 +1,213 @@
+type TransformedEvent = {
```
saikumarrs, line 1: Replace all the references of `Object` in this file with the `Record<string, any>` type wherever the requirement is strictly an object.

src/types/index.ts
```diff
@@ -0,0 +1,213 @@
...
+  integrations?: Object;
+  messageId?: string;
+  properties?: Object;
+  traits?: Object;
+};
+
+type ErrorDetailer = {
+  module?: string;
+  implementation?: string;
+  feature?: string;
```
saikumarrs, line 166: These should not be optional.

src/types/index.ts
```diff
@@ -0,0 +1,213 @@
...
+  IDENTIFY = "identify",
+  TRACK = "track",
+  PAGE = "page",
+  SCREEN = "screen",
+  GROUP = "group",
+  ALIAS = "alias",
+  AUDIENCE_LIST = "audiencelist"
+}
+
+type RudderMessage = {
```
saikumarrs, line 148: We're missing `timestamp` property.

src/services/destination/nativentegration.ts
```diff
@@ -0,0 +1,197 @@
+import groupBy from "lodash/groupBy";
```
saikumarrs, line 1: There is a typo in this file's name.

src/types/index.ts
```diff
@@ -0,0 +1,213 @@
...
+  Transformations: UserTransformationInput[];
+  RevisionID: string;
+};
+
+type UserTransformationLibrary = {
+  VersionID: string;
+};
+
+type ProcessorTransformRequest = {
+  request?: Object;
```
saikumarrs, line 81: What does `request` contain?

src/types/index.ts
saikumarrs, line 1: As we are doing a great job of defining all the types, let's also add documentation for the relevant ones to improve maintainability.

src/types/index.ts
utsabc: This majorly used by WH destination where they are parsing data from query parameters and passing here (Since I was not aware of exact types hence I kept open)

src/services/destination/nativentegration.ts
utsabc: Fixed

src/types/index.ts
saikumarrs: Ok. Please add some documentation.

src/util/serviceSelector.ts
utsabc: I have relocated these into respective classes but I have kept sourceHandler and destinationHandler different as there is not major overlap on native handlers

src/controllers/destination.controller.ts
utsabc: @koladilip can you confirm this?

src/controllers/destination.controller.ts
utsabc: @koladilip any further thoughts?

src/util/dynamicConfigInjector.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  private static getDynamicConfig(event: ProcessorRequest | RouterRequestData) {
+    const resultantEvent = cloneDeep(event);
+    const { Config } = event.destination;
+    resultantEvent.destination.Config = this.configureVal(Config, event);
+    return resultantEvent;
+  }
+
+  public static processDynamicConfig(
```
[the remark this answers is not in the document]
saikumarrs: ?

Dockerfile
```diff
@@ -1,40 +1,70 @@
-# syntax=docker/dockerfile:1.4
```
saikumarrs, line 1: Restore this line.

Dockerfile
```diff
@@ -1,40 +1,70 @@
...
+ENV transformer_build_version=$version
+ENV git_commit_sha=$GIT_COMMIT_SHA
+
 # Create app directory
+WORKDIR /home/node/app
+ADD . .
+RUN chown -R node:node /home/node/app
+USER node
+
+RUN npm install
```
saikumarrs, line 28: This should be `npm ci`

Dockerfile
```diff
@@ -1,40 +1,70 @@
-# syntax=docker/dockerfile:1.4
-FROM node:14.21.1-alpine3.15
+FROM node:14.21.1-alpine3.15 AS base
```
saikumarrs, line 1: Can you expand on the changes in this file?

src/util/dynamicConfigInjector.ts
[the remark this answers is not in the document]
utsabc: Will addressing it will push it

Dockerfile
utsabc: You can refer more here https://docs.docker.com/build/building/multi-stage/
On high level you can name a stage and re-use that particular stage

saikumarrs: I was not asking about that but more about the need for different build types, etc.

utsabc: On high level this is what I am achieving

Updated Dockerfile to support multistage build, have broken build to two stages
- development [Everything is present including dev node modules, tests, and src files along with dist]
 -  Will be used primarily for building and testing purposes
- production Only prod node modules and dist is present
 - Drastically reduces build size
Updated all build spec
- rudderlabs/develop-rudder-transformer Will now contain the larger development builds which can be directly tested also can be run
- rudderlabs/rudder-transformer Will now contain the production image which is much smaller
Even though we are pushing images without tests in production builds but we are testing development counter-parts of the same build before pushing

src/controllers/profile.controller.ts
utsabc: This is addressed

src/types/index.ts
```diff
@@ -0,0 +1,213 @@
...
+  message: Object;
+  metadata: Metadata;
+  destination: Destination;
+  libraries: UserTransformationLibrary[];
+};
+
+type RouterTransformRequestData = {
```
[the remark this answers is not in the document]
utsabc: We already have `RouterTransformRequest` which encompases this type

src/types/index.ts
utsabc: Addressed

src/controllers/regulation.controller.ts
utsabc: This is handled now with refactoring work

src/util/dynamicConfigInjector.ts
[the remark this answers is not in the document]
utsabc: addressed

src/types/index.ts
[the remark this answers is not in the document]
utsabc: addressed

src/types/index.ts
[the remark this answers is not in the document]
utsabc: addressed

src/types/index.ts
```diff
@@ -0,0 +1,191 @@
...
+
+type Destination = {
+  ID: string;
+  Name: string;
+  DestinationDefinition: DestinationDefinition;
+  Config: Object;
+  Enabled: boolean;
+  WorkspaceID: string;
+  Transformations: UserTransformation[];
+  RevisionID: string;
```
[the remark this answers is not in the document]
utsabc: addressed

Dockerfile
utsabc: Addressed

Dockerfile
utsabc: Addressed

src/types/index.ts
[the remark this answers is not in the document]
utsabc: I have revert this change as this was causing some issues we can iterative look back at this one I will be creating a action items for the ones I am leaving behind in this PR

src/types/index.ts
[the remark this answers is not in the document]
utsabc: Lets add it in next iteration I need some help on this

src/util/utils.js
utsabc: Added for future extension

src/types/index.ts
utsabc: Addressed

src/types/index.ts
```diff
@@ -0,0 +1,213 @@
+type TransformedEvent = {
+  version: string;
+  type: string;
+  method: string;
+  endpoint: string;
+  userId: string;
+  headers: Record<string, unknown>;
```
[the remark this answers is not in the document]
utsabc: Addressed

src/types/index.ts
```diff
@@ -0,0 +1,213 @@
...
+  destination: Destination;
+};
+
+type RouterTransformRequest = {
+  input: RouterTransformRequestData[];
+  destType: string;
+};
+
+type ProcessorTransformResponse = {
+  output?: TransformedEvent | RudderMessage | Object;
```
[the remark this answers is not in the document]
utsabc: For User transform scenario we need this you can check UserTransform.ts under services for detailed info on high level in this module we are expecting a RudderMessage in place of transformed event in case of error we are returning object

This module is currently getting worked on by JC need to review if we can scope a update for this

src/types/index.ts
utsabc: What about `errors` ?

utsabc: Or functions ?

src/types/index.ts
utsabc: Addressed

src/types/index.ts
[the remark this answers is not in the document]
utsabc: ref: https://github.com/rudderlabs/rudder-transformer/pull/1710/commits/dbe84b69e94dc33c4353f40fa06ff4f6114e1382

src/types/index.ts
utsabc: Yes we can but we have to think answers to some questions
- For whom will these docs be helpful
- Is it for external views as well
- Are we sure the type definitions we have defined are stable and not going to change any time soon?

src/controllers/destination.ts
```diff
@@ -0,0 +1,169 @@
...
+        ctx,
+      ) as ProcessorTransformationRequest[];
+      resplist = await integrationService.processorRoutine(
+        events,
+        destination,
+        version,
+        requestMetadata,
+      );
+    } catch (error) {
+      resplist = events.map((ev) => {
```
koladilip, line 44: Is it possible to create a function for this as it is repeated?

src/controllers/routeActivation.ts
```diff
@@ -0,0 +1,148 @@
...
+const transformerTestModeEnabled = process.env.TRANSFORMER_TEST_MODE
+  ? process.env.TRANSFORMER_TEST_MODE.toLowerCase() === 'true'
+  : false;
+const areFunctionsEnabled = process.env.ENABLE_FUNCTIONS === 'false' ? false : true;
+const destinationFilterList = process.env.DESTINATION_FILTER_LIST?.toLocaleLowerCase();
+const sourceFilteList = process.env.SOURCE_FILTER_LIST?.toLocaleLowerCase();
+const deliveryFilterList = process.env.DESTINATION_DELIVERY_FILTER_LIST?.toLocaleLowerCase();
+
+export default class RouteActivationController {
+  public static isDestinationRouteActive(ctx: Context, next: any) {
```
koladilip, line 22: next doesn't have a type?

src/controllers/routeActivation.ts
```diff
@@ -0,0 +1,148 @@
...
+    if (!startDestTransformer) {
```
koladilip, line 23: all the below functions can be written using one helper function.

```js
function isRouteActive(condition: boolean, ctx: Context, next: any) {
    if (!condition) {
      ctx.status = 404;
      ctx.body = 'This route is disabled'
      return ctx;
    } else {
      return next();
    }
}
```

src/controllers/routeActivation.ts
```diff
@@ -0,0 +1,148 @@
...
+      ctx.body = 'This route is disabled';
+      return ctx;
+    } else {
+      return next();
+    }
+  }
+
+  public static destinationProcFilter(ctx: Context, next: any) {
+    const { destination }: { destination: string } = ctx.params;
+    if (
```
koladilip, line 84: condition is repeated move to function

src/helpers/fetchHandlers.ts
```diff
@@ -0,0 +1,40 @@
+import MiscService from '../services/misc';
+
+export default class FetchHandler {
+  private static sourceHandlerMap: Map<string, any> = new Map();
+  private static destHandlerMap: Map<string, any> = new Map();
+  private static deletionHandlerMap: Map<string, any> = new Map();
+
+  public static getDestHandler(dest: string, version: string) {
+    let destinationHandler: any;
+    if (this.destHandlerMap.get(dest)) {
```
koladilip, line 10: if(this.destHandlerMap.has(dest)) change all the place where this pattern is relevant

src/helpers/serviceSelector.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  private static isCdkV2Destination(destinationDefinitionConfig: Object) {
+    return !!destinationDefinitionConfig?.['cdkV2Enabled'];
+  }
+
+  private static fetchCachedService(serviceType: string) {
+    let service: any;
+    if (this.serviceMap.has(serviceType)) {
+      service = this.serviceMap.get(serviceType);
```
koladilip, line 24: use return stmt here so that we can rewrite else block without else

src/interfaces/DestinationService.ts
```diff
@@ -0,0 +1,54 @@
...
+  init(): void;
+
+  getTags(
+    destType: string,
+    destinationId: string,
+    workspaceId: string,
+    feature: string,
+  ): MetaTransferObject;
+
+  processorRoutine(
```
koladilip, line 23: Did we discuss naming conventions for these?
routerRoutine -> doRouterTransform?
processorRoutine -> doProcessorTransform?

src/services/destination/postTransformation.ts
```diff
@@ -0,0 +1,151 @@
...
+  ProcessorTransformationOutput,
+  DeliveryResponse,
+  MetaTransferObject,
+  UserDeletionResponse,
+} from '../../types/index';
+import { generateErrorObject } from '../../v0/util';
+import ErrorReportingService from '../errorReporting';
+
+export default class PostTransformationDestinationService {
+  public static handleSuccessEventsAtProcessorDest(
```
koladilip, line 17: handleSuccessEventsAtProcessorDest -> handleProcessTransformSucessEvents?

src/controllers/delivery.ts
```diff
@@ -0,0 +1,61 @@
...
+      );
+    } catch (error) {
+      const metaTO = integrationService.getTags(
+        destination,
+        event.metadata?.destinationId || 'Non-determininable',
+        event.metadata?.workspaceId || 'Non-determininable',
+        tags.FEATURES.DATA_DELIVERY,
+      );
+      metaTO.metadata = event.metadata;
+      deliveryResponse = PostTransformationServiceDestination.handleFailureEventsAtDeliveryDest(
```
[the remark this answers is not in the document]
utsabc: This is addressed

src/controllers/delivery.ts
```diff
@@ -0,0 +1,61 @@
...
+export default class DeliveryController {
+  public static async deliverToDestination(ctx: Context) {
+    logger.debug('Native(Delivery):: Request to transformer::', JSON.stringify(ctx.request.body));
+    let deliveryResponse: DeliveryResponse;
+    let requestMetadata = MiscService.getRequestMetadata(ctx);
+    let event = ctx.request.body as ProcessorTransformationOutput;
+    const { version, destination }: { version: string; destination: string } = ctx.params;
+    const integrationService = ServiceSelector.getNativeDestinationService();
+    try {
+      deliveryResponse = await integrationService.deliveryRoutine(
```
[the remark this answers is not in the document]
utsabc: This is addressed

src/controllers/delivery.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  public static async testDestinationDelivery(ctx: Context) {
+    logger.debug(
+      'Native(Delivery-Test):: Request to transformer::',
+      JSON.stringify(ctx.request.body),
+    );
+    const { version, destination }: { version: string; destination: string } = ctx.params;
+    const { deliveryPayload, destinationRequestPayload } = ctx.request.body as any;
+    const response = await DeliveryTestService.deliverTestRoutine(
```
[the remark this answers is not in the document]
utsabc: This is addressed

src/controllers/delivery.ts
```diff
@@ -0,0 +1,61 @@
...
+    return ctx;
+  }
+
+  public static async testDestinationDelivery(ctx: Context) {
+    logger.debug(
+      'Native(Delivery-Test):: Request to transformer::',
+      JSON.stringify(ctx.request.body),
+    );
+    const { version, destination }: { version: string; destination: string } = ctx.params;
+    const { deliveryPayload, destinationRequestPayload } = ctx.request.body as any;
```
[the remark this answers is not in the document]
utsabc: Added types

src/controllers/destination.ts
```diff
@@ -0,0 +1,169 @@
...
+    const metaTags = MiscService.getMetaTags(events[0].metadata);
+    const { version, destination }: { version: string; destination: string } = ctx.params;
+    const integrationService = ServiceSelector.getDestinationService(events);
+    try {
+      integrationService.init();
+      events = PreTransformationDestinationService.preProcess(
+        events,
+        ctx,
+      ) as ProcessorTransformationRequest[];
+      resplist = await integrationService.processorRoutine(
```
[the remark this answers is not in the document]
utsabc: Addressed

src/controllers/destination.ts
utsabc: Actually the processor and router handling has difference we can create two functions but already we are using `getTags` other aspect is different for proc and router flow for adding metadata infos

src/controllers/routeActivation.ts
utsabc: handled

src/controllers/routeActivation.ts
utsabc: handled

src/helpers/fetchHandlers.ts
utsabc: this is handled

src/helpers/serviceSelector.ts
utsabc: updated

src/interfaces/DestinationService.ts
utsabc: Yes updated

src/services/destination/postTransformation.ts
utsabc: addressed

koladilip (dismissed):
Looks good to me

src/controllers/routeActivation.ts
utsabc: addressed

.github/workflows/build-pr-artifacts.yml
```diff
@@ -52,14 +52,15 @@ jobs:
...
+          target: development
           load: true
           tags: |
             ${{ env.REPO_NAME }}:${{ env.TAG_NAME }}
 
       - name: Run Tests
         run: |
           docker run ${{ env.REPO_NAME }}:${{ env.TAG_NAME }} npm run test:ci
-
+          docker run ${{ env.REPO_NAME }}:${{ env.TAG_NAME }} npm run test:integration
```
saikumarrs, line 63: We cannot directly run integration tests. It needs more setup (see test.yml).

utsabc: This is not the user transform integration tests these are `service.integration.test.js` which is running test e2e with supertest

saikumarrs: Got it, but why are we explicitly ignoring those tests in the config file?

utsabc: Because we want to have separate build and deployment for user-transform images

saikumarrs: I don't think we'd discard that functionality from DT images anytime soon. Migration has to happen for all the users (including open-source).
So, we still have to run those tests on the DT images. Try to find a different way of doing it.

src/controllers/routeActivation.ts
```diff
@@ -0,0 +1,105 @@
+import { Context, Next } from 'koa';
```
sanpj2292, line 1: Not sure if this file should be in `controllers` as I feel this acts more like a `middleware`. I'd change the location of this file to `src/middlewares/routeActivation.ts`. Please see if this change is applicable and make the changes

utsabc: Got it will address

src/controllers/routeActivation.ts
```diff
@@ -0,0 +1,105 @@
...
+    );
+  }
+
+  public static destinationBatchFilter(ctx: Context, next: Next) {
+    const routerRequest = ctx.request.body as RouterTransformationRequest;
+    const destination = routerRequest.destType;
+    return RouteActivationController.executeActivationRule(
+      ctx,
+      next,
+      Array.isArray(destinationFilterList) &&
```
sanpj2292, line 82: As `destinationFilterList` will always be a string(taken from env). `Array.isArray(destinationFilterList)` will always evaluate to `false`, so finally the condition would evaluate to `false`.
Even when you don't want a particular destination to use new routes, since the condition evaluates to `false` it will go via new routes

Please let me know if my analysis is right.

I would suggest to change to this
```suggestion
        !destinationFilterList.split(',').includes(destination.toLowerCase())
```

src/controllers/routeActivation.ts
```diff
@@ -0,0 +1,105 @@
...
+  ? process.env.TRANSFORMER_TEST_MODE.toLowerCase() === 'true'
+  : false;
+const areFunctionsEnabled = process.env.ENABLE_FUNCTIONS === 'false' ? false : true;
+const destinationFilterList = process.env.DESTINATION_FILTER_LIST?.toLocaleLowerCase();
+const sourceFilteList = process.env.SOURCE_FILTER_LIST?.toLocaleLowerCase();
+const deliveryFilterList = process.env.DESTINATION_DELIVERY_FILTER_LIST?.toLocaleLowerCase();
+
+export default class RouteActivationController {
+  private static executeActivationRule(ctx: Context, next: Next, condition: boolean) {
+    if (condition) {
```
sanpj2292, line 23: I'd rename the name of variable `condition` as `dontActivateRoute`
Or 
For easiness, I'd replace condition itself in below way

```javascript
// In this way the caller should change the condition
private static executeActivationRule(ctx: Context, next: Next, shouldActivate: boolean) {
    if (shouldActivate) {
      return next();
    }
    ctx.status = 404;
    ctx.body = 'RouteActivationController route is disabled';
    return ctx;
  }
````

But this can be written other way round as well
```javascript
private static executeActivationRule(ctx: Context, next: Next, dontActivate: boolean) {
    if (!dontActivate) {
      return next();
    }
    ctx.status = 404;
    ctx.body = 'RouteActivationController route is disabled';
    return ctx;
  }

// OR

private static executeActivationRule(ctx: Context, next: Next, shouldActivate: boolean) {
    if (dontActivate) {
       ctx.status = 404;
      ctx.body = 'RouteActivationController route is disabled';
      return ctx;
    }
    return next();
  }
```

src/controllers/routeActivation.ts
utsabc: Yes make sense

utsabc: Addressed

src/controllers/routeActivation.ts
utsabc: Addressed

src/controllers/routeActivation.ts
utsabc: Addressed

src/controllers/source.ts
```diff
@@ -0,0 +1,38 @@
...
+      const metaTO = integrationService.getTags();
+      const resp = PostTransformationServiceSource.handleFailureEventsSource(err, metaTO);
+      ctx.body = [resp];
+    }
+    ControllerUtility.postProcess(ctx);
+    logger.debug(
+      'Native(Source-Transform):: Response from transformer::',
+      JSON.stringify(ctx.body),
+    );
+    return ctx;
```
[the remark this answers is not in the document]
utsabc: I didn;t get you are you asking why are we not returning `ctx.body`?

src/helpers/serviceSelector.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  private static getSecondaryServiceName(destinationDefinitionConfig: Object): string {
+    return destinationDefinitionConfig['camparisonService'];
+  }
+
+  private static fetchCachedService(serviceType: string) {
+    if (this.serviceMap.has(serviceType)) {
+      return this.serviceMap.get(serviceType);
+    }
+    switch (serviceType) {
```
sanpj2292, line 38: I am seeing same set of values which match the cases. Rather than this, we can achieve this in a very simple manner

```javascript
const validServicesMap = {
 INTEGRATION_SERVICE.CDK_V1_DEST: CDKV1DestinationService,
 INTEGRATION_SERVICE.CDK_V2_DEST: CDKV2DestinationService,
 INTEGRATION_SERVICE.NATIVE_DEST: NativeIntegrationDestinationService,
 INTEGRATION_SERVICE.NATIVE_SOURCE: NativeIntegrationSourceService
}

private static fetchCachedService(serviceType: string) {
    if (!this.serviceMap.has(serviceType)) {
      // One more variation -> You can evaluate serviceTypeMap.keys() only once as it is not going to change in run-time
      if (serviceTypeMap.has(serviceType)) {
        const Service = serviceTypeMap[serviceType];
        this.serviceMap.set(
          INTEGRATION_SERVICE.NATIVE_DEST,
          new Service(),
        );
      }
    }
    return this.serviceMap.get(serviceType);
}
```

sanpj2292: A similar pattern can be implemented wherever such kind of switch cases or if-else combinations are seen. The goal to make sure that the logic doesn't look too bulky and is very readable. Please check if this pattern can be applied anywhere else 🙂

src/middlewares/routeActivation.ts
```diff
@@ -0,0 +1,97 @@
+import { Context, Next } from 'koa';
```
sanpj2292, line 1: The logic for destinationProcFilter, delivery is same & the logic for BatchFilter, RtFilter is same. We might need to commonise the logic accordingly and proceed. WDYT ?

src/routes/obs.delivery.js
```diff
@@ -1,6 +1,14 @@
...
+ * --------------------------------------
+ * ---------TO BE DEPRICIATED------------
+ * --------------------------------------
+ * --------------------------------------
+ */
+
 const path = require('path');
-const { DestProxyController } = require('../controllers/destinationProxy');
+const KoaRouter = require('@koa/router');
+const { DestProxyController } = require('../controllers/obs.delivery');
```
sanpj2292, line 11: Just asking, what does `obs` mean in this context ?. Feel free to resolve the comment, once you're done providing explanation

src/middlewares/routeActivation.ts
utsabc: I intentionally kept these in different functions even though its code repetition but it helps in modifying specific logic in case need arises in future.

src/routes/obs.delivery.js
utsabc: It means Obsolete we have https://github.com/rudderlabs/rudder-transformer/blob/8e26b5046b5756e33bc4350f7798715300d5f64a/src/controllers/delivery.ts#L11 as replacement for these

src/services/comparator.ts
```diff
@@ -0,0 +1,389 @@
...
+  ): MetaTransferObject {
+    const metaTO = {
+      errorDetails: {
+        destType: destType.toUpperCase(),
+        module: tags.MODULES.DESTINATION,
+        implementation: this.primaryService.getName(),
+        feature,
+        destinationId,
+        workspaceId,
+        context: '[Native Integration Service] Failure During Processor Transform',
```
sanpj2292, line 56: Can you let me know what is the reason for having `context` as one of the labels `statTags` ?

utsabc: It is used in bugsnag, but I think this will also populate our metrics. Ok need to segregate this

src/services/destination/cdkV2Integration.ts
```diff
@@ -0,0 +1,152 @@
...
+  ): RouterTransformationResponse[] {
+    throw new TransformationError('CDKV2 Does not Implement Batch Transform Routine');
+  }
+
+  public deliver(
+    _event: ProcessorTransformationOutput,
+    _destinationType: string,
+    _requestMetadata: Object,
+  ): Promise<DeliveryResponse> {
+    throw new TransformationError('CDV2 Does not Implement Delivery Routine');
```
sanpj2292, line 143:
```suggestion
    throw new TransformationError('CDKV2 Does not Implement Delivery Routine');
```

sanpj2292: Please correct similar cases

src/services/destination/postTransformation.ts
```diff
@@ -0,0 +1,156 @@
+import cloneDeep from 'lodash/cloneDeep';
```
sanpj2292, line 1: This file looks a little too bulky to me. I mean I felt hard to read. If we can split this into something like below
```typescript
abstract class AbstractPostTransformationHandler {
 static success(...args: any[]): any | any[];
 static failure(...args: any[]): any;
}

// src/services/destination/postTransformation/processor.ts
class ProcessorPostTransformationHandler extends  AbstractPostTransformationHandler{
 success(a, b, c) {
  // your logic
 }
 failure(error, d, e) {
  // your logic
 }
}
```
Similarly we can move each of the success, failure handlers to their specific files.
The goal is to make the file more readable and easily understandable.

Let me know your thoughts

sanpj2292: If making this change is a concern wrt time or something we can do this later as well. I feel this'd be much better approach & if we can do this now, it'll be nice 🙂

src/services/source/postTransformation.ts
```diff
@@ -0,0 +1,34 @@
+import { MetaTransferObject, RudderMessage, SourceTransformationResponse } from '../../types/index';
```
sanpj2292, line 1: Same approach as put forth [here](https://github.com/rudderlabs/rudder-transformer/pull/1710/files#r1152741141)

src/helpers/serviceSelector.ts
utsabc: This is addressed

src/middlewares/routeActivation.ts
utsabc: Yep addressed

src/services/comparator.ts
utsabc: This is also fixed

src/services/destination/cdkV2Integration.ts
utsabc: Yep addressed

koladilip (dismissed):

src/helpers/serviceSelector.ts
```diff
@@ -52,7 +58,11 @@ export default class ServiceSelector {
...
 
-  public static getDestinationService(
+  private static getDestinationServiceByName(name: string): DestinationService {
+    return this.fetchCachedService(name);
+  }
+
+  private static getPrimaryDestinationService(
     events: ProcessorTransformationRequest[] | RouterTransformationRequestData[],
   ): DestinationService {
     const destinationDefinitionConfig: Object = events[0].destination.DestinationDefinition.Config;
```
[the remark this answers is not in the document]
sanpj2292: ➕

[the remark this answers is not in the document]
utsabc: Can you elaborate a bit on this?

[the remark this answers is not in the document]
sanpj2292: Instead of directly accessing the keys from the object, @koladilip is recommending of using optional chaining while accessing the keys in the object

src/helpers/serviceSelector.ts
```diff
@@ -6,41 +6,47 @@ import DestinationService from '../interfaces/DestinationService';
...
-          INTEGRATION_SERVICE.NATIVE_DEST,
-          new NativeIntegrationDestinationService(),
-        );
-        break;
-      case INTEGRATION_SERVICE.NATIVE_SOURCE:
-        this.serviceMap.set(
-          INTEGRATION_SERVICE.NATIVE_SOURCE,
-          new NativeIntegrationSourceService(),
-        );
+    const Service = this.services[serviceType];
```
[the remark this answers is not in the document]
sanpj2292: I think this variable represents the service-type class hence started with uppercase letter

src/controllers/userTransform.ts
```diff
@@ -98,4 +98,42 @@ export default class UserTransformController {
...
+        code,
+        versionId,
+        validateImports,
+        additionalLibraries,
+        language,
+        testMode || versionId === 'testVersionId',
+      );
+      ctx.body = obj;
+    } catch (err: any) {
+      ctx.status = 400;
```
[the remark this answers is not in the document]
utsabc: @koladilip these parts of the code we haven't really changed these are existing code only. As @Jayachand will be taking up a separate refactoring work for the UT flow we kept it as its

src/helpers/serviceSelector.ts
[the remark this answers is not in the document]
utsabc: Yes I was on the same line with @sanpj2292

src/util/libExtractor.js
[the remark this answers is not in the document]
utsabc: Not sure @Jayachand can you check>?

src/util/customTransformer-faas.js
```diff
@@ -1,21 +1,85 @@
...
+const libVersionIdsCache = new NodeCache();
+
+function generateFunctionName(userTransformation, libraryVersionIds, testMode) {
+  if (userTransformation.versionId === FAAS_AST_VID) return FAAS_AST_FN_NAME;
 
-function generateFunctionName(userTransformation, testMode) {
   if (testMode) {
-    const funcName = `fn-test-${userTransformation.testName?.replace('_', '-')}-${uuidv4()}`;
+    const funcName = `fn-test-${uuidv4()}`;
     return funcName.substring(0, 63).toLowerCase();
```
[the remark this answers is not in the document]
utsabc: This is UT specific code @Jayachand is it ok if we acommodate this suggestion?

src/helpers/serviceSelector.ts
```diff
@@ -68,4 +78,24 @@ export default class ServiceSelector {
...
     // Implement source event based descision logic for selecting service
   }
+
+  public static getDestinationService(
+    events: ProcessorTransformationRequest[] | RouterTransformationRequestData[],
+  ): DestinationService {
+    const destinationDefinition = events[0].destination.DestinationDefinition;
+    const destinationDefinitionConfig = destinationDefinition.Config;
+    const primaryService = this.getPrimaryDestinationService(events);
+    if (this.isComparatorEnabled(destinationDefinitionConfig)) {
```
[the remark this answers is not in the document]
utsabc: Addressed

src/helpers/serviceSelector.ts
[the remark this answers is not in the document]
utsabc: Addressed

sanpj2292 (approved):

koladilip (approved):

PR #1715, opened 2022-12-19
fix: update the metadata fetch code to send to bugsnag

utsabc:
## Description of the change

Update the get metadata code to fetch info
- jobId
- destType
- destConfig

For each request failure to bugsnag

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/src/versionedRouter.js
+++ b/src/versionedRouter.js
@@ -1235,6 +1235,17 @@ const handleDeletionOfUsers = async ctx => {
   const getReqMetadata = () => {
     try {
       const reqBody = ctx.request.body;
+      if (Array.isArray(reqBody)) {
+        return {
+          jobInfos: reqBody.map(req => {
+            return {
+              jobId: req.jobId,
+              destType: req.destType,
+              destConfig: req.config
+            };
+          })
+        };
+      }
       return { destType: reqBody?.destType };
     } catch (error) {
       // Do nothing


src/versionedRouter.js
```diff
@@ -1235,6 +1235,17 @@ const handleDeletionOfUsers = async ctx => {
   const getReqMetadata = () => {
     try {
       const reqBody = ctx.request.body;
+      if (Array.isArray(reqBody)) {
+        return {
+          jobInfos: reqBody.map(req => {
+            return {
+              jobId: req.jobId,
+              destType: req.destType,
+              destConfig: req.config
```
krishna2020, line 1244: Is entire destConfig required to return ?
Is it fine to send jobInfos[] in case of reqBody being an array vs {destType} in case of reqBody not being an array ?
Is it not possible to make reqBody consistently an array ?

utsabc: Yeah I was having second thougths sending destConfig is lot of info but it makes it easier to check in bugsnag. Actually server always sends reqBody as array only but we were wrong not considering it, I kept the code backward compatible.

Though better to keep a Array.isarray check (since server is another service) being extra precocious.

saikumarrs: Destination config contains sensitive information. We should not send it to Bugsnag.

saikumarrs: @utsabc fixed the code on your behalf.

saikumarrs (approved):

src/versionedRouter.js
utsabc: Thanks

PR #1740, opened 2022-12-29
refactor: salesforce test-cases with mocks

sanpj2292:
Signed-off-by: Sai Sankeerth <sanpj2292@github.com>

## Description of the change

We have implemented tests for `salesforce` in a way that is more extensible and common for `processor` & `router` transformation for all the destinations.
If this pattern looks good, we can plan to migrate to this kind of testing pattern for all future destinations

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/test/__tests__/transformation/proc/proc.test.js
+++ b/test/__tests__/transformation/proc/proc.test.js
@@ -1,0 +1,39 @@
+const { readdirSync } = require("fs");
+const path = require("path");
+const { formAxiosMock } = require("../../../__mocks__/gen-axios.mock");
+
+const destinations = readdirSync(path.join(__dirname, "data"), {
+  withFileTypes: true
+})
+  .filter(dirent => dirent.isDirectory())
+  .map(dirent => dirent.name);
+
+destinations.forEach(destination => {
+  const inputData = require(`./data/${destination}/input.json`);
+  const expectedData = require(`./data/${destination}/output.json`);
+  const {
+    process
+  } = require(`../../../../src/v0/destinations/${destination}/transform`);
+
+  if (process) {
+    describe(`${destination} Processor Tests`, () => {
+      beforeAll(() => {
+        const axiosResponses = inputData
+          .filter(inp => Array.isArray(inp.mockNetworkResponses))
+          .map(inp => inp.mockNetworkResponses);
+        formAxiosMock(axiosResponses);
+      });
+
+      inputData.forEach(async (input, index) => {
+        it(`Payload - ${index}`, async () => {
+          try {
+            const output = await process(input);
+            expect(output).toEqual(expectedData[index]);
+          } catch (error) {
+            expect(error.message).toEqual(expectedData[index].error);
+          }
+        });
+      });
+    });
+  }
+});
--- a/test/__tests__/data/salesforce_input.json
+++ b/test/__tests__/data/salesforce_input.json
@@ -180,24 +180,6 @@
     },
     "mockNetworkResponses": [
       {
-        "type": "post",
-        "reqParams": [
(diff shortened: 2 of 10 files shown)


test/__tests__/transformation/proc/proc.test.js
```diff
@@ -0,0 +1,39 @@
...
+      });
+
+      inputData.forEach(async (input, index) => {
+        it(`Payload - ${index}`, async () => {
+          try {
+            const output = await process(input);
+            expect(output).toEqual(expectedData[index]);
+          } catch (error) {
+            expect(error.message).toEqual(expectedData[index].error);
+          }
```
saikumarrs, line 34: This is a much better approach as the test case input dictates the expectations, and also, we'll be using the Jest framework components.
```suggestion
          if (expectedData[index].error) {
            expect(() => {
              await process(input);
              }).toThrow(expectedData[index].error);
          } else {
            const output = await process(input);
            expect(output).toEqual(expectedData[index]);
          }
```

saikumarrs (commented):
LGTM, but @ujjwal-ab is already working on streamlining the test suite. You can pass on these learnings to him to avoid duplication of work.

For now, I'd recommend fixing the existing test suite to add the missing cases, if possible.

sanpj2292:
> For now, I'd recommend fixing the existing test suite to add the missing cases, if possible.

Missing case which is causing the `Cannot read property 'access_token' of undefined` cannot be added without making changes to existing `axios.mock.js` & it is quite hard to do it.

Hence actually made complete U-turn & implement everything from scratch kind-of

test/__tests__/transformation/proc/proc.test.js
sanpj2292: Makes sense 👍🏼

saikumarrs:
> > For now, I'd recommend fixing the existing test suite to add the missing cases, if possible.
> 
> Missing case which is causing the `Cannot read property 'access_token' of undefined` cannot be added without making changes to existing `axios.mock.js` & it is quite hard to do it.
> 
> Hence actually made complete U-turn & implement everything from scratch kind-of

Ok. @ujjwal-ab, let's account for this format in your investigation and come up with a final approach for the overall refactoring.

ujjwal-ab:
> > > For now, I'd recommend fixing the existing test suite to add the missing cases, if possible.
> > 
> > 
> > Missing case which is causing the `Cannot read property 'access_token' of undefined` cannot be added without making changes to existing `axios.mock.js` & it is quite hard to do it.
> > Hence actually made complete U-turn & implement everything from scratch kind-of
> 
> Ok. @ujjwal-ab, let's account for this format in your investigation and come up with a final approach for the overall refactoring.

Yeah, this is kind of what we are thinking. But not sure, is it possible here to run proc/router tests for a single destination if we want to?

sanpj2292:
> > > > For now, I'd recommend fixing the existing test suite to add the missing cases, if possible.
> > > 
> > > 
> > > Missing case which is causing the `Cannot read property 'access_token' of undefined` cannot be added without making changes to existing `axios.mock.js` & it is quite hard to do it.
> > > Hence actually made complete U-turn & implement everything from scratch kind-of
> > 
> > Ok. @ujjwal-ab, let's account for this format in your investigation and come up with a final approach for the overall refactoring.
> 
> Yeah, this is kind of what we are thinking. But not sure, is it possible here to run proc/router tests for a single destination if we want to?

You can run single test bybusing specific jest command using regex while executing jest command

sanpj2292:
Implemented another way of mocking network requests such that we don't need to split the test-cases for processor & router transformations. The changes can be seen in [this commit](https://github.com/rudderlabs/rudder-transformer/pull/1740/commits/38dca1550bfbd7dae7319551e34b03d6bbba9a3a)

PR #1794, opened 2023-01-17
feat: ga4 page calls are discarded if hybrid mode is enabled

mihir-4116:
## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


src/v0/destinations/ga4/transform.js
```diff
@@ -428,8 +428,12 @@ const process = (event) => {
       break;
     case EventType.PAGE:
```
saikumarrs, line 429: Do we still need to handle this case when we already filter it with allowed message types in the definition config?

src/v0/destinations/ga4/transform.js
```diff
@@ -428,8 +428,12 @@ const process = (event) => {
...
       // GA4 custom event 'page_view' is fired for page
-      message.event = 'page_view';
-      response = responseBuilder(message, destination);
+      if (!isHybridModeEnabled(Config)) {
+        message.event = 'page_view';
+        response = responseBuilder(message, destination);
+      }else{
+        throw new InstrumentationError("GA4 Hybrid mode is enabled, page calls will be sent through device mode")
```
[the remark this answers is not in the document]
krishna2020: Create new error class - `UnsupportedEventError` for this filtering logic

src/v0/destinations/ga4/transform.js
mihir-4116: yes, @saikumarrs because there are customers who are using ga4 cloud mode page calls so we can't remove that support. right now this is a temporary solution but we need to come up with permanent solution which is server should know if hybrid mode is enabled then for a destination which calls should pass through cloud mode and which not

mihir-4116: Also can you check above comment by @krishna2020 can we do something like that?

saikumarrs: Yeah, let's proceed.

krishna2020 (approved):

ItsSudip (approved):

PR #1809, opened 2023-01-19
chore(warehouse): remove setVarCharMax form redshift

achettyiitr:
## Description of the change

> Removed setVarCharMax for redshift.

> https://www.notion.so/rudderstacks/Figure-out-way-to-set-varchar-max-for-new-customers-Establish-if-it-is-ok-from-cost-impact-perfo-809f7548af144562ad0aafa4e5035f75

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

- [x] "Ready for review" label attached to the PR and reviewers mentioned in a comment
- [x] Changes have been reviewed by at least one other engineer
- [x] Issue from task tracker has a link to this pull request

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,34 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.12.0](https://github.com/rudderlabs/rudder-transformer/compare/v1.11.1...v1.12.0) (2023-01-19)
+
+
+### Features
+
+* **BQStream:** add batch support ([#1377](https://github.com/rudderlabs/rudder-transformer/issues/1377)) ([14c7531](https://github.com/rudderlabs/rudder-transformer/commit/14c7531635b5348ef518dcad483f25d4adeddddd))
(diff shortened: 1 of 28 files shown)


koladilip (approved):

lvrach:
Should we remove https://github.com/rudderlabs/rudder-transformer/blob/a33adc6c12a4f7cd6b62955bc29d58206034b3c4/v0/destinations/rs/transform.js#L16-L22 as well ?

```
  if (options.rsAlterStringToText === "true" && val) {
    const stringifiedVal = Array.isArray(val) ? JSON.stringify(val) : val;
    if (stringifiedVal.length > RSStringLimit) {
      return "text";
    }
  }
```

achettyiitr:
> Should we remove
> 
> https://github.com/rudderlabs/rudder-transformer/blob/a33adc6c12a4f7cd6b62955bc29d58206034b3c4/v0/destinations/rs/transform.js#L16-L22
> 
> as well ?
> ```
>   if (options.rsAlterStringToText === "true" && val) {
>     const stringifiedVal = Array.isArray(val) ? JSON.stringify(val) : val;
>     if (stringifiedVal.length > RSStringLimit) {
>       return "text";
>     }
>   }
> ```

We need this for backward compatibility. 

- For newer customers, we will be creating varchar(65535) columns so we don't have an issue with the datatype sent by the transformer whether it is string or text.
- For older customers, 
  - If we make the default override as text, then we will be converting all the varchar(512) columns to varchar(65535) columns which may not be required.
  - We can't make the default override a string. because for Alter to happen it needs to be text column.

lvrach (approved):

abhimanyubabbar (approved):

PR #1832, opened 2023-01-30
chore: refactor utility function test scripts

saikumarrs:
## Description of the change

- Created a standard test script driver for all the utility functions (src/v0/utils). Other utils for now have their own test scripts.
- Migrated the existing independent test scripts to the new format.

All the test data files are named after the corresponding utility function names. Now, adding tests for a new function is simply 

**What still needs to be done?**
Update the existing test case descriptions to make them more intuitive. However, I've tried that to some extent.

## Type of change

## Related issues

N/A

### Development

- [ ] Lint rules pass locally

### Code review 


test/__tests__/__integration_util_test__/testHelper.js
```diff
@@ -0,0 +1,12 @@
+const fs = require('fs');
+const path = require('path');
+
+const getFuncTestData = (funcName) => {
+  const fileData = fs.readFileSync(path.resolve(__dirname, `./data/${funcName}.json`));
```
sanpj2292, line 5: Why don't we just require a json and make use of them based on the function name ? Not dynamically but more like below
```javascript
const getFormData = require("/path/to/getFormData.json");

const funcTestMap = {
 getFormData,
 ...
};
const getFuncTestData = (funcName) => {
 return funcTestMap[funcName];
};
```

saikumarrs: Yeah, we could do that. As the files are read just once during the test execution, I didn't go to that extent.
Do you see any major advantage over the current approach with the one you suggested?

utsabc (approved):

test/__tests__/__integration_util_test__/testHelper.js
sanpj2292: - Easier to understand
- Simpler to implement and extend
- Don't have to do read and parse all the time

saikumarrs: Since this is for a test script, I would care less about the above aspects (except for readability). Also, "static requires" would mean that the function names are tightly coupled at two places now, here and in the main driver test script.
I plan to remove the dependency of names within the driver test script in the first place (I will do that in the next round of refactoring).

I imagined this function to be an independent utility that returns the test case data but is not like a cache. Let's ignore it and move ahead.

sanpj2292: Please continue as you please

PR #1862, opened 2023-02-13
feat(integrations): mocker for migration of destinations to transformer proxy

sanpj2292:
## Description of the change

This PR contains changes to aid in testing migration of a destination to transformer proxy
https://www.notion.so/rudderstacks/Plan-for-automation-of-destination-migration-to-proxy-d87c456eaf464a6eac815559e139c5d1

## Type of change
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally
- [ ] The code changed/added as part of this pull request has been covered with tests
- [ ] All tests related to the changed code pass in development

### Code review 

- [ ]  This pull request has a descriptive title and information useful to a reviewer. There may be a screenshot or screencast attached
- [ ] "Ready for review" label attached to the PR and reviewers mentioned in a comment
- [ ] Changes have been reviewed by at least one other engineer
- [ ] Issue from task tracker has a link to this pull request

PR #1918, opened 2023-03-04
chore: unit-test fw for data-delivery using transformer proxy

sanpj2292:
## Description of the change

We have in this PR, created an easy way to write tests for delivering data to destination using transformer proxy.
Changes:
- The approach taken is to mock the `proxy` method that we define for every network handler
- To make sure we mock, there are couple of changes needed
   1. We need to move the method `prepareProxyRequest` method from `proxy` method itself
   2. We need to make changes to how we define `proxy` in every networkHandler
      a. We need to also make sure that `proxy` is added to the function's prototype if networkHandler is defined using function
      b. We need to make sure that `proxy` is defined as an instance method if networkHandler is defined using class semantics

**Note**:
- Changes from #1862 have creeped into this PR. Kindly please ignore them
- This PR will fail for `google_adwords*` destinations, as they have multiple axios calls being made and preparation of request is being done as part of the logic in the respective `proxy` methods.

#### What do we want out of this PR ?
- Decide on how does the overall schema wrt test-cases for a destination look ?
- Decision on the mocking strategy implemented
- Understanding on the strategy implemented
- Arrive at a consensus about having multiple API calls support in `proxy` method(in-future) -- Not my preferred way

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


koladilip:
Looks good, should we make this change on top of @utsabc changes if not urgent?

PR #2041, opened 2023-04-05
feat(clevertap): ignore stringification for objects for Charged event

ujjwal-ab:
## Description of the change
Ignore stringification for objects for Charged event for Clevertap

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


src/v0/destinations/clevertap/transform.js
```diff
@@ -291,7 +291,7 @@ const responseBuilderSimple = (message, category, destination) => {
     }
     eventPayload.type = 'event';
     // stringify the evtData if it's an Object or array.
-    if (eventPayload.evtData) {
+    if (eventPayload.evtData && eventPayload.evtName !== 'Charged') {
```
ItsSudip, line 294: We have to skip `Items` property only from stringification. All other properties should be a string. And we should also check if `Items` property value is an array of objects.

ujjwal-ab: Added check

src/v0/destinations/clevertap/transform.js
```diff
@@ -79,12 +79,21 @@ const responseWrapper = (payload, destination) => {
...
+        if (
+          event === 'Charged' &&
+          key === 'Items' &&
+          Array.isArray(payload[key]) &&
+          typeof payload[key][0] === 'object'
+        ) {
+          finalPayload[key] = payload[key];
+        } else {
+          finalPayload[key] = JSON.stringify(payload[key]);
+        }
```
ItsSudip, line 96: Rather than checking for each element, we can check outside of the for loop and if the event is charged and the key is items we can non-stringify that.

ujjwal-ab: But in this way we can't check if the Items is an array of objects or not.

ItsSudip: After non-stringifying we can check right?

ujjwal-ab: Done

ItsSudip (approved):

sandeepdsvs (approved):

PR #2086, opened 2023-04-20
fix: shopify default check for anonymousId

anantjain45823:
## Description of the change

> We are changing the behaviour of default case of mapping the `anonymousId` as there are certain events for which `cartToken` is not present and so we are now checking there is `userId` or not present and mapping `userId` as `shopify-admin` if `userId` is not present. 
Added some metrics for data transaction failure between transformer and redis. 
Added `userName` in `redisConnector`

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/src/v0/sources/shopify/util.js
+++ b/src/v0/sources/shopify/util.js
@@ -135,6 +135,7 @@ const setAnonymousIdorUserIdFromDb = async (message, metricMetadata) => {
      */
     case RUDDER_ECOM_MAP.checkouts_create:
     case RUDDER_ECOM_MAP.checkouts_update:
+    case SHOPIFY_TRACK_MAP.checkouts_delete:
     case SHOPIFY_TRACK_MAP.orders_cancelled:
     case SHOPIFY_TRACK_MAP.orders_fulfilled:
     case SHOPIFY_TRACK_MAP.orders_paid:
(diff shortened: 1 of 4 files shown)


koladilip:
Why are we doing this change?

utsabc (dismissed):

src/v0/sources/shopify/util.js
```diff
@@ -140,42 +141,31 @@ const setAnonymousIdorUserIdFromDb = async (message, metricMetadata) => {
...
     case SHOPIFY_TRACK_MAP.carts_update:
-      cartToken = message.properties?.id || message.properties?.token;
+      cartToken = properties?.id || properties?.token;
       break;
     // https://help.shopify.com/en/manual/orders/edit-orders -> order can be edited through shopify-admin only
     // https://help.shopify.com/en/manual/orders/fulfillment/setting-up-fulfillment -> fullfillments wont include cartToken neither in manual or automatiic
     case SHOPIFY_TRACK_MAP.orders_edited:
     case SHOPIFY_TRACK_MAP.orders_delete:
     case SHOPIFY_TRACK_MAP.fulfillments_create:
     case SHOPIFY_TRACK_MAP.fulfillments_update:
```
ItsSudip, line 159: If there is not any computation to do we can remove these cases.

ItsSudip (changes requested):

src/v0/sources/shopify/util.js
anantjain45823: Done

ItsSudip (approved):

ItsSudip:
> Why are we doing this change?

For some events, we are not getting the cartToken as those are generated due to admin activity. Let's say the `user disabled` event will occur when an admin abandons a user. So to identify those events we are adding `shopify-admin` as userId. With the previous release we could not handle this logic properly so we are making these changes.

yashasvibajpai (approved):

utsabc (approved):

PR #2134, opened 2023-05-08
fix: handle 404 response code for proxy route

utsabc:
## Handle 404 response code for proxy route

Added constants for status code and update the response code for 404 response for proxy route


sanpj2292 (dismissed):

saikumarrs (changes requested):
Let's handle this in the older code base, also, as we're still maintaining both versions now.
CC @krishna2020

src/v0/util/constant.js
```diff
@@ -10,9 +10,79 @@ const API_CALL = 'api_call_count';
 
 const JSON_MIME_TYPE = 'application/json';
 
+const HttpStatus = {
```
saikumarrs, line 13:
```suggestion
const HTTP_STATUS_MAP = {
```

In the future, this should become an enum.

src/controllers/util/index.ts
```diff
@@ -1,15 +1,23 @@
 import { Context } from 'koa';
 import { API_VERSION } from '../../routes/utils/constants';
 import { isHttpStatusSuccess } from '../../v0/util';
+import { HttpStatus } from '../../v0/util/constant';
 
 export default class ControllerUtility {
+  private static handleForNotfoundStatus(status: number) {
+    if (status === HttpStatus.NOT_FOUND) {
+      return HttpStatus.BAD_REQUEST;
```
saikumarrs, line 9: To generalize, we can term it as a bad request but for the destination under discussion, shouldn't it be a configuration error?

src/controllers/util/index.ts
```diff
@@ -1,15 +1,23 @@
 import { Context } from 'koa';
 import { API_VERSION } from '../../routes/utils/constants';
 import { isHttpStatusSuccess } from '../../v0/util';
+import { HttpStatus } from '../../v0/util/constant';
 
 export default class ControllerUtility {
+  private static handleForNotfoundStatus(status: number) {
```
koladilip, line 7: make this function generic to handle other status codes

sanpj2292: Done.

src/controllers/util/index.ts
saikumarrs: I meant this status code translation should happen within that destination (Slack). The existing one suffices for the generic handler.

koladilip (approved):

src/controllers/util/index.ts
sanpj2292: Yes, ideally we should handle in specific destination's handlers explicitly.

src/adapters/network.js
```diff
@@ -6,8 +6,9 @@ const http = require('http');
 const https = require('https');
 const axios = require('axios');
 const log = require('../logger');
-const { removeUndefinedValues } = require('../v0/util');
+const { removeUndefinedValues, isHttpStatusSuccess } = require('../v0/util');
```
krishna2020, line 9: remove unused imports

src/v0/util/constant.js
saikumarrs: Some more alternatives I can think of:
- HTTP_STATUS_CODES
- HTTP_STATUSES

src/controllers/util/index.ts
```diff
@@ -1,15 +1,19 @@
 import { Context } from 'koa';
 import { API_VERSION } from '../../routes/utils/constants';
-import { isHttpStatusSuccess } from '../../v0/util';
+import { getCompatibleStatusCode } from '../../adapters/utils/networkUtils';
 
 export default class ControllerUtility {
+  private static getCompatibleStatusCode(status: number): number {
```
saikumarrs, line 6: Is there a plan to use this API anywhere outside this class? Otherwise, we can remove this and use the utility directly.

sanpj2292: At a future we want to have `getCompatibleStatusCode` in this utility itself. We have only added this for supporting the javascript routes(`versionedRouter.js`)

sanpj2292:
>https://github.com/rudderlabs/rudder-transformer/pull/2134#discussion_r1188136038

Done this

src/adapters/network.js
sanpj2292: Removed

sanpj2292 (approved):

PR #2139, opened 2023-05-08
fix: undefined errors in braze, firehose, gaoc

yashasvibajpai:
## Description of the change

> Fix 'of undefined' errors in Braze, Firehose, GAOC.
> Added support for retryable error in tiktok ads.

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


src/v0/destinations/braze/transform.js
```diff
@@ -169,8 +169,8 @@ function getUserAttributesObject(message, mappingJson, destination) {
         if (destKey === 'gender') {
           value = formatGender(value);
         }
-        if (destKey === 'email') {
-          value = value.toLowerCase();
+        if (destKey === 'email' && value !== null) {
+          value = value?.toLowerCase();
```
utsabc, line 173: Why optional chaining after null check.
Also shouldn't we be checking for undefined?

src/v0/destinations/firehose/transform.js
```diff
@@ -6,7 +6,7 @@ function getDeliveryStreamMapTo(event) {
   const { mapEvents } = destination.Config;
   const hashMap = getHashFromArray(mapEvents, 'from', 'to');
   return (
-    (message.event ? hashMap[message.event.toLowerCase()] : null) ||
+    (message.event ? hashMap[message?.event.toLowerCase()] : null) ||
```
utsabc, line 9: We are already checking validity of event. Why do we need optional chaining here? 
Maybe this issue could be type issue maybe `event` is not a string

src/v0/destinations/google_adwords_offline_conversions/utils.js
```diff
@@ -92,7 +92,7 @@ const getConversionActionId = async (headers, params) => {
     searchStreamResponse = processAxiosResponse(searchStreamResponse);
     if (!isHttpStatusSuccess(searchStreamResponse.status)) {
       throw new AbortedError(
-        `[Google Ads Offline Conversions]:: ${searchStreamResponse.response[0].error.message} during google_ads_offline_conversions response transformation`,
+        `[Google Ads Offline Conversions]:: ${searchStreamResponse.response[0].error?.message} during google_ads_offline_conversions response transformation`,
```
utsabc, line 95: Just send a generic error message with `JSON.stringify(searchStreamResponse.response)`

utsabc (changes requested):

src/v0/destinations/firehose/transform.js
yashasvibajpai: In metadata I see that eventName is an empty string but with that this check would not have failed. Should I remove it? The error was intermittent for a short while and not occurring now.

src/v0/destinations/braze/transform.js
yashasvibajpai: As per [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining), a null value will short-circuit with the value of `undefined`. Hence the check for null. And for email to be undefined, the first check should work.

src/v0/destinations/firehose/transform.js
utsabc: Better to check  `(isString(message.event) and message.event.length != 0 ? hashMap[message?.event.toLowerCase()] : null)`

yashasvibajpai:
@utsabc Also added support for retryable errors in case of tiktok ads.

src/v0/destinations/tiktok_ads/networkHandler.js
```diff
@@ -15,6 +15,9 @@ const responseHandler = (destinationResponse) => {
       destinationResponse,
     };
   }
+  if (code === 40100) {
+    throw new RetryableError(`Request failed with status: ${code}`, 500, destinationResponse);
+  }
```
utsabc, line 20: `"code":40100,"message":"Too many requests. Please retry in some time`

status code should be 429

utsabc (changes requested):

src/v0/destinations/tiktok_ads/networkHandler.js
yashasvibajpai: Addressed. Put 500 in lieu of this being some server issue over at Tiktok, but since it is retryable, it should be 429.

utsabc (dismissed):

src/v0/destinations/braze/transform.js
```diff
@@ -301,7 +302,7 @@ function getPurchaseObjs(message) {
 
   const purchaseObjs = [];
 
-  if (products) {
+  if (isArray(products)) {
```
sanpj2292, line 305: We can `Array.isArray`(the native one) instead of lodash right ?

sanpj2292: If that satisfies our use-case, let's try to use native one. Also as a rule of thumb, let's native wherever it satisfies our criteria and depend on only external libraries under extreme circumstances

ItsSudip (dismissed):

src/v0/destinations/braze/transform.js
yashasvibajpai: Yes sure.

ItsSudip (approved):

utsabc (approved):

ItsSudip (approved):

PR #2166, opened 2023-05-11
chore: add depreciation for old routes

utsabc:


koladilip (dismissed):

saikumarrs:
A deprecation notice is usually added for the APIs and classes exposed to the users consuming them. It's a good candidate for utility methods, etc. The modified file is not used anywhere apart from the entry script.

You probably want to rename the current file(s) and move it under the `legacy` directory so that users unaware of the latest changes cannot find it easily. Add comments in the entry script to further explain the situation.

src/versionedRouter.js
```diff
@@ -1,3 +1,6 @@
+// =============================================================================
+// DEPRICIATION NOTICE: THIS FILE IS GETTING DEPRECATED AND WILL BE REMOVED IN FUTURE RELEASE
```
saikumarrs, line 2:
```suggestion
// DEPRECATION NOTICE: THIS FILE IS GETTING DEPRECATED AND WILL BE REMOVED IN FUTURE RELEASE
```

saikumarrs (approved):

koladilip (approved):

PR #2215, opened 2023-05-29
feat: remove custom property dependency for fb pixel

yashasvibajpai:
## Description of the change

> We are removing the property mapping that was mandatory in order to send the mapped properties here inside Other Standard events of out FB Pixel integration.

Removing this, will bring standard events in line with custom events, in terms of properties being sent along with the event.

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/src/v0/destinations/facebook_pixel/utils.js
+++ b/src/v0/destinations/facebook_pixel/utils.js
@@ -132,7 +132,6 @@ const transformedPayloadData = (
   blacklistPiiProperties,
   whitelistPiiProperties,
   isStandard,
-  eventCustomProperties,
   integrationsObj,
 ) => {
   const defaultPiiProperties = [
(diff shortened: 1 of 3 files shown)


Gauravudia (approved):

src/v0/destinations/facebook_pixel/utils.js
```diff
@@ -192,10 +184,6 @@ const transformedPayloadData = (
         delete clonedCustomData[eventProp];
       }
     }
-    const isCustomProperty = customEventProperties[eventProp] || false;
-    if (isStandard && !isCustomProperty && !isDefaultPiiProperty) {
-      delete clonedCustomData[eventProp];
-    }
```
ItsSudip, line 198: Could you please let me know why you're removing this entirely? It does not seem the same with device mode implementation.

ItsSudip, line 198: Could you please let me know why you're removing this entirely? It does not seem the same with device mode implementation.

yashasvibajpai: In line 195 we are checking if the current property being iterated is a mapped property for standard event. (This mapping is removed hence we are removing this. )
Refer removed line 173

> `finalEventCustomProperties.forEach((property) => {
>     const singularConfigInstance = property;
>     customEventProperties[singularConfigInstance.eventCustomProperties] = true;
>   });`

Now in the next check, we were deleting this custom property if the event was a standard event, the property is not mapped in the now removed mapping and the property is not a defaultPiiProperty. SInce [FB supports passing any custom property for standard events](https://developers.facebook.com/docs/marketing-api/conversions-api/parameters/custom-data#custom-properties), removing this condition

ItsSudip (approved):

PR #2263, opened 2023-06-13
fix(klaviyo): add check for listId for subscribe

ujjwal-ab:
## Description of the change

Add check for `listId` before creating the payload for subscribe endpoint if the subscribe flag is true(user wants to subscribe the profile)

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/src/v0/destinations/klaviyo/transform.js
+++ b/src/v0/destinations/klaviyo/transform.js
@@ -99,8 +99,11 @@ const identifyRequestHandler = async (message, category, destination) => {
   // Update Profile
   const responseArray = [profileUpdateResponseBuilder(payload, profileId, category, privateApiKey)];
 
-  // check if user wants to subscribe profile or not
-  if (traitsInfo?.properties?.subscribe) {
+  // check if user wants to subscribe profile or not and listId is present or not
+  if (
+    traitsInfo?.properties?.subscribe &&
+    (traitsInfo.properties?.listId || destination.Config?.listId)
+  ) {
     responseArray.push(subscribeUserToList(message, traitsInfo, destination));
     return responseArray;
   }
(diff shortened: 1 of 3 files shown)


utsabc (approved):

src/v0/destinations/klaviyo/transform.js
```diff
@@ -99,8 +99,11 @@ const identifyRequestHandler = async (message, category, destination) => {
...
-  // check if user wants to subscribe profile or not
-  if (traitsInfo?.properties?.subscribe) {
+  // check if user wants to subscribe profile or not and listId is present or not
+  if (
+    traitsInfo?.properties?.subscribe &&
+    (traitsInfo.properties?.listId || destination.Config?.listId)
+  ) {
     responseArray.push(subscribeUserToList(message, traitsInfo, destination));
     return responseArray;
   }
```
krishna2020, line 109: why only first element of `responseArray` being returned ?

ujjwal-ab: Here, for identify call if user has sent the subscribe flag as true, then two payloads will be formed and we are returning them in an array, but if subscribe flag is false then we have only one payload and we are returning the same using reponseArray[0]

krishna2020 (approved):

src/v0/destinations/klaviyo/transform.js
krishna2020: Much of the code functionality is dependent on the array length & the position of an element in array which is bad practice. 
And also, a single function having only one response type is always recommended. Here based on the scenario - either the array (of updateProfile & subscribe) or object (updateProfile) is being sent. Instead it should always return an array and have a flag in the object to determine whether it's updateProfile or subscribe
@ujjwal-ab Take this as backlog & come up with the refactoring design

ujjwal-ab: Will refactor it with Unit tests

PR #2301, opened 2023-06-22
fix: braze source transform timestamp conversion issue

koladilip:
## Description of the change
Fixes: Incorrect Timestamp conversion in Braze source

## Type of change

## Related issues

> Fix [#1](https://www.notion.so/rudderstacks/Incorrect-Timestamp-conversion-in-Braze-source-35fda81df8e24a25b3cb0501b6e6d883?pvs=4) 

### Development

- [ ] Lint rules pass locally

### Code review 


src/v0/sources/braze/transform.js
```diff
@@ -40,7 +40,9 @@ const processEvent = (event) => {
 
     // set timestamp for the event
     if (event.time) {
-      message.setProperty('timestamp', formatTimeStamp(event.time, 'YYYY-MM-DD[T]HH:mm:ss.SSS[Z]'));
+      // Refer: https://www.braze.com/docs/user_guide/data_and_analytics/braze_currents/event_glossary/message_engagement_events
+      const timestamp = parseInt(event.time, 10) * 1000;
```
sanpj2292, line 44: I was trying to do similar thing. WDYT of the below changes instead of this
`v0/util/index.js`
```javascript
// Important !@!
// Generic timestamp formatter
const formatTimeStamp = (dateStr, format) => {
  let date = new Date(dateStr);
  if (_.isNumber(dateStr)) {
    date = new Date(dateStr * 1000)
  }
  // moment format is passed. format accordingly
  if (format) {
    return moment.utc(date).format(format);
  }

  // return default format
  return date.getTime();
};
```

sanpj2292: This would any such issues for all who are using `formatTimeStamp` function. Let me know your thoughts @koladilip

koladilip: but in common util we can't assume that timestamp always comes in seconds right? in case of braze we know for sure that they will be sending seconds

sanpj2292: True. The condition `_.isNumber` handles it for us as well, if a number(like epoch) is sent, it would be correctly formatted. That was my thinking

sanpj2292: Feel free to ignore this comment

sanpj2292 (approved):
Changes LGTM

src/v0/sources/braze/transform.js
koladilip: timestamp could in seconds or mills seconds right, even though it is number, I wouldn't assume that it will be always seconds in common util.

sanpj2292 (approved):

utsabc (approved):

krishna2020 (approved):

PR #2307, opened 2023-06-26
fix(mixpanel): timestamp error handling

sanpj2292:
## Description of the change

- added a try catch block for getTimeEvent function for mixpanel destination.
- added two test cases for the same.

Hotfix version of #2276 

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/src/v0/util/index.js
+++ b/src/v0/util/index.js
@@ -47,7 +47,13 @@ const flattenMap = (collection) => _.flatMap(collection, (x) => x);
 // GENERIC UTLITY
 // ========================================================================
 
-const getEventTime = (message) => new Date(message.timestamp).toISOString();
+const getEventTime = (message) => {
+  try {
+    return new Date(message.timestamp).toISOString();
+  } catch (err) {
+    return new Date(message.originalTimestamp).toISOString();
+  }
+};
 
 const base64Convertor = (string) => Buffer.from(string).toString('base64');
 
@@ -1276,7 +1282,8 @@ function toTitleCase(payload) {
       .replace(/([a-z])(\d)/gi, '$1 $2')
       .replace(/(\d)([a-z])/gi, '$1 $2')
       .trim()
-      .replace(/(_)/g, ` `).replace(/(?:^|\s)(\w)/g, (match) => match.toUpperCase());
+      .replace(/(_)/g, ` `)
+      .replace(/(?:^|\s)(\w)/g, (match) => match.toUpperCase());
     newPayload[newKey] = value;
   });
   return newPayload;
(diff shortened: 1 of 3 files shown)


test/__tests__/data/mp_output.json
```diff
@@ -2033,5 +2033,103 @@
...
+            "$screen_height": 2984,
+            "$screen_width": 1440,
+            "$session_id": "1662363980",
+            "$wifi": true,
+            "anonymousId": "39da706ec83d0e90",
+            "build": 4,
+            "distinct_id": "39da706ec83d0e90",
+            "mp_device_model": "sdk_gphone64_arm64",
+            "mp_lib": "com.rudderstack.android.sdk.core",
+            "time": null,
```
krishna2020, line 2070: shouldn't `time` load with originalTimestamp field value with the fallback logic introduced now ?

sanpj2292: The field that you pointed here, is something else
The fix made as part of the bugsnag issue was to populate the field `$time` in the `params.data.$append[]`
The field `time` pointed here is making use of `message.timestamp` itself. But in order to populate this field, we have to make changes elsewhere. If it is confirmed to make changes, I can make them

sanpj2292: There are other parts of the code where in `timestamp` is being used. We need to also change it but for now. I can make the changes for only `track` call alone.

@krishna2020 Can you confirm if `time` also needs to be change such that when `timestamp` is not proper, we need to use `originalTimestamp` ?

krishna2020 (approved):

src/v0/util/index.js
```diff
@@ -47,7 +47,13 @@ const flattenMap = (collection) => _.flatMap(collection, (x) => x);
 // GENERIC UTLITY
 // ========================================================================
 
-const getEventTime = (message) => new Date(message.timestamp).toISOString();
+const getEventTime = (message) => {
+  try {
+    return new Date(message.timestamp).toISOString();
+  } catch (err) {
+    return new Date(message.originalTimestamp).toISOString();
```
koladilip, line 54: @sanpj2292  last time we discussed that we shouldn't use original timestamp post server, why are we using here?

sanpj2292: I think that was the fix & it was decided by the team. I just raised a PR for hotfix
@Sanjay-Veernala or @krishna2020  can you answer @koladilip 's query ?

krishna2020: It's being used only as fallback, otherwise nothing goes.

koladilip (dismissed):

sanpj2292:
This is not needed as we have include these changes in #2318

src/v0/util/index.js
krishna2020: In the cases where user sends `timestamp` as part of payload, server passes it as-is. So in case of empty timestamp being sent in the payload, this scenario can occur.
Instead rudder-server can validate & put the current timestamp otherwise

PR #2394, opened 2023-07-21
feat(customerio): send url instead of name

mauhai:
## Description of the change

Found a small but impactful ‘bug’ in your customerio destination.
https://www.rudderstack.com/docs/sources/event-streams/sdks/rudderstack-javascript-sdk/supported-api/
—> defines name as name of the page
https://www.customer.io/docs/sdk/web/events/#page-view-events
—> defines name as url of the page

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


anantjain45823:
Hey @mauhai, I looked into their docs as well as your query and here are some pointers from my side:
1. **customerIo** accepts `custom` page name as well and **not only url** . They map url to page name in case we do not send anything to them while making the page call. Example:
  ```
  _cio.page("ABC",{
        url:"https://abc.com",
        referrer: "some url"  
      });
```
Above call will look like this in dashbaord: 
Screenshot 2023-07-28 at 11 46 09 AM

2. For your query that it is a bug that we do not send `properties.url` as priority would be held `valid` if they would have used url format only in page name and since they do not I don't think there is any requirement for change here.
If there are any concerns LMK otherwise I will close this PR

anantjain45823:
In addition to this why we use name over url is due to ease of use, flexibility and more consistent and standardized.

PR #2421, opened 2023-07-31
chore: outgoing request latency metric

chandumlg:
## Description of the change

Adding a new metric to measure the outgoing requests latency.

Tags:
1. integration: Which destination transformation is making the request
2. type: What path is the request made in. This can take the following three values
    a. proxy
    b. transformation
    c. deleteUsers

Introducing a new wrapper (myAxios) around axios. This is temporary. We already have network.js which we should be using for all the network calls. Once we migrate all the network calls, myAxios can be deleted safely.

## Type of change

### Development

- [ ] Lint rules pass locally

### Code review 


utsabc (approved):

sanpj2292:
@chandumlg Can we rename the labels from `integration` to `destType` & `type` to `feature` ?

src/adapters/network.js
```diff
@@ -72,13 +86,17 @@ const httpSend = async (options) => {
  *
  * handles http GET requests returns promise as a response throws error in case of non 2XX statuses
  */
-const httpGET = async (url, options) => {
+const httpGET = async (url, options, statTags = {}) => {
```
koladilip, line 89: can use httpSend method for all httpGET, httpPOST etc? @sanpj2292

src/util/myAxios.js
```diff
@@ -0,0 +1,86 @@
...
+  let response;
+  try {
+    response = await axios(options);
+  } finally {
+    fireLatencyStat(startTime, statTags);
+  }
+  return response;
+};
+
+const get = async (url, options, statTags = {}) => {
```
koladilip, line 24: why do we need both network adapter and myAxios? can we combine both?

chandumlg:
> @chandumlg Can we rename the labels from `integration` to `destType` & `type` to `feature` ?

Sure. If everyone agrees. Will wait a day before I make this change. (This has to be changed in a lot of places :D)

src/util/myAxios.js
chandumlg: Like I mentioned in the PR description, myAxios is temporary. Once all the network calls are migrated to use network adapter, myAxios can be removed safely.

Having this metric asap can be helpful. This can be a quick win.

chandumlg:
> > @chandumlg Can we rename the labels from `integration` to `destType` & `type` to `feature` ?
> 
> Sure. If everyone agrees. Will wait a day before I make this change. (This has to be changed in a lot of places :D)

Done.

utsabc (approved):

sanpj2292:
`feature` has to be one of constants mentioned [here](https://github.com/rudderlabs/rudder-transformer/blob/develop/src/v0/util/tags.js#L30-L36)
We might need to think through how to pass it along to the http functions.

src/adapters/network.js
sanpj2292: Yes. But that's different to the intention of this PR. We can probably take it up separately

koladilip (dismissed):

chandumlg:
> `feature` has to be one of constants mentioned [here](https://github.com/rudderlabs/rudder-transformer/blob/develop/src/v0/util/tags.js#L30-L36) We might need to think through how to pass it along to the http functions.

Makes sense. But again, that needs some effort. Can we go ahead with this and fix tags later? There shouldn't be any problem with updating tag values later for this metric.

sanpj2292 (approved):

koladilip (approved):

PR #2431, opened 2023-08-04
fix(INT-256): handling non float price input for braze purchase events

utsabc:
## Description of the change
Braze purchase events require the price value to be float. This commit is purposed to solve that

### DIFF
--- a/src/v0/destinations/braze/braze.util.test.js
+++ b/src/v0/destinations/braze/braze.util.test.js
@@ -1,6 +1,6 @@
 const _ = require('lodash');
 const { handleHttpRequest } = require('../../../adapters/network');
-const { BrazeDedupUtility, addAppId } = require('./util');
+const { BrazeDedupUtility, addAppId, getPurchaseObjs } = require('./util');
 const { processBatch } = require('./util');
 const {
   removeUndefinedAndNullValues,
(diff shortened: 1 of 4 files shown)


mihir-4116 (approved):

src/v0/destinations/braze/transform.js
```diff
@@ -326,7 +326,7 @@ function getPurchaseObjs(message) {
         }
         let purchaseObj = addMandatoryPurchaseProperties(
           productId,
-          price,
+          parseFloat(price),
```
mihir-4116, line 329: is this scenario covered in test cases ?

mihir-4116 (approved):

src/v0/destinations/braze/transform.js
utsabc: Added tests

ItsSudip (approved):

mihir-4116 (approved):

src/v0/destinations/braze/braze.util.test.js
```diff
@@ -986,3 +986,237 @@ describe('addAppId', () => {
     expect(addAppId(payload, message)).toEqual(payload);
   });
 });
+
```
krishna2020, line 989: No positive test case for Shopify scenario ? 
Price: "10.99"

PR #2448, opened 2023-08-07
fix: valid cart_update event extra check

anantjain45823:
## Description of the change

> Resolves INT-164
We are receiving `cart_update` event generated by rudderstack tracker before `rudderIdentifier` could reach us from tracker which is resulting in no redis mapping and we are assuming the event to be valid but since it is a rudderstack generated event it should not be a valid event and we should drop it.
To deal with such scenarios we are introducing some more check which will basically check the following conditions and if all true then will drop the `cart_update` event.

1. Nothing is retrieved from redis for particular cart_token.
2. `line_items` array is empty
3.  Difference between cart `updated_at` and `created_at` less than decided time limit ( currently 10 sec)

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/src/v0/sources/shopify/util.js
+++ b/src/v0/sources/shopify/util.js
@@ -229,7 +229,11 @@ const checkAndUpdateCartItems = async (inputEvent, redisData, metricMetadata) =>
     await updateCartItemsInRedis(cartToken, newCartItemsHash, metricMetadata);
   } else {
     const { created_at, updated_at } = inputEvent;
-    if (Date.parse(updated_at) - Date.parse(created_at) < maxTimeToIdentifyRSGeneratedCall && inputEvent?.line_items?.length === 0) {
+    const timeDifference = Date.parse(updated_at) - Date.parse(created_at);
+    const isTimeWithinThreshold = timeDifference < maxTimeToIdentifyRSGeneratedCall;
+    const isLineItemsEmpty = inputEvent?.line_items?.length === 0;
+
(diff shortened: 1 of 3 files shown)


src/v0/sources/shopify/util.js
```diff
@@ -226,6 +227,11 @@ const checkAndUpdateCartItems = async (inputEvent, redisData, metricMetadata) =>
       return false;
     }
     await updateCartItemsInRedis(cartToken, newCartItemsHash, metricMetadata);
+  } else {
+    const { created_at, updated_at } = inputEvent;
+    if (Date.parse(updated_at) - Date.parse(created_at) < maxTimeToIdentifyRSGeneratedCall && inputEvent?.line_items?.length === 0) {
+      return false;
+    }
```
mihir-4116, line 234:
```
const timeDifference = Date.parse(updated_at) - Date.parse(created_at);
const isTimeWithinThreshold = timeDifference < maxTimeToIdentifyRSGeneratedCall;
const isLineItemsEmpty = inputEvent?.line_items?.length === 0;

if (isTimeWithinThreshold && isLineItemsEmpty) {
    return false;
}
```

mihir-4116 (changes requested):

sanpj2292 (approved):

ItsSudip (approved):

PR #2509, opened 2023-08-23
chore: add label for clevertap, hs, klaviyo, mp

yashasvibajpai:
## Description of the change

> Resolves INT-377
Add endpoint label for latency metric

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


src/v0/util/index.js
```diff
@@ -47,6 +47,12 @@ const flattenMap = (collection) => _.flatMap(collection, (x) => x);
 // GENERIC UTLITY
 // ========================================================================
 
+const getEndpointPath = (endpoint) => {
+  // extract the path details from the whole endpoint url passed
+  const url = new URL(endpoint);
+  return url.pathname.replace(/^\/+/, '');
```
koladilip, line 53: this replaces "/" from url? but there may be dynamic data in the url right like user id or email etc

yashasvibajpai: I used URL.pathname to extract whole of the path after the domain. It returns '/' always at start so handled that here. [Ref](https://developer.mozilla.org/en-US/docs/Web/API/URL/pathname)

krishna2020:
@yashasvibajpai Avoid ticket id in the PR title, `Resolves` keyword in description should suffice.

yashasvibajpai:
> @yashasvibajpai Avoid ticket id in the PR title, `Resolves` keyword in description should suffice.

Somehow did not trigger for me. Is something wrong with the format

krishna2020:
> > @yashasvibajpai Avoid ticket id in the PR title, `Resolves` keyword in description should suffice.
> 
> Somehow did not trigger for me. Is something wrong with the format

No, the linkup happened - you can check the linear ticket. Linear bot commenting the ticket ID has been disabled now.

src/v0/util/index.js
yashasvibajpai: I am replacing the dynamic portions of url with 'xx'. Does this suffice?

src/v0/util/index.js
```diff
@@ -47,6 +47,12 @@ const flattenMap = (collection) => _.flatMap(collection, (x) => x);
 // GENERIC UTLITY
 // ========================================================================
 
+const getEndpointPath = (endpoint) => {
+  // extract the path details from the whole endpoint url passed
+  const url = new URL(endpoint);
+  return url.pathname.replace(/\$%7B(.*?)%7D/g, 'xx').replace(/^\/+/, '');
```
koladilip, line 53: what does this pattern match? we can replace with ""

yashasvibajpai: I matched with the pattern of adding dynamic variables, i.e `${variable_name}`, and replaced with xx.
Replacing it with "" as suggested

yashasvibajpai: Updated to destination specific handling from a common util fxn.

koladilip (dismissed):

src/v0/destinations/hs/util.js
```diff
@@ -506,6 +509,7 @@ const getExistingData = async (inputs, destination) => {
 
   while (checkAfter) {
     const endpoint = IDENTIFY_CRM_SEARCH_ALL_OBJECTS.replace(':objectType', objectType);
+    const endpointPath = `objects/""/search`;
```
krishna2020, line 512: For readability
```suggestion
    const endpointPath = `objects/"<objectType>"/search`;
```

yashasvibajpai: We were avoiding any dynamic values in our endpointPath label so as to get a general count of the endpoint else here for each objectType (which will be created from externalId as HS-"some-object-name" we will have a value and hence the metric will be segregated for all such labels rather than a unified one

sanpj2292: **Point I think @krishna2020 is trying to make**
The endpoint actually might contain some dynamic values in endpoint but the endpoint string we send to metrics doesn't contain details of the dynamic-value
There is a possibility for us to assume that the endpoint in the metric is exactly same as what we send to destination(transformation/proxy).
Hence having a value-indicator in endpoint string we send to metrics is helpful for us

Please correct me if I'm wrong @krishna2020

krishna2020: Yes @sanpj2292 
endpoint string we send to metrics doesn't contain details of the dynamic-value

So just having the placeholder of the variable name instead of "" helps in readability

src/v0/destinations/hs/util.js
```diff
@@ -506,6 +509,7 @@ const getExistingData = async (inputs, destination) => {
 
   while (checkAfter) {
     const endpoint = IDENTIFY_CRM_SEARCH_ALL_OBJECTS.replace(':objectType', objectType);
+    const endpointPath = `objects/:objectType/search`;
```
krishna2020, line 512: `:objectType` won't this fill up the actual value ?
```suggestion
    const endpointPath = `objects/<objectType>/search`;
```

sanpj2292: No, it is just a placeholder

sanpj2292 (approved):

PR #2604, opened 2023-09-13
feat: tslint rules enable

aashishmalik:
## Description of the change

> tslint rules enable in tranformer

> Change type 'Object' to 'object` . 
Ref.  `error`  Don't use `Object` as a type.
- If you want a type meaning "any object", you probably want `object` instead.
- If you want a type meaning "any value", you probably want `unknown` instead.

> Create a type for `any` as wasn't sure on the structure, to remove lint error on further code added to repository . 

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


koladilip (dismissed):

src/controllers/bulkUpload.ts
```diff
@@ -42,16 +56,16 @@ export const fileUpload = async (ctx) => {
   let response;
   try {
     response = await destFileUploadHandler.processFileData(ctx.request.body);
-  } catch (error: any) {
+  } catch (error: CatchErr) {
     response = {
-      statusCode: error?.response?.status || error?.status || 400,
-      error: error.message || 'Error occurred while processing payload.',
+      statusCode: error.response ? error.response.status : 400,
```
sanpj2292, line 61: `error.status` is not required at all ? Did we confirm about this ?

src/controllers/profile.ts
```diff
@@ -30,7 +29,6 @@ export default class ProfileController {
       }'
   */
   public static async profile(ctx: Context) {
-    const requestMetadata = MiscService.getRequestMetadata(ctx);
```
sanpj2292, line 33: Did we remove because it was un-used ?

src/middlewares/routeActivation.ts
```diff
@@ -78,6 +78,7 @@ export default class RouteActivationMiddleware {
     );
   }
 
+  // eslint-disable-next-line sonarjs/no-identical-functions
```
sanpj2292, line 81: Why are we adding this ?

src/services/source/nativeIntegration.ts
```diff
@@ -9,6 +9,8 @@ import PostTransformationServiceSource from './postTransformation';
 import FetchHandler from '../../helpers/fetchHandlers';
 import tags from '../../v0/util/tags';
 import stats from '../../util/stats';
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
sanpj2292, line 12: Why is this line added ? I'm seeing we are using CatchErr and FixMe in this file

src/testRouter.js
```diff
@@ -28,8 +28,8 @@ const isSupportedContentType = (contentType) => {
 
 const testRouter = new Router({ prefix: '/test-router' });
 
-const getDestHandler = (version, destination) =>
-  require(`./${version}/destinations/${destination}/transform`);
+const getDestHandler = (versionNo, destination) =>
```
sanpj2292, line 31: Currently version is not a number but rather a string as `v0`. So I don't think the variable `versionNo` makes sense. Let me know if you think the same.

src/v0/destinations/active_campaign/transform.js
```diff
@@ -337,7 +337,8 @@ const customListProcessor = async (message, category, destination, contactId) =>
   // status information
   // Ref: https://developers.activecampaign.com/reference/update-list-status-for-contact/
   const promises = [];
-  listArr.map(async (li) => {
+  // eslint-disable-next-line no-restricted-syntax
+  for (const li of listArr) {
```
sanpj2292, line 341: Instead of `for...of` loop, we can use map function to put things into promise
```javascript
const promises = listArr.map(async (li) => {
  return someVar;
})
.filter(v => undefined) // We would need to test this in order to understand how we remove undefined promises
```
This should be preferred approach

src/controllers/bulkUpload.ts
aashishmalik: reverted

src/testRouter.js
aashishmalik: changed to ver, version is giving lint issue as version is declared in outer scope

src/middlewares/routeActivation.ts
aashishmalik: there were two different functions which were identical but with different fn names in two different flows, so going with current implementation of maintaining two different fn names with same body to accommodate changes in 1 flow if comes later

src/v0/destinations/active_campaign/transform.js
aashishmalik: I will take `no-restricted-syntax` in another PR, as changes introduced with new develop merge results these in different files. For now going with disabling the rule

sanpj2292 (approved):

src/v0/destinations/mp/util.js
```diff
@@ -143,6 +143,7 @@ const isImportAuthCredentialsAvailable = (destination) =>
 const findExistingBatch = (batch, metadataMap) => {
   let existingBatch = null;
 
+  // eslint-disable-next-line no-restricted-syntax
```
Gauravudia, line 146: can you resolve this? We need to disable the eslint rule on next line?

koladilip (approved):

src/v0/destinations/mp/util.js
aashishmalik: the existing code written with `for of` loop was shoeing this error, needed to update that code . Planning to take these in different PRas develop merge results in this errors for differnt files as well

PR #2647, opened 2023-09-24
feat: use axios-mock-adapter for mocking axios requests

sanpj2292:
## Description of the change

Resolves INT-371

We are making use of axios-mock-adapter library to mock axios requests

- migrate several destinations already in component test suite to this
- migrate all user-deletion destinations to compnent test suite
- remove global axios mock formation automatically
- Corrected several user-deletion test-cases
- Added support to apply mocks(both http requests, function) at test-case level

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


test/integrations/component.test.ts
```diff
@@ -67,80 +68,53 @@ afterAll(async () => {
...
+  mock = new MockAxiosAdapter(axios, { onNoMatch: 'passthrough' });
+  const registerAxiosMocks = (axiosMocks: MockHttpCallsData[]) => {
+    axiosMocks.forEach((axiosMock) => {
+      const { url, method, data: reqData, ...opts } = axiosMock.httpReq;
+      const { data, headers, status } = axiosMock.httpRes;
+
+      switch (method.toLowerCase()) {
+        case 'get':
+          // @ts-ignore
+          mock.onGet(url, reqData, opts.headers).reply(status, data, headers);
```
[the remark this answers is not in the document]
sanpj2292: objectContaining with `axios-mock-adapter` is creating an issue
https://github.com/ctimmerm/axios-mock-adapter/issues/323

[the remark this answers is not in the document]
sanpj2292: Hence implemented in a different way using `asymmetricMatch` function provided by `axios-mock-adapter`

test/integrations/destinations/clevertap/network.ts
```diff
@@ -0,0 +1,284 @@
...
+  {
+    httpReq: {
+      method: 'post',
+      timeout: 60000,
+      withCredentials: false,
+      responseType: 'json',
+      responseEncoding: 'utf8',
+      maxBodyLength: 100000000,
+      maxRedirects: 5,
+      httpAgent: {
```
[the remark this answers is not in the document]
sanpj2292: Removed them. Please check and let me know if all looks good

test/integrations/component.test.ts
```diff
@@ -79,26 +80,32 @@ if (!opts.generate || opts.generate === 'false') {
...
+        asymmetricMatch: function (actual) {
+          return isMatch(actual, opts.headers);
+        },
+      };
+
       switch (method.toLowerCase()) {
         case 'get':
           // @ts-ignore
-          mock.onGet(url, reqData, opts.headers).reply(status, data, headers);
+          mock.onGet(url, reqData, headersAsymMatch).reply(status, data, headers);
```
koladilip, line 92: do we need to same with reqData like headers use headersAsymMatch?
also for url should we use "new RegExp(`(${url}`)" this way it will match contains also.

sanpj2292: Don't think that is required. I think the data match here should be exact as there is no additional transformation being done.

The same logic for URL.

sanpj2292: If there are any cases that we see where in something else is happening inside Axios, we can probably do an `asymmetricMatch`

koladilip: idea here is, we don't want to mention all the details in test cases and only mention required things so that it improves readability

sanpj2292: The point is, the data or url might involve transformational changes where in exactness is more necessary. Headers might contain  content-type or authorization which doesn't involve a lot of complexity. Hence headers having asymmetric match makes sense but having data or url, I'm not 100% convinced about it. But do let me know if you think otherwise

koladilip: we can defer this and add it later if use case arises too

koladilip (approved):

koladilip (approved):

PR #2651, opened 2023-09-25
feat: cm360 batching support

aashishmalik:
## Description of the change

> cm360 batching support

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


src/v0/destinations/campaign_manager/data/CampaignManagerTrackConfig.json
```diff
@@ -12,11 +12,7 @@
   {
     "destKey": "timestampMicros",
     "sourceKeys": "timestamp",
-    "sourceFromGenericMap": true,
-    "required": true,
```
krishna2020, line 16: Is it not mandatory parameter ?

src/v0/destinations/campaign_manager/transform.js
```diff
@@ -138,35 +154,17 @@ function postValidateRequest(response) {
...
-  if (count !== 1) {
+  if (
+    !response.body.JSON.conversions[0].gclid &&
+    !response.body.JSON.conversions[0].matchId &&
+    !response.body.JSON.conversions[0].dclid &&
+    !response.body.JSON.conversions[0].encryptedUserId &&
+    !response.body.JSON.conversions[0].encryptedUserIdCandidates &&
+    !response.body.JSON.conversions[0].mobileDeviceId &&
+    !response.body.JSON.conversions[0].impressionId
+  ) {
```
anantjain45823, line 165: Can we refactor this code to include all these fields inside an array and then validate the payload

src/v0/destinations/campaign_manager/transform.js
```diff
@@ -72,7 +76,19 @@ function processTrack(message, metadata, destination) {
     delete requestJson.childDirectedTreatment;
     delete requestJson.limitAdTracking;
   }
-  requestJson.timestampMicros = requestJson.timestampMicros.toString();
+
+  const date = new Date(requestJson.timestampMicros);
+  let unixTimestamp = date.getTime();
```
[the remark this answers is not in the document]
aashishmalik: If input is `1668624722555000` , the date utility still considers it is in milliseconds and convert this to 054846-08-12T03:22:35.000Z (year 5000) , when we call getTime it return 1668624722555000, so doing no of digits check here.

Same case with `1668624722` it considers it in ms, and getTime return it in `1668624722`, so need to multiply by 6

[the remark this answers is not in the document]
aashishmalik: Also, I think one alternative thing we can do if they pass` timestamp in properties`, we can make it mandatory to pass it in microSeconds and get rid of this conversion at our end

src/v0/destinations/campaign_manager/data/CampaignManagerTrackConfig.json
aashishmalik: Since this field will always be present, so removed this parameter.

src/v0/destinations/campaign_manager/networkHandler.js
```diff
@@ -9,14 +9,36 @@ const { AbortedError, RetryableError, NetworkError } = require('../../util/error
...
+      }] */
+      for (const st of status) {
+        st.errors.forEach(err => {
+          // if code is any of these, event is not retryable
+          if (
+            err.code === 'PERMISSION_DENIED' ||
+            err.code === 'INVALID_ARGUMENT' ||
+            err.code === 'NOT_FOUND'
+          ) {
+            isRetryable = false;
```
ItsSudip, line 36: You can break from here.

src/v0/destinations/campaign_manager/transform.js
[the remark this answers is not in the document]
koladilip:
```js
function convertToMicroseconds(input) {
  const timestamp = Date.parse(input);

  if (!isNaN(timestamp)) {
    // If the input is a valid date string, timestamp will be a number
    if (input.includes("Z")) {
      // ISO 8601 date string with milliseconds
      return timestamp * 1000;
    } else {
      // ISO 8601 date string without milliseconds
      return timestamp * 1000000;
    }
  } else if (/^\d+$/.test(input)) {
    // If the input is a numeric string (assume microseconds or milliseconds)
    if (input.length <= 13) {
      // Length less than or equal to 13 indicates milliseconds
      return parseInt(input) * 1000;
    } else {
      // Otherwise, assume microseconds
      return parseInt(input);
    }
  } else {
    // Invalid input
    throw new Error("Invalid input format");
  }
}
```

// Example usage:
const input1 = "2023-10-11T12:34:56.789Z"; // ISO 8601 date string with milliseconds
const input2 = "1633956896789"; // Milliseconds
const input3 = "1633956896789000"; // Microseconds

console.log(convertToMicroseconds(input1)); // Output: 1633956896789000
console.log(convertToMicroseconds(input2)); // Output: 1633956896789000
console.log(convertToMicroseconds(input3)); // Output: 1633956896789000

This code defines the convertToMicroseconds function, which first checks if the input is a valid date string (ISO 8601 format). If it is, it checks whether the input has milliseconds or not and converts it accordingly. If the input is a numeric string, it checks its length to determine whether it's in milliseconds or microseconds. If the input doesn't match any of these formats, it throws an error for an invalid input format.
Source: chatgpt

src/v0/destinations/campaign_manager/transform.js
```diff
@@ -194,9 +201,110 @@ function process(event) {
...
+  const conversions = [];
+  let encryptionInfo = {};
+  const metadata = [];
+  // extracting destination, message from the first event in a batch
+  const { destination, message } = events[0];
+  // Batch event into dest batch structure
+  events.forEach((ev) => {
+    conversions.push(...ev.message.body.JSON.conversions);
+    metadata.push(ev.metadata);
+    if (ev.message.body.JSON.encryptionInfo) {
```
koladilip, line 215: this will override the encryptionInfo, is this expected?

src/v0/destinations/campaign_manager/transform.js
```diff
@@ -194,9 +201,110 @@ function process(event) {
...
+const generateBatch = (eventKind, events) => {
+  const batchRequestObject = defaultBatchRequestConfig();
+  const conversions = [];
+  let encryptionInfo = {};
+  const metadata = [];
+  // extracting destination, message from the first event in a batch
+  const { destination, message } = events[0];
+  // Batch event into dest batch structure
+  events.forEach((ev) => {
+    conversions.push(...ev.message.body.JSON.conversions);
```
koladilip, line 213: you could keep this ev.message.body.JSON in a variable and reuse it instead of typing long strings

src/v0/destinations/campaign_manager/transform.js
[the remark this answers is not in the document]
aashishmalik: console.log(convertToMicroseconds("2022-11-17T00:22:02.903+05:30"));
o/p -->>1668624722903000000 (in nanosweconds)
will modify this a bit

PR #2658, opened 2023-09-26
fix(dm): updated the stats definition for the tracking plan service

abhimanyubabbar:
## Description of the change

As part of this hotfix, we are fixing the definition of couple of critical stats for tracking plan in the prometheus.js file to align with the stat fired from the service. The consequence of the issue is that we have critical P0 alerts which are not getting fired for the DM team.

Ticket: https://linear.app/rudderstack/issue/DAT-404/hotfix-for-the-user-transformer-service

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/src/util/prometheus.js
+++ b/src/util/prometheus.js
@@ -229,7 +229,7 @@ class Prometheus {
         name: 'tp_errors',
         help: 'tp_errors',
         type: 'counter',
-        labelNames: ['sourceType', 'destinationType', 'k8_namespace', 'workspaceId', 'transformationId'],
+        labelNames: ['sourceType', 'destinationType', 'k8_namespace', 'workspaceId', 'trackingPlanId'],
       },
       {
         name: 'tp_events_count',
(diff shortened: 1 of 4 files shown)


src/util/prometheus.js
```diff
@@ -614,18 +614,6 @@ class Prometheus {
           'k8_namespace',
         ],
       },
-      {
-        name: 'user_transform_function_latency',
```
koladilip, line 618: https://github.com/rudderlabs/rudder-transformer/blob/e7dbd2cc2b040683ca65356bf9d4258b748aaa2c/src/legacy/router.js#L631 this is referred here so it might cause issue if someone is using legacy routes.

abhimanyubabbar: It's legacy router and not used anymore. We plan to remove this file eventually.

Jayachand (dismissed):

sanpj2292 (approved):

src/util/prometheus.js
koladilip: what about vpc customers? may be we won't support UTs for vpc customers right?

PR #2752, opened 2023-10-19
feat: add schema validations for component tests

koladilip:
## Description of the change

Adding schema validations for different APIs, some cases are failing so need some help for that.
Note: for I kept it everything as optional so only data format will be validated if value is present.

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/test/integrations/testTypes.ts
+++ b/test/integrations/testTypes.ts
@@ -1,4 +1,5 @@
 import { AxiosResponse } from 'axios';
+import { ZodTypeAny, z } from 'zod';
 import MockAdapter from 'axios-mock-adapter';
 
 export interface requestType {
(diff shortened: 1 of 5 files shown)


test/integrations/testTypes.ts
```diff
@@ -44,3 +45,98 @@ export type MockHttpCallsData = {
...
+      })
+      .optional(),
+    metadata: z.record(z.string(), z.any()).optional(),
+    statusCode: z.number().optional(),
+  }),
+]);
+
+export const batchTransformSchema = z.array(
+  z.object({
+    batchedRequest: arrayOrSingleElement(eventResponseSchema).optional(),
```
krishna2020, line 109: already `metadata` is captured in eventResponseSchema.output.metadata

test/integrations/testTypes.ts
```diff
@@ -44,3 +45,98 @@ export type MockHttpCallsData = {
   httpReq: Record<string, any>;
   httpRes: Partial<AxiosResponse>;
 };
+
+export const arrayOrSingleElement = (schema: ZodTypeAny) => z.union([z.array(schema), schema]);
+export const formElementSchema = arrayOrSingleElement(
+  z.union([z.string(), z.number(), z.boolean()]),
+);
+export const eventResponseSchema = z.union([
```
krishna2020, line 53: Need to enforce either success or error event related fields for each response - how will that be achieved if all are mentioned as optional ?

koladilip: currently, starting with only types validations, once this version is released, we can tighten it more

test/integrations/testTypes.ts
koladilip: that is mainly for process response

test/integrations/testTypes.ts
```diff
@@ -44,3 +45,98 @@ export type MockHttpCallsData = {
...
+  z.object({
+    error: z.string(),
+    statTags: z
+      .object({
+        destType: z.string().optional(),
+        errorCategory: z.string().optional(),
+        errorType: z.string().optional(),
+        feature: z.string().optional(),
```
sanpj2292, line 61: Can we equate this to be one of `FEATURES` in `src/v0/util/tags.js` that we already have ?

sanpj2292:
@koladilip ,

We need to have capability to selectively enable this schema validation for some destTypes that way we can make sure this is usable in a stricter way. As a separate task, we can migrate some of the most used destinations one-by-one.

**Extension of this feature**: I think we need to make sure there are stricter schema rules that too at destType level as only then can we make sure that we are sending adequate information via dest transformation.

test/integrations/testTypes.ts
koladilip: this is done

koladilip:
> @koladilip ,
> 
> We need to have capability to selectively enable this schema validation for some destTypes that way we can make sure this is usable in a stricter way. As a separate task, we can migrate some of the most used destinations one-by-one.
> 
> **Extension of this feature**: I think we need to make sure there are stricter schema rules that too at destType level as only then can we make sure that we are sending adequate information via dest transformation.

This is exactly we want to avoid, we don't want to any destination specific logic come here as this is agnostic to destinations, we are trying to make sure that we don't break rudder server.

krishna2020:
I see the test run github action being failed, are there any destinations which are not following rudder-server contract  ?

koladilip:
> I see the test run github action being failed, are there any destinations which are not following rudder-server contract ?

yes some 7-8 destinations are failing due to form handling. Some implementations support sending complex data using forms but it is recommended.
```json
  {
    x: '1',
    'arr[]': [ '1', '2', '3' ],
    'arr2[0]': '1',
    'arr2[1][0]': '2',
    'arr2[2]': '3',
    'arr3[]': [ '1', '2', '3' ],
    'users[0][name]': 'Peter',
    'users[0][surname]': 'griffin',
    'users[1][name]': 'Thomas',
    'users[1][surname]': 'Anderson'
  }
```
In this example we are encoding complex objects but I am not use all destination apis can handle such data. Current rudder-server doesn't support this but I found a go library supports this: https://github.com/go-playground/form. We need to research before integrating it as it will change current response structure. @krishna2020 I need help here

devops-github-rudderstack:

PR #2796, opened 2023-11-06
fix: add check to remove null and undefined properties before sending

yashasvibajpai:
## Description of the change

Resolves INT-980
Disallow null and undefined values from being passed to Adobe's XML payload

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/test/__tests__/data/adobe_analytics.json
+++ b/test/__tests__/data/adobe_analytics.json
@@ -1010,7 +1010,6 @@
         },
         "messageId": "1578564113557-af022c68-429e-4af4-b99b-2b9174056383",
         "properties": {
-          "order_id": "1234",
           "affiliation": "Apple Store",
           "value": 20,
           "revenue": 15.0,
@@ -1019,6 +1018,7 @@
           "discount": 1.5,
           "coupon": "ImagePro",
           "currency": "USD",
+          "purchaseId": "p101",
           "products": [
             {
               "product_id": "123",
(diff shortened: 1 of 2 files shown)


sanpj2292 (approved):

test/__tests__/data/adobe_analytics.json
```diff
@@ -1010,7 +1010,6 @@
         },
         "messageId": "1578564113557-af022c68-429e-4af4-b99b-2b9174056383",
         "properties": {
```
krishna2020, line 1012: what's this test case change for ?
There's no test case for the newly added change - where null or undefined ones should be empty in the output xml payload.

yashasvibajpai: Yes, I handled it in this change
<transaction_id> is mapped to either `properties.transactionId` or `properties.order_id` so it is not showing up in the final XML now as both these properties are not there, and it is not coming as undefined as it is being handled from the `removeUndefinedAndNullValues` function.

krishna2020: how come purchaseId was mapped to `1234` before ? what's the mapping rule of purchaseId

krishna2020 (approved):

test/__tests__/data/adobe_analytics.json
yashasvibajpai: `purchaseId`  is mapped to `properties.purchaseId` OR `properties.order_id` , since we had order_id in properties, hence the value 1234

krishna2020: why all these multiple fields mapping support ? Are we not following ecom spec here ? cc: @Gauravudia

yashasvibajpai: No @krishna2020 we are not, we have some destination specific customisations too like "override fields" in adobe

yashasvibajpai: Resolving this conversation for now for merging this PR. 🙇

PR #2856, opened 2023-11-29
fix: add support for custom properties for braze purchase events

utsabc:
## Description of the change

Adding support for `properties` fields for Purchase events to Braze [ref](https://www.braze.com/docs/api/objects_filters/purchase_object/)

Screenshot 2023-11-29 at 6 06 27 PM

The `properties` field is mapped with all the extra fields perent in each of the `product` elements present in `Order Completed` event.

This change introduces additional data to be sent into Braze

Refer to linear ticket INT-1076 for further details

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/src/v0/destinations/braze/util.js
+++ b/src/v0/destinations/braze/util.js
@@ -21,6 +21,7 @@ const {
   SUBSCRIPTION_BRAZE_MAX_REQ_COUNT,
   ALIAS_BRAZE_MAX_REQ_COUNT,
   TRACK_BRAZE_MAX_REQ_COUNT,
+  BRAZE_PURCHASE_STANDARD_PROPERTIES,
 } = require('./config');
 const { JSON_MIME_TYPE, HTTP_STATUS_CODES } = require('../../util/constant');
 const { isObject } = require('../../util');
@@ -539,7 +540,7 @@ function addMandatoryPurchaseProperties(productId, price, currencyCode, quantity
   };
 }
 
-function getPurchaseObjs(message, Config) {
+function getPurchaseObjs(message, config) {
   // ref:https://www.braze.com/docs/api/objects_filters/purchase_object/
   const validateForPurchaseEvent = (message) => {
     const { properties } = message;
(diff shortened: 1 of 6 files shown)


src/v0/destinations/braze/util.js
```diff
@@ -634,6 +634,10 @@ function getPurchaseObjs(message) {
       parseInt(quantity, 10),
       timestamp,
     );
+    const extraProperties = _.omit(product, ['product_id', 'sku', 'price', 'quantity', 'currency']);
```
koladilip, line 637: should be enable this via destination config to avoid impacting all customer, we don't know if everyone needs this or not

utsabc: Its not a good `product UX` that for a small feature like this we add a config in UI?

ItsSudip (dismissed):

src/v0/destinations/braze/util.js
utsabc: We are handling this behind a destination config

koladilip, line 637: if (Config.sendPurchaseEventWithExtraProperties) {
do omit logic then add to purchase object

doing omit without this config flag is useless right?
}

src/v0/destinations/braze/util.js
```diff
@@ -539,7 +539,7 @@ function addMandatoryPurchaseProperties(productId, price, currencyCode, quantity
   };
 }
 
-function getPurchaseObjs(message) {
+function getPurchaseObjs(message, Config) {
```
krishna2020, line 542:
```suggestion
function getPurchaseObjs(message, config) {
```

src/v0/destinations/braze/util.js
krishna2020, line 637: can we define this list at one place & use it for mappings & for omitting here ?
so that any new field mapping requirement, change at one place would be sufficient.
this can go as part of different PR too

utsabc: I am also checking the `extraProperties` length along with config hence I am extracting it outside

src/v0/destinations/braze/util.js
utsabc: Addressed

src/v0/destinations/braze/util.js
utsabc: Addressed

koladilip: but omitting operation itself is not need right if condition is not met?

ItsSudip (approved):

PR #2873, opened 2023-12-05
fix: salesforce transformer proxy response handling issue for authorization flow

shrouti1507:
## Description of the change

> Salesforce network handler was not sending `authorizationFlow` variable. We have added logic to set it. Also handled it specifically for salesforce ouath version, inside the `responseHandler` function

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/src/v0/destinations/salesforce/networkHandler.js
+++ b/src/v0/destinations/salesforce/networkHandler.js
@@ -1,14 +1,18 @@
 const { proxyRequest, prepareProxyRequest } = require('../../../adapters/network');
 const { processAxiosResponse } = require('../../../adapters/utils/networkUtils');
+const { LEGACY, OAUTH } = require('./config');
 const { salesforceResponseHandler } = require('./utils');
 
 const responseHandler = (destinationResponse, destType) => {
   const message = `Request for destination: ${destType} Processed Successfully`;
 
+  const authorizationFlow = destType === 'salesforce' ? LEGACY : OAUTH;
+
   salesforceResponseHandler(
     destinationResponse,
     'during Salesforce Response Handling',
     destinationResponse?.rudderJobMetadata?.destInfo?.authKey,
+    authorizationFlow
   );
 
   // else successfully return status as 200, message and original destination response
--- a/src/v0/destinations/salesforce/utils.js
+++ b/src/v0/destinations/salesforce/utils.js
@@ -33,7 +33,7 @@ const salesforceResponseHandler = (destResponse, sourceMessage, authKey, authori
     const matchErrorCode = (errorCode) =>
       response && Array.isArray(response) && response.some((resp) => resp?.errorCode === errorCode);
     if (status === 401 && authKey && matchErrorCode('INVALID_SESSION_ID')) {
-      if (authorizationFlow === OAUTH)  {
+      if (authorizationFlow === OAUTH) {
         throw new RetryableError(
           `${DESTINATION} Request Failed - due to "INVALID_SESSION_ID", (Retryable) ${sourceMessage}`,
           500,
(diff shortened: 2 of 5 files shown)


src/v0/destinations/salesforce/networkHandler.js
```diff
@@ -1,14 +1,18 @@
 const { proxyRequest, prepareProxyRequest } = require('../../../adapters/network');
 const { processAxiosResponse } = require('../../../adapters/utils/networkUtils');
+const { LEGACY, OAUTH } = require('./config');
 const { salesforceResponseHandler } = require('./utils');
 
 const responseHandler = (destinationResponse, destType) => {
   const message = `Request for destination: ${destType} Processed Successfully`;
 
+  const authorizationFlow = destType === 'salesforce' ? LEGACY : OAUTH;
```
utsabc, line 9: Can you explain this logic?

src/v0/destinations/salesforce/utils.js
```diff
@@ -33,16 +33,22 @@ const salesforceResponseHandler = (destResponse, sourceMessage, authKey, authori
...
       }
+      // checking for invalid/expired token errors and evicting cache in that case
+      // rudderJobMetadata contains some destination info which is being used to evict the cache
+      ACCESS_TOKEN_CACHE.del(authKey);
       throw new RetryableError(
         `${DESTINATION} Request Failed - due to "INVALID_SESSION_ID", (Retryable) ${sourceMessage}`,
         500,
         destResponse,
-        authorizationFlow === LEGACY ? '' : getAuthErrCategoryFromStCode(status),
+        '',
```
utsabc, line 51: No need to pass empty string for this. This field is only needed for Oauth

src/v0/destinations/salesforce/networkHandler.js
shrouti1507: when both salesforce and salesforce_oauth will be activated, destType will be either salesforce or salesforce_oauth. for salesforce we will use legacy way of authentication and vice versa

src/v0/destinations/salesforce/utils.js
shrouti1507: removed

src/v0/destinations/salesforce/networkHandler.js
krishna2020: can we have a consistent logic to infer LEGACY/OAUTH implementation ?
Here's it's different - https://github.com/rudderlabs/rudder-transformer/blob/9407b951f0d7a81400098492227c3bd3ad8c8644/src/v0/destinations/salesforce/utils.js#L165 ---> relying on destType is fool proof

sanpj2292: With OAuth - destType would be salesforce_oauth
Without OAuth - destType would be salesforce

`responseHandler` should be sent with `authorizationFlow` arg all the way from controller. We need to figure out which destination type supports which authorization flow.

I think this is the reason the mentioned logic has been added
Please confirm @shrouti1507

src/v0/destinations/salesforce/utils.js
```diff
@@ -33,16 +33,21 @@ const salesforceResponseHandler = (destResponse, sourceMessage, authKey, authori
     const matchErrorCode = (errorCode) =>
```
krishna2020, line 33: To make the code readable, the above `if` condition can be moved to calling class. There's no point of having only if without else in this function

krishna2020: this can be part of different PR

src/v0/destinations/salesforce/networkHandler.js
krishna2020: collectAuthorizationInfo logic also need to be driven via destType - can be different PR

krishna2020 (approved):

src/v0/destinations/salesforce_oauth/networkHandler.js
krishna2020, line 1: do we always get `destType` in lowercase from rudder-server as per the current contract ?

src/v0/destinations/salesforce/networkHandler.js
shrouti1507: added dedicated network handler as discussed

PR #2888, opened 2023-12-07
bug fix in email tracking

adityapatadia:


koladilip:
@adityapatadia thanks for this PR to fix the issue, can you please add description on how did you come up with this fix?

devops-github-rudderstack:

PR #2951, opened 2023-12-25
chore: add endpoint labels

yashasvibajpai:

Resolves INT-1244
We are adding endpoint labels to facilitate us in monitoring the latency of various api endpoints.

### Type of change

If the pull request is a **bug-fix**, **enhancement** or a **refactor**, please fill in the details on the changes made.

- Existing capabilities/behavior

- New capabilities/behavior

If the pull request is a **new feature**,

N/A

N/A

N/A

### If the PR has changes in more than 10 files, please mention why the changes were not split into multiple PRs.

N/A

### If multiple linear tasks are associated with the PR changes, please elaborate on the reason:

N/A

<hr>

### Developer checklist

- [ ] Are all related docs linked with the PR?

- [ ] Are all changes manually tested?

- [ ] Does this change require any documentation changes?

### Reviewer checklist


ItsSudip (approved):

src/v0/destinations/user/utils.js
```diff
@@ -277,6 +278,7 @@ const updateCompany = async (message, destination, company) => {
   const response = await httpPUT(endpoint, payload, requestOptions, {
     destType: 'user',
     feature: 'transformation',
+    endpointPath: `/companies/`,
```
sanpj2292, line 281: We should probably include `httpMethod` as well because in this case, `/companies` API is being used in POST & PUT http calls but as per stats we will only have a single endpointPath(`/companies/`) & not be able to differentiate between the POST / PUT calls.

sanpj2292: Else in `endpointPath` include PUT / POST as a prefix or suffix

sanpj2292: cc: @koladilip @yashasvibajpai

yashasvibajpai: Do you mean to add this specifically for user destination, or for all destinations across?

koladilip: @sanpj2292 does API limits differ based on PUT or POST method? if we decided to do then @yashasvibajpai better to do all destinations as it will be consistent

yashasvibajpai: @koladilip The destinations' implementations have exclusively one out of POST or PUT endpoints, so they don't have different limits for different request methods. 
on the structure, does this suffice?
for eg. like below
"[METHOD]/. . . /companies/"

koladilip: in that case I am fine either ways adding or not adding both are fine, if we add better to add as dedicated label rather than appending to endPoint label

sanpj2292: If limits is the use-case for which we are adding `endpointPath` label & if limits per http methods doesn't change, it is fine to proceed without adding another label or prefixing

yashasvibajpai: Yes, majorly those were the reasons. Not adding then

sanpj2292: Cool. Resolving the comment then

sanpj2292 (approved):

PR #2953, opened 2023-12-26
fix: for gainsight px only new users to have default signUp date

shrouti1507:

resolves INT-1179
We added the default signup date and create date for all the users ( existing / nonexisting). For any existing users, even if the message.traits object was not having signUpDate or createdDate properties it was being populated from message.timestamp / message.original timestamp field and every time the signUp and createdDate was being updated.

In this PR we are fixing the issues. We will only fallback to originalTimestamp / timestamp values if the user is new.

### Type of change

 **bug-fix**

- Existing capabilities/behavior

We added the default signup date and create date for all the users ( existing / nonexisting). For any existing users, even if the message.traits object was not having signUpDate or createdDate properties it was being populated from message.timestamp / message.original timestamp field and every time the signUp and createdDate was being updated.

- New capabilities/behavior
In this PR we are fixing the issues. We will only fallback to originalTimestamp / timestamp values if the user is new.

N/A

N/A

N/A

### If the PR has changes in more than 10 files, please mention why the changes were not split into multiple PRs.

N/A

### If multiple linear tasks are associated with the PR changes, please elaborate on the reason:

N/A

<hr>

### Developer checklist

- [ ] Are all related docs linked with the PR?

- [x] Are all changes manually tested?

- [ ] Does this change require any documentation changes?

### Reviewer checklist


yashasvibajpai (approved):

src/v0/destinations/gainsight_px/transform.js
```diff
@@ -56,6 +60,16 @@ const identifyResponseBuilder = async (message, { Config }) => {
     payload.firstName = fName;
     payload.lastName = lName;
   }
+  // Only for the case of new user creation, if signUpDate is not provided in traits, timestamp / originalTimestamp is mapped
+  if (!isPresent && !payload.signUpDate) {
```
koladilip, line 64: what is Present is it userID or traits? make it more clearer

Gauravudia:
LGTM

src/v0/destinations/gainsight_px/transform.js
shrouti1507: at line number 54:   
`const { success: isPresent } = await objectExists(userId, Config, 'user');`

using the userID we are checking if the user is already present. and keeping the status in isPresent

koladilip: can we rename as isUserPresent?

shrouti1507: addressed

koladilip (approved):

yashasvibajpai (approved):

PR #2968, opened 2024-01-05
feat: add structured logging

sanpj2292:

Adding structured logging to the service. We have used `winston` for this purpose.

Resolves INT-1087

N/A

N/A

N/A

N/A

<hr>

### Developer checklist

### Reviewer checklist

### DIFF
--- a/src/services/destination/nativeIntegration.ts
+++ b/src/services/destination/nativeIntegration.ts
@@ -26,6 +26,7 @@ import networkHandlerFactory from '../../adapters/networkHandlerFactory';
 import { FetchHandler } from '../../helpers/fetchHandlers';
 import tags from '../../v0/util/tags';
 import stats from '../../util/stats';
+import { MiscService } from '../misc';
 
 export class NativeIntegrationDestinationService implements DestinationService {
   public init() {}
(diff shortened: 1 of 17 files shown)


koladilip:
We need to have discussion on this, few questions:
1. Are we planning to send a child logger with all the context to destination or source handlers? If yes how does that work?
2. I thought from root level module we need keep passing a logger to child modules functions with more context, but there is no child logger being used here. Since the context changes per request we need to create child loggers per request that may be costly (not sure if that is really concern or not) else we need to do something golang does with context, we can create child context from parent and so on, we can keep one logger at module level but context keeps changing.

sanpj2292:
> Are we planning to send a child logger with all the context to destination or source handlers? If yes how does that work?

Most of use-case regarding logging surrounds around having a good logger that could tell us clearly about what's happening in the service. The logger requirement at destination/source level is quite less at this point. But when we consider the logger to be used at destination/source level currently this is the thought process.

> I thought from root level module we need keep passing a logger to child modules functions with more context, but there is no child logger being used here. Since the context changes per request we need to create child loggers per request that may be costly (not sure if that is really concern or not) else we need to do something golang does with context, we can create child context from parent and so on, we can keep one logger at module level but context keeps changing.

We have to look into the possibility of having a single parent logger with changing context.


As per our discussion offline, will try to implement that version & we can look at it in detail

src/services/destination/nativeIntegration.ts
```diff
@@ -113,8 +121,12 @@ export class NativeIntegrationDestinationService implements DestinationService {
           tags.FEATURES.ROUTER,
         );
         try {
+          const metadataWithSvcCtx = {
+            ...requestMetadata,
+            serviceContext: MiscService.getLoggableData(metaTO.errorDetails),
```
koladilip, line 126: this is more of request context right, service context might be misleading?

src/services/destination/nativeIntegration.ts
```diff
@@ -63,23 +64,30 @@ export class NativeIntegrationDestinationService implements DestinationService {
...
     const respList: ProcessorTransformationResponse[][] = await Promise.all(
       events.map(async (event) => {
+        const metaTO = this.getTags(
+          destinationType,
+          event.metadata?.destinationId,
+          event.metadata?.workspaceId,
+          tags.FEATURES.PROCESSOR,
+        );
+        metaTO.metadata = event.metadata;
+        const metadataWithSvcCtx = {
```
koladilip, line 74: I thought we can simply add to requestMetadata like following:
requestMetadata.context = {...};

src/services/destination/nativeIntegration.ts
sanpj2292: The thought process behind naming was to reflect the context at service level.

src/services/destination/nativeIntegration.ts
sanpj2292: We are getting an eslint error when we do as suggested hence adopted this approach

src/services/destination/nativeIntegration.ts
koladilip: may be check with others and choose the one with less confusing

src/services/destination/nativeIntegration.ts
```diff
@@ -63,23 +64,30 @@ export class NativeIntegrationDestinationService implements DestinationService {
...
+          ...requestMetadata,
+          serviceContext: MiscService.getLoggableData(metaTO.errorDetails),
```
koladilip, line 76: why only from error details? we should use 
```
destinationType,
event.metadata?.destinationId,
event.metadata?.workspaceId,
tags.FEATURES.PROCESSOR,
```

sanpj2292: We are storing the information like destinationType, destinationId etc,. at service layer level in this type. We are just re-using the same type.

devops-github-rudderstack:

PR #2994, opened 2024-01-16
fix: klaviyo: Event validation to be string before operating on it

anantjain45823:
If message.event for track call is sent as non-string then it will result in transformation issue that we do not handle as we operate on it assuming it would be string

Resolves INT-1353

Putting a check for event to be string before doing operations on it

N/A

N/A

N/A

N/A

<hr>

### Developer checklist

### Reviewer checklist

### DIFF
--- a/test/integrations/destinations/klaviyo/processor/data.ts
+++ b/test/integrations/destinations/klaviyo/processor/data.ts
@@ -1493,6 +1493,115 @@ export const data = [
 	},
 	{
 		"name": "klaviyo",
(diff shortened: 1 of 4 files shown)


test/integrations/destinations/klaviyo/processor/data.ts
```diff
@@ -1491,6 +1491,115 @@ export const data = [
 			}
 		}
 	},
+	{
```
shrouti1507, line 1494: @anantjain45823 please format the file

shrouti1507 (approved):

PR #3015, opened 2024-01-22
Updated license to ELv2

arnab-p:
Updated license to ELv2


krishna2020:
@arnab-p rudder-transformer is already on ELv2 license - https://github.com/rudderlabs/rudder-transformer/blob/main/LICENSE
`master` branch is no longer being used. `main` is the active branch

PR #3104, opened 2024-02-19
fix: version deprecation failure false positive

sanpj2292:

We are setting whole error object in destinationResponse which is triggering version deprecation alert(false-positive)

Resolves INT-1617

N/A

N/A

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist


devops-github-rudderstack:
Test report for this run is available at: https://test-integrations-dev.s3.amazonaws.com/integrations-test-reports/rudder-transformer/3104/test-report.html

test/integrations/destinations/gainsight_px/network.ts
```diff
@@ -1,222 +1,514 @@
...
-                    "street": "",
-                    "postalCode": "",
-                    "continent": "",
-                    "regionName": "",
-                    "timeZone": "",
-                    "coordinates": {
-                        "latitude": 0.0,
-                        "longitude": 0.0
-                    }
+        httpsAgent: {
```
koladilip, line 278: do we need to keep the whole thing?

sanpj2292: It was only included to suggest that we got an AxiosError that was very long.

koladilip: Got it but do we need to test any code functionality?

sanpj2292: Yes, in a way. So we saw in our analysis that we had a keyword `upgrade required`, which is why the score of version deprecation increased leading to false-positive alerts.
In the bytes below, I think we would have some information about the keyword I was talking about hence put the information explicitly here

Through the test-case we are making sure that we don't put such information into payload when there is a failure from destination(that leads to `AxiosError`)

test/integrations/destinations/gainsight_px/network.ts
```diff
@@ -1,222 +1,514 @@
...
-            method: 'GET',
+  },
+  // Axios Error
+  {
+    httpReq: {
+      url: 'https://api.aptrinsic.com/v1/users/myUId',
+      headers: { 'X-APTRINSIC-API-KEY': 'sample-api-key', 'Content-Type': 'application/json' },
+      method: 'GET',
+    },
+    httpRes: {
```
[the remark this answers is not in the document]
sanpj2292: As requested, I have trimmed down the response to a more readable one. Can you check and confirm, if the changes are looking good ?

koladilip (approved):

PR #3124, opened 2024-02-25
feat: add record event support for redis destination

yashasvibajpai:

Add record event support for redis destination

Resolves INT-978

N/A

N/A

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist


devops-github-rudderstack:
Test report for this run is available at: https://test-integrations-dev.s3.amazonaws.com/integrations-test-reports/rudder-transformer/3124/test-report.html

src/v0/destinations/redis/transformV2.js
```diff
@@ -0,0 +1,45 @@
...
+//     "fields": {
+//        MODEL_ID: '1691755780',
+//        VALID_AT: '2023-08-11T11:32:44.963062Z',
+//        USER_MAIN_ID: 'rid5530313526204a95efe71d98cd17d5a1',
+//        CHURN_SCORE_7_DAYS: 0.027986,
+//        PERCENTILE_CHURN_SCORE_7_DAYS: 0,
+//     },
+// }
+
+function handleRecordEventsForRedis(message, destination, metadata) {
```
saikumarrs, line 35: There is no need to add "Redis" in the name. It's part of the Redis directory, so it's understood.

src/v0/destinations/redis/transform.js
```diff
@@ -35,7 +36,10 @@ const transformSubEventTypeProfiles = (message, workspaceId, destinationId) => {
   // form the hash
   const hash = `${workspaceId}:${destinationId}:${message.context.sources.profiles_entity}:${message.context.sources.profiles_id_type}:${message.userId}`;
   const key = `${message.context.sources.profiles_model}`;
-  const value = JSON.stringify(message.traits);
+  let value = JSON.stringify(message.traits);
```
saikumarrs, line 39: We should not JSON stringify traits by default if it is possible to not use them.

src/v0/destinations/redis/transformV2.js
```diff
@@ -0,0 +1,45 @@
+const { transformSubEventTypeProfiles } = require('./transform');
```
saikumarrs, line 1: Do we really need a "v2" file? It doesn't seem like a convention.

saikumarrs (changes requested):

src/v0/destinations/redis/transformV2.js
yashasvibajpai: Fixed it to a utils.js file name

src/v0/destinations/redis/transform.js
yashasvibajpai: It was being stringified before also, should I remove the stringification?

saikumarrs: The implementation should have looked something like this,

```
if event type is record {
  // do record specific logic
} else {
  // stringify traits
}
```

src/v0/destinations/redis/transformV2.js
saikumarrs: Maybe you haven't pushed the changes to upstream.

koladilip (commented):
I am thinking of adding action to all cases and base the rudder-server logic on that

yashasvibajpai:
> I am thinking of adding action to all cases and base the rudder-server logic on that 
@koladilip 
But if we mandate action in all events by default, in case of non-profiles standard identify events, older implementation of using HMSET to replace the whole hash would become redundant

src/v0/destinations/redis/transform.js
yashasvibajpai: Right now the flow is
event -
- if event is a record type event
- - handle insert/delete action related logic

- if event is having profiles' contextual data
- -  it is a profiles-identify event (backward compatible, only insertion supported)

- for none of the above
-- it is a standard redis destination's identify event

koladilip:
> > I am thinking of adding action to all cases and base the rudder-server logic on that
> > @koladilip
> > But if we mandate action in all events by default, in case of non-profiles standard identify events, older implementation of using HMSET to replace the whole hash would become redundant

Lets discuss over all with @saikumarrs

src/v0/destinations/redis/transform.js
saikumarrs: I don't think you're getting it.

I meant that you should not stringify traits by default. Instead, do it only for non-record type of event flows.

devops-github-rudderstack:

PR #3167, opened 2024-03-05
feat: onboard destination movable ink

Gauravudia:

Onboard new destination: Movable Ink

- Minimal transformation is required.
- The entire payload is passed to the Movable Ink destination.
- Movable Ink platform has a predefined schema event.
- Users need to configure the properties path in the Movable Ink platform and ensure values are passed in the same path in RudderStack.
- Movable will discard additional properties at their end.
- Added batch support. There is no hard batch limit set by Movable Ink. In the transformer, adding a rough batch limit from load test.

Resolves INT-1737

https://github.com/rudderlabs/rudder-integrations-config/pull/1252

N/A

N/A

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist


devops-github-rudderstack:
Test report for this run is available at: https://test-integrations-dev.s3.amazonaws.com/integrations-test-reports/rudder-transformer/3167/test-report.html

src/cdk/v2/destinations/movable_ink/procWorkflow.yaml
```diff
@@ -0,0 +1,71 @@
...
+      const userId = .message.().(
+          {{{{$.getGenericPaths("userIdOnly")}}}};
+      );
+      const email = .message.().(
+          {{{{$.getGenericPaths("email")}}}};
+      ); 
+      const timestampInUnix = $.toUnixTimestampInMS(.message.().(
+            {{{{$.getGenericPaths("timestamp")}}}};
+      ));
+      $.context.payload = {
```
shrouti1507, line 48: As I gather from research spec, movable ink supports most of the rudderstack standard ecomm events. And I see that we have documented the by default mapping in the research spec as well. But in implementation, I do not see any mapping.json file. Which means we are actually allowing whatever fields coming as input, straight as output event. What is the thought behind this? 
Is it just because that the user needs to configure every single field inside the movable, that is why we are giving such liberty ?
But in the spec, I see we have multiple required field. Where is the check for handling that if missing?

Gauravudia: In movable ink we have to configure the each event schema, So, we don't have a way to know if user is configured the schema in movable ink or not. Thats why we are passing the event as it is.
It is more like a documented heavy destination than the implementation wise.

shrouti1507: Which means, even if some fields are stated as mandatory, if we do not send, it will not throw error?

Gauravudia: Yes

shrouti1507 (dismissed):

shrouti1507 (approved):

src/cdk/v2/destinations/movable_ink/procWorkflow.yaml
krishna2020, line 48: Is transformation of payload handled at Movable end ? from rudderstack convention to movable convention ?

Gauravudia: - User has to configure payload schema in movable ink dashboard for each event. Added a comment for this step.
- Movable will discard any additional fields that are not part of schema.

Gauravudia: Info related to this is already mentioned in PR description

shrouti1507 (approved):

yashasvibajpai (approved):

utsabc (approved):

PR #3433, opened 2024-05-31
feat(integrations/auth0): add Auth0 event type to event context

zenpaul:

The Auth0 event type was added to the transformed message at the top level, but that field gets stripped out when messages are exported to warehouse destinations (see [this documentation](https://www.rudderstack.com/docs/destinations/warehouse-destinations/faq/#why-am-i-not-able-to-see-the-properties-added-at-the-top-level-of-an-event-in-warehouse-destination)). Add this field to the context blob as well, so that it will be available when exported to warehouses.

N/A

No.

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist

### DIFF
--- a/src/v0/sources/auth0/mapping.json
+++ b/src/v0/sources/auth0/mapping.json
@@ -65,6 +65,6 @@
   },
   {
     "sourceKeys": "type",
-    "destKeys": "source_type"
+    "destKeys": ["source_type", "context.source_type"]
   }
 ]
--- a/src/v0/sources/auth0/mapping.json
+++ b/src/v0/sources/auth0/mapping.json
@@ -65,6 +65,6 @@
   },
   {
     "sourceKeys": "type",
-    "destKeys": ["source_type", "properties.source_type"]
+    "destKeys": "properties.source_type"
   }
 ]
--- a/src/v0/sources/auth0/mapping.json
+++ b/src/v0/sources/auth0/mapping.json
@@ -65,6 +65,6 @@
   },
   {
     "sourceKeys": "type",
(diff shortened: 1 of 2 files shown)


src/v0/sources/auth0/mapping.json
```diff
@@ -65,6 +65,6 @@
   },
   {
     "sourceKeys": "type",
-    "destKeys": "source_type"
+    "destKeys": ["source_type", "context.source_type"]
```
Gauravudia, line 68:
```suggestion
    "destKeys": ["source_type", "properties.source_type"]
```

Gauravudia: Better to add this under properties, since we are already mapping other fields from Auth0 to RudderStack in same object.

Gauravudia (approved):

src/v0/sources/auth0/mapping.json
```diff
@@ -65,6 +65,6 @@
   },
   {
     "sourceKeys": "type",
-    "destKeys": "source_type"
+    "destKeys": ["source_type", "properties.source_type"]
```
Gauravudia, line 68:
```suggestion
    "destKeys": "properties.source_type"
```

Gauravudia: We can remove the root level source_type, Since it will be anyway dropped before exported to warehouse destinations
cc- @krishna2020

Gauravudia (approved):

utsabc (approved):

PR #3582, opened 2024-07-22
feat: use event name table map

lvrach:
### Motivation
A prospect customer stumble upon an incompatibility between segment and us, on mapping event names to table names.

We map `omega v2` as `omega_v_2`, while segment does `omega_v2`. Although, the second mapping makes more sense, we can not change the behaviour in our system for backwards compatibility reasons. 

We want to introduce a new mechanism, to workaround this problem for the customer, but also address similar issues when a custom mapping is need.

Ideally, customers should be able to do this for control-plane during destination configuration. 

### Solution

This PR provides a way to quickly support custom table name mappings, using an environment variables.

`WAREHOUSE_EVENT_NAME_TABLE_MAP={"<destinationID>": {"omega v2": "omega_v2"}}`

Until, we develop a way to support this from control-plane, destination settings.

### Considerations

Existing ways for passing configuration:

1. Pass them as integration options on user transformations, like we have done for disabling user and track tables. -> It will pollute user transformation, since we are not sure how we want to handle product wise. 
2. Pass it as query param from rudder-server. Will require changes in two systems, passing a mapping as query param could be more complex and error prone. We could consider passing it in the JSON payload, but it would pollute rudder-server even further.

Resolves INT-XXX

### DIFF
--- a/src/warehouse/v1/util.js
+++ b/src/warehouse/v1/util.js
@@ -4,10 +4,18 @@ const reservedANSIKeywordsMap = require('../config/ReservedKeywords.json');
 const { isDataLakeProvider } = require('../config/helpers');
 const { TransformationError } = require('@rudderstack/integrations-lib');
 
+let eventNameTableMap = JSON.parse(process.env.WAREHOUSE_EVENT_NAME_TABLE_MAP ?? '{}');
+
 function safeTableName(options, name = '') {
   const { provider } = options;
   const skipReservedKeywordsEscaping =
     options.integrationOptions?.skipReservedKeywordsEscaping || false;
+
+  let customTableName = eventNameTableMap[options.destinationId]?.[name];
+  if (customTableName) {
+    return customTableName;
+  }
+
   let tableName = name;
   if (tableName === '') {
     throw new TransformationError('Table name cannot be empty.');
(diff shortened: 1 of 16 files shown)


devops-github-rudderstack:
Test report for this run is available at: https://test-integrations-dev.s3.amazonaws.com/integrations-test-reports/rudder-transformer/3582/test-report.html

src/warehouse/v1/util.js
```diff
@@ -4,10 +4,18 @@ const reservedANSIKeywordsMap = require('../config/ReservedKeywords.json');
 const { isDataLakeProvider } = require('../config/helpers');
 const { TransformationError } = require('@rudderstack/integrations-lib');
 
+let eventNameTableMap = JSON.parse(process.env.WAREHOUSE_EVENT_NAME_TABLE_MAP ?? '{}');
+
 function safeTableName(options, name = '') {
   const { provider } = options;
```
koladilip, line 10: `destinationId` can be defined here

koladilip (dismissed):

devops-github-rudderstack:

devops-github-rudderstack:

devops-github-rudderstack:

PR #3728, opened 2024-09-12
fix: handle circular structure response payload

Jayachand:

Write a brief explainer on your code changes.

Resolves DAT-1571

N/A

N/A

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist


devops-github-rudderstack:
Test report for this run is available at: https://test-integrations-dev.s3.amazonaws.com/integrations-test-reports/rudder-transformer/3728/test-report.html

src/controllers/userTransform.ts
```diff
@@ -39,12 +39,19 @@ export class UserTransformController {
...
     const processedRespone: UserTransformationServiceResponse =
       await UserTransformService.transformRoutine(events, ctx.state.features, requestSize);
-    ctx.body = processedRespone.transformedEvents;
-    ControllerUtility.postProcess(ctx, processedRespone.retryStatus);
-    logger.debug(
-      '(User transform - router:/customTransform ):: Response from transformer',
-      ctx.response.body,
-    );
+    try {
+      ctx.body = JSON.stringify(processedRespone.transformedEvents);
```
koladilip, line 43: earlier you are setting JSON object and now you are setting string so won't that have an issue?

koladilip: Looks like it is not efficient to detect cycles using JSON stringify so can we write a utility like this:
```js
function hasCycle(obj) {
  const seenObjects = new Set();

  function detectCycle(obj) {
    if (obj && typeof obj === 'object') {
      if (seenObjects.has(obj)) {
        // Cycle detected
        return true;
      }

      // Add the current object to the set of seen objects
      seenObjects.add(obj);

      // Recursively check all keys of the object
      for (const key in obj) {
        if (obj.hasOwnProperty(key) && detectCycle(obj[key])) {
          return true;
        }
      }

      // Remove the object from the set after the recursion finishes
      seenObjects.delete(obj);
    }

    return false;
  }

  return detectCycle(obj);
}

// Example usage:
const a = {};
const b = { a };
a.b = b; // Creates a cycle

console.log(hasCycle(a)); // true
```

abhimanyubabbar: In 99% of the cases we won't have a cycle, so JSON.stringify would create the resultant string and we would return back with content-type application json. 

In 1% or maybe less we could see a cycle and then it would be better to use your solution. Given the occurrence is very less, I would argue that we only use stringify and catch the errors and return.

Let me know what you think ? @koladilip

koladilip: sure

koladilip (approved):

abhimanyubabbar (approved):

devops-github-rudderstack:

PR #3879, opened 2024-11-15
fix: adding logger for undefined source event

shrouti1507:

We have got bugsnag error :
Screenshot 2024-11-15 at 12 05 03 PM

Adding logger to get more info

Resolves INT-2873

N/A

N/A

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist

### DIFF
--- a/src/controllers/source.ts
+++ b/src/controllers/source.ts
@@ -4,13 +4,27 @@ import { MiscService } from '../services/misc';
 import { SourcePostTransformationService } from '../services/source/postTransformation';
 import { ControllerUtility } from './util';
 import logger from '../logger';
+import { FixMe } from '../util/types';
 
 export class SourceController {
+  private static enrichMetadata(
+    metadata: { namespace: string; cluster: string; features: FixMe },
+    source: string,
+    version: string,
+  ) {
+    return {
+      ...metadata,
+      source,
+      version,
+    };
+  }
+
   public static async sourceTransform(ctx: Context) {
     logger.debug('Native(Source-Transform):: Request to transformer::', ctx.request.body);
     const requestMetadata = MiscService.getRequestMetadata(ctx);
     const events = ctx.request.body as object[];
     const { version, source }: { version: string; source: string } = ctx.params;
+    const enrichedMetadata = this.enrichMetadata(requestMetadata, source, version);
     const integrationService = ServiceSelector.getNativeSourceService();
 
     try {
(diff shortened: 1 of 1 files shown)


devops-github-rudderstack:
Test report for this run is available at: https://test-integrations-dev.s3.amazonaws.com/integrations-test-reports/rudder-transformer/3879/test-report.html

sanpj2292 (approved):

src/controllers/source.ts
```diff
@@ -28,6 +28,11 @@ export class SourceController {
       );
       ctx.body = resplist;
     } catch (err: any) {
```
krishna2020, line 30: No need to mention module again in error message - we have structured logging now. Focus on mentioning the error message.
`events` can't be logged, it's a security risk - how did you overlooked it ? Is there specific reason to log events ?
cc: @sanpj2292

sanpj2292: > how did you overlooked it ? Is there specific reason to log events ?

The inclusion of `events` was a call we took to understand what is the event which is causing a problem. The reason is we don't have a lot of context on what kind of event was received to us & even though we get the error message, we would not be in a position to properly identify the problem and come up with a fix.

In this particular bugsnag issue, we saw that `sourceEvents` came in as a non-array which means there is some problem in adaption logic. And is it undefined, string, object -- we absolutely have no-clue

We will remove `events` from getting logged but we should definitely find an alternate (if not already in consideration or present) to address this

sanpj2292: @shrouti1507 let's remove `events` from the log getting emitted for an error

src/controllers/source.ts
```diff
@@ -4,13 +4,27 @@ import { MiscService } from '../services/misc';
...
+      version,
+    };
+  }
+
   public static async sourceTransform(ctx: Context) {
     logger.debug('Native(Source-Transform):: Request to transformer::', ctx.request.body);
     const requestMetadata = MiscService.getRequestMetadata(ctx);
     const events = ctx.request.body as object[];
     const { version, source }: { version: string; source: string } = ctx.params;
+    const enrichedMetadata = this.enrichMetadata(requestMetadata, source, version);
```
krishna2020, line 27: do we really need a function here ?
```suggestion
    const enrichedMetadata = {
      ...metadata,
      source,
      version,
    };
```

sanpj2292: Done

krishna2020 (approved):

src/controllers/source.ts
krishna2020, line 1: If required, sourceId can also be logged from `events[0].config`

Issue #3892, opened 2024-11-19
MoEngage Destination | Anon ID is overwriting User ID


jmlandi:
https://github.com/rudderlabs/rudder-transformer/blob/79e597907eee126b4187e4534b2aa2253d1431da/src/v0/destinations/moengage/transform.js#L52

Hello, everyone! 

My name is Marcos Landi, I'm using RudderStack at CloudWalk and found an issue connecting to MoEngage. When we have anonymousId and userId at the same event, the anonymousId overwrite the userId.

Following other destinations' behavior, analytics value and the fact that an Anon ID is temporary and a User ID permanent, it is better for us (and possibly all other RudderStack users) to prioritize the User ID over Anon ID.

If you guys need a quick call or evidence from our account, please let me know!

gitcommitshow:
Thanks for reporting the bug @jmlandi , will fix it. Or if you want, feel free to raise a PR fixing this.

jmlandi:
@gitcommitshow thanks! I just submitted a PR to the develop branch #3914, but my CLA is still pending. Let me know if there's anything else you need or if I made any mistakes. I'm eager to contribute more to the project, so any feedback on this first PR would be greatly appreciated!

gitcommitshow:
Thank you for the contribution.  LGTM. Do sign the CLA, I will get the PR reviewed asap.

devops-github-rudderstack:
This issue is considered to be stale. It has been open for 30 days with no further activity thus it is going to be closed in 7 days. To avoid such a case please consider removing the stale label manually or add a comment to the issue.

PR #3914, opened 2024-12-05
fix: anonymousId overwriting userId at MoEngage destination

jmlandi:

Changed the preference for `userId` over `anonymousId` for MoEngage destination.

No related Linear task for this change.

### Any changes to existing capabilities/behavior? Mention the reason & what changes were made.

This change adjusts MoEngage destination behavior to prefer `userId` over `anonymousId` when both are provided. 
Currently, when both `userId` and `anonymousId` are present, the anonymous ID overwrites the user ID, leading to unexpected behavior, such as users receiving an `anonymousId` from a web session that overrides the main user identification.

For more details, see the opened issue #3892.

No.

### Any new generic utility introduced or modified? Please explain the changes.

No.

### Any technical or performance-related considerations with this change?

No.

@coderabbitai review

---

### Developer checklist

- [X] My code follows the style guidelines of this project.
- [X] **No breaking changes are being introduced.**
- [X] All related docs linked with the PR.
- [X] All changes manually tested.
- [ ] Any documentation changes needed with this change.
- [X] Is the PR limited to 10 file changes?
- [ ] Is the PR limited to one Linear task?
- [ ] Are relevant unit and component test cases added in **new readability format**?

### Reviewer checklist


krishna2020 (approved):

sanpj2292 (approved):

gitcommitshow:
@jmlandi without this change, did you see any issue in the event you receive at the destination.
Can you please share the transformed event sample you see (you may use the RudderStack live debugger for MoEngage destination to get it)?

The reason I am asking for this is because engineering team tested the current MoEngage destination (without this PR change) and it worked as expected.

This is what I learned from the engineering team

> The variable you changed acts only as a placeholder for the data transformer we send back to server which has not real value, this property is not even sent to destination.
> The actual payload is the payload variable you need to follow this
https://github.com/rudderlabs/rudder-transformer/blob/ba85d230f2fb684606f026b3e6dda875f8ce4672/src/v0/destinations/moengage/transform.js#L27
> And the order by which the customer_id is filled is this order https://github.com/rudderlabs/rudder-transformer/blob/ba85d230f2fb684606f026b3e6dda875f8ce4672/src/v0/destinations/moengage/data/MOENGAGEIdentifyConfig.json#L3
> You can use the identify API to identify a user in MoEngage. You can also create a new user property or update existing user properties using the identify call.
> MoEngage needs a unique identifier to identify a user. So, you must provide a userId in your identify call that RudderStack passes as the MoEngage customer_id . If userId is absent, RudderStack sends the anonymousId field instead.
> So our implementation is in accordance to our docs.
> So can you please confirm if the destination is not working as documented? If not then we have to debug and find it which is another point.

jmlandi:
@gitcommitshow, thank you for your response. I understand that the` customer_id` is the identifier referred to in MoEngage's [Create Event Documentation](https://developers.moengage.com/hc/en-us/articles/4413174104852-Create-Event). However, we are still encountering cases where users are being registered with the anonymousId instead of the userId. I am also reaching out to MoEngage for further investigation on their side.

That said, the userId in the message seems to be overwritten, which I believe is not the expected behavior—even if it doesn't fully solve my issue. Could you confirm if this is correct? Here is evidence from a recent event.

In this image, you may see the `correct customer_id`, as engineering claimed, but a `wrong userId`.
image

gitcommitshow:
Thanks for sharing the details @jmlandi.

> That said, the userId in the message seems to be overwritten, which I believe is not the expected behavior—even if it doesn't fully solve my issue. Could you confirm if this is correct? 

Yes, that is correct. So we will go ahead with this PR as a fix for this different issue

And we will separately investigate that "users being registered with the anonymousId" issue, the CS team has already taken up that issue.

gitcommitshow:
@jmlandi you may delete the attached image in your comment now

koladilip (approved):

PR #3925, opened 2024-12-09
chore(release): pull release/v1.86.0 into main

devops-github-rudderstack:
:crown: *An automated PR*


achettyiitr (dismissed):

src/v0/destinations/braze/util.js
```diff
@@ -490,6 +540,7 @@ const processBatch = (transformedEvents) => {
   prepareGroupAndAliasBatch(mergeUsersArrayChunks, responseArray, destination, 'merge');
 
   if (successMetadata.length > 0) {
+    console.log(`Response 1 batchRequest ${JSON.stringify(responseArray)}`);
```
krishna2020, line 543: @aashishmalik remove the log here

src/v0/destinations/google_adwords_offline_conversions/transform.js
```diff
@@ -48,6 +49,9 @@ const getConversions = (message, metadata, { Config }, event, conversionType) =>
       filteredCustomerId,
       eventLevelConsentsData,
     );
+    convertedPayload.payload.conversions[0] = updateConversion(
```
krishna2020, line 52: Are no checks required here on non-empty conversions here ? And why handling only one conversion ? @ItsSudip

src/v0/destinations/braze/util.js
aashishmalik: removed

krishna2020 (approved):

utsabc (approved):

src/v0/destinations/google_adwords_offline_conversions/transform.js
ItsSudip: We don't need to check that. The way config json is written will always create one conversion.

krishna2020: then why do we have it as an array ? legacy code ?

ItsSudip: We need to send this in array format only. Conversions is an array of conversion objects.

krishna2020: Then handling in as part of payload preparation itself would have sufficed - https://github.com/rudderlabs/rudder-transformer/blob/develop/src/v0/destinations/google_adwords_offline_conversions/utils.js#L385

Now the transformation logic is split in 2 places. Is updateConversion handles if Array.isArray(products) is false ?

ItsSudip: No that is not required. That is already being handled before the `updateConversion` function call.

utsabc (approved):

ItsSudip (dismissed):

shrouti1507 (dismissed):

krishna2020 (approved):

utsabc (approved):

PR #4025, opened 2025-01-28
fix: add phone e164 format validation for klaviyo

yashasvibajpai:

We have a popular widely used library [libphonenumber-js](https://www.npmjs.com/package/libphonenumber-js) here to validate the phone number being passed to Klaviyo, as Klaviyo rejects events with phone number not in the [E.164](https://en.wikipedia.org/wiki/E.164) standard. Klaviyo [Reference](https://developers.klaviyo.com/en/v2024-06-15/reference/events_api_overview#receiving-a-400).
As a precaution, the phone number is first sanitized ( i.e remove spaces, dashes etc ) and then the library is used on it.

Resolves INT-3180

The events which were having the phone numbers in incorrect format and would earlier hit the klaviyo APIs and then get rejected, will now fail at the transformation layer, thereby saving the API calls on the failed events.

N/A

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "koa": "^2.15.3",
     "koa-bodyparser": "^4.4.0",
     "koa2-swagger-ui": "^5.7.0",
-    "libphonenumber-js": "^1.11.12",
+    "libphonenumber-js": "^1.11.18",
     "lodash": "^4.17.21",
     "match-json": "^1.3.5",
     "md5": "^2.3.0",
--- a/src/v0/destinations/klaviyo/util.js
+++ b/src/v0/destinations/klaviyo/util.js
@@ -1,5 +1,6 @@
 const set = require('set-value');
 const lodash = require('lodash');
+const { parsePhoneNumberFromString } = require('libphonenumber-js');
 const { NetworkError, InstrumentationError } = require('@rudderstack/integrations-lib');
 const { WhiteListedTraits } = require('../../../constants');
 const {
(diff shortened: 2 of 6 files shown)


devops-github-rudderstack:
- Allure Report: https://rudderlabs.github.io/rudder-transformer/202/index.html

package.json
```diff
@@ -95,7 +95,7 @@
     "koa": "^2.15.3",
     "koa-bodyparser": "^4.4.0",
     "koa2-swagger-ui": "^5.7.0",
-    "libphonenumber-js": "^1.11.12",
+    "libphonenumber-js": "^1.11.18",
```
koladilip, line 98: we are already using validator library and why do we need another library

koladilip: https://www.npmjs.com/package/validator

koladilip: we have another use case for this so it is fine to use this.

src/v0/destinations/klaviyo/util.js
```diff
@@ -32,6 +33,24 @@ const {
 
 const REVISION_CONSTANT = '2023-02-22';
 
+/**
+ * This function is used to check if the phone number is in E.164 format. It uses libphonenumber-js library to parse the phone number
+ * @param {*} phoneNumber
+ * @returns boolean
+ */
+function isValidE164PhoneNumber(phoneNumber) {
```
koladilip, line 41: validator.isMobilePhone is taking care of hyphens so better we use it.

koladilip: as discussed validator is missing few cases.

koladilip (approved):

PR #4038, opened 2025-01-31
chore: add validation fix for zoho

utsabc:

This PR adds validation for fields which are used for searching in Zoho. Also refactored test cases to updated format

Resolves INT-3206


test/integrations/destinations/zoho/router/data.ts
```diff
@@ -1,4 +1,3167 @@
-import { upsertData } from './upsert';
-import { deleteData } from './deletion';
+/**
```
utsabc, line 1: Review these test cases in allure rather than here

devops-github-rudderstack:
- Allure Report: https://rudderlabs.github.io/rudder-transformer/240/index.html

devops-github-rudderstack:
- Allure Report: https://rudderlabs.github.io/rudder-transformer/241/index.html

src/cdk/v2/destinations/zoho/utils.js
krishna2020, line 1: @utsabc I don't think just filtering the undefined/null values is the right fix here. In the example payload of customer, record with `Client_Do_Not_Sell_Status` field as null should be retrieved. Instead, this code will filter it out --> meaning other records with all possible values of `Client_Do_Not_Sell_Status` field will be picked up for deletion.
Right fix will be to convert the null/undefined values to right format of zoho query

krishna2020 (changes requested):
this change needs to be re-think on how `=null` condition need to be passed for zoho instead of skipping the field

koladilip:
> Allure Test reports for this run are available at:
> 
>     * Allure Report: https://rudderlabs.github.io/rudder-transformer/240/index.html

It is showing 404

devops-github-rudderstack:
- Allure Report: https://rudderlabs.github.io/rudder-transformer/300/index.html

PR #4220, opened 2025-03-31
feat: support compacted destination transformation payloads

atzoum:

Transformer now advertises support for a new feature, through which it supports **compacted payloads** for the following endpoints:

```
POST /:version/destinations/:destination
POST /routerTransform
POST /batch
```

Clients wanting to take advantage of this feature, e.g. rudder-server, need to use a special HTTP request header `X-Content-Format: json+compactedv1` along with the compacted payload in the request's body.

Compacted payloads, omit the `destination` and `connection` fields from each individual event and include a lookup table instead, e.g.
```json
{
    "input": [
        {
            "message": {...},
            "metadata": {
                "sourceId": "source-1",
                "destinationId": "destination-1"
            }
        },
        {
            "message": {...},
            "metadata": {
                "sourceId": "source-1",
                "destinationId": "destination-1"
            }
        },{
            "message": {...},
            "metadata": {
                "sourceId": "source-1",
                "destinationId": "destination-1"
            }
        },{
            "message": {...},
            "metadata": {
                "sourceId": "source-1",
                "destinationId": "destination-2"
            }
        }
    ],
    "destinations": {
        "destination-1": {...},
        "destination-2": {...}
    },
    "connections": {
        "source-1:destination-1": {...},
        "source-1:destination-2": {...}
    }
}
```
This compaction can provide significant benefits, taking into consideration that destination objects are quite big and heavy, usually bigger than the messages themselves:
- Faster marshalling in rudder-server, requiring less cpu (rudder-server now spends ~50% of its cpu time for marshalling/unmarshalling json payloads)
- Faster network transport times
- Faster unmarshalling at rudder-transformer, requiring less cpu

Resolves PIPE-1992


devops-github-rudderstack:
- Allure Report: [View Report](https://rudderlabs.github.io/test-reports/rudder-transformer/4220/index.html)

devops-github-rudderstack:
- Allure Report: [View Report](https://rudderlabs.github.io/test-reports/rudder-transformer/4220/index.html)

devops-github-rudderstack:
- Allure Report: [View Report](https://rudderlabs.github.io/test-reports/rudder-transformer/4220/index.html)

src/middlewares/destTransformCompactedPayloadV1.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Middleware to transform a compacted payload for destination transformations
+ * to its decompacted form, as it is expected by the destination transformers.
+ * Compacted payloads are used to reduce the size of the payload by removing
+ * connection and destination details from each input and instead providing them
+ * in separate lookup tables.
+ */
+export class DestTransformCompactedPayloadV1Middleware {
+  public static async handle(ctx: Context, next: Next): Promise<void> {
+    if (ctx.get('x-content-format') === 'json+compactedv1') {
```
utsabc, line 13: Are we planning to add metrics on usage of this structure in rudder-server side?

utsabc (dismissed):

koladilip (dismissed):
Looks good to me

src/types/destinationTransformation.ts
```diff
@@ -98,6 +104,13 @@ export type RouterTransformationRequest<M = RudderMessage, MD = Metadata> = {
   destType: string;
 };
 
+export type RouterCompactedTransformationRequest<M = RudderMessage, MD = Metadata> = {
+  input: Omit<RouterTransformationRequestData<M, MD>, 'destination' | 'connection'>[];
```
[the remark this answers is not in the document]
vinayteki95: This is a valid error. Needs to be fixed.

src/middlewares/__tests__/destTransformCompactedPayloadV1.test.ts
```diff
@@ -0,0 +1,112 @@
...
+        } as ProcessorCompactedTransformationRequest,
+      },
+    } as Koa.Context;
+    ctx.get = jest.fn((v) => (v === 'x-content-format' ? 'json+compactedv1' : ''));
+
+    DestTransformCompactedPayloadV1Middleware.handle(ctx, async () => {
+      expect(ctx.request.body).toEqual(expectedBody);
+    });
+  });
+});
```
vinayteki95, line 112: We can add test cases for mismatch of destination or connection config id.
Ideal case is to throw an error with clear logs.

src/middlewares/routerTransformCompactedPayloadV1.ts
```diff
@@ -0,0 +1,27 @@
...
+      ctx.request.body = {
+        input: body.input.map((input) => ({
+          ...input,
+          connection:
+            body.connections[`${input.metadata.sourceId}:${input.metadata.destinationId}`],
+          destination: body.destinations[input.metadata.destinationId],
+        })),
+        destType: body.destType,
+      } as RouterTransformationRequest;
+    }
```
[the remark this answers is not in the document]
vinayteki95: Let's not throw transformationErrors as we've configured them to be not alerted by bugsnag. Let's throw a generic Error.

src/middlewares/__tests__/destTransformCompactedPayloadV1.test.ts
atzoum: Are there any validations now on the normal flow for `connection` or `destination` objects, e.g. that they are properly filled?

devops-github-rudderstack:
- Allure Report: [View Report](https://rudderlabs.github.io/test-reports/rudder-transformer/4220/index.html)

src/middlewares/__tests__/destTransformCompactedPayloadV1.test.ts
atzoum: imo, these would be most beneficial, because they would handle both scenarios, of compacted and non-compacted payloads.

src/middlewares/routerTransformCompactedPayloadV1.ts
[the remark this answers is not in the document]
atzoum: What are we trying to solve here? the code generated by coderrabit doesn't look that great...
- Schema validations, if that's what is worrying us should be applied in all endpoints using some common infrastructure, e.g. json schema and not selectively on middlewares and handlers
- The same would stand for other kinds of business validations.

Does rudder-transformer use some framework for schema/business validations already? If not we should prioritize that, instead of adding validations in random places. If yes, I would be happy to use that.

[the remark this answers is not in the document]
vinayteki95: Currently we do not have a central validation framework. Ideally yes, we should have a central framework.
We can ignore the suggestion from coderabbitai. Don't think it is doing a good job here.

The expectation is, if a particular event has destinationId "dest_id" and we do not have that available in the destination map we are sending along with the request, it should throw an error that corresponding destination config for the id "dest_id" is missing.

This will clearly state that there is some core bug at the server end. Because our destination integration logic doesn't check if config is available or not explicitly.

src/middlewares/__tests__/destTransformCompactedPayloadV1.test.ts
atzoum: Assuming that a combination is missing, the result would be an empty destination or connection field.
My point is, do we protect ourselves now against empty destination or connection fields? If not, why should we do it here at this middleware?

ps. integration with rudder server will be verified through rudder-server's integration tests with rudder-transformer, i.e. this feature will be enabled while running all rudder-server tests

devops-github-rudderstack:
- Allure Report: [View Report](https://rudderlabs.github.io/test-reports/rudder-transformer/4220/index.html)

devops-github-rudderstack:
- Allure Report: [View Report](https://rudderlabs.github.io/test-reports/rudder-transformer/4220/index.html)

src/middlewares/__tests__/destTransformCompactedPayloadV1.test.ts
krishna2020: If validations aren't in place, any observability measures can we implement to ensure that the lookup never returns empty?

For example, in the case of Braze, if the destination configuration is missing the [dataCenter](https://github.com/rudderlabs/rudder-transformer/blob/35cda1ed1b49eee6548a55d04d0711dff83055a6/src/v0/destinations/braze/transform.js#L210) field, instrumentation errors will be thrown. If the lookup returns an empty destConfig, the same errors will be thrown. How can we differentiate between customer misinstrumentation and a potential bug in the middleware?

atzoum: Rudder server's code preparing these payloads will be quite straightforward and we are going to test all affected endpoints through rudder-server tests.
IMO validation at the middleware layer is not needed, but if you believe that validation about keys missing is required, I shall add it

atzoum: Are we sure that we are not going to introduce some regression though? E.g. do ALL transformations require a destination/connection object and WILL FAIL if it is missing?
I still believe that these kind of validations at this layer are not a good idea and could cause more issues than they would try to solve.

devops-github-rudderstack:
- Allure Report: [View Report](https://rudderlabs.github.io/test-reports/rudder-transformer/4220/index.html)

src/middlewares/__tests__/destTransformCompactedPayloadV1.test.ts
koladilip: If that happens, this is an edge case and platform bug, and we should throw a 500 error from the transformer. In most cases we easily cover this using unit tests and some integration tests. we don't need to put validations, anyways if that happens it will fail while processing so we don't need to explicitly put validations during event processing as it is costly

koladilip: @atzoum as we start this new process now, initially to be on the safe side let's go with validations and throw specific errors (PlatformError) so that we don't accidentally end up aborting events when it is a platform issue. we can't control 200+ integrations' behavior, they might behave differently as over time different people developed them and there is no consistency in them. Once we are confident about this we can remove validations for efficiency.

atzoum: I added some validation with respect to connections and destinations as requested.
`@koladilip` I am using a `PlatformError` and explicitly setting the error code to `500` because my guess was that the default one is `400`. Am I missing something important? Are we using some special middleware which performs special handling for PlatformErrors? What is the error code we should return here? Codebase seems a bit confusing to me...

devops-github-rudderstack:
- Allure Report: [View Report](https://rudderlabs.github.io/test-reports/rudder-transformer/4220/index.html)

src/middlewares/__tests__/destTransformCompactedPayloadV1.test.ts
krishna2020: PlatformError with `500` should be right one here so that events get retried. But to know this behaviour, [platform-errors](https://github.com/rudderlabs/rudderstack-operator/blob/master/operator-helm/alerts/cross-customer-pipelines/integrations/platform-errors.yaml) alert should be triggered - @koladilip should the statTags be explicitly added here as the alert definition depends on `errorCategory` key ?

koladilip (approved):

src/middlewares/__tests__/destTransformCompactedPayloadV1.test.ts
koladilip: https://github.com/rudderlabs/rudderstack-operator/blob/f98fbb703c3522c397a9bd18c5b3e0f95a1a6245/operator-helm/alerts/cross-customer-pipelines/integrations/platform-errors.yaml#L3

https://github.com/rudderlabs/rudder-integrations-lib/blob/fe534c8349f7e85d7a27dd2ca8e0369987e7bf61/src/errors/platform_error.ts#L4

errorCategory=platform is automatically added

devops-github-rudderstack:
- Allure Report: [View Report](https://rudderlabs.github.io/test-reports/rudder-transformer/4220/index.html)

krishna2020 (approved):

vinayteki95 (approved):

PR #4311, opened 2025-05-10
chore: add string check for phone in gaoc

yashasvibajpai:

This pull request includes a minor improvement to the `getAddConversionPayload` function in the `src/v0/destinations/google_adwords_offline_conversions/utils.js` file. The change enhances the validation logic for the `phone` field before applying the `sha256` hash.

Validation improvement:

* Updated the condition for hashing the `phone` field to include a check for whether it is a string (`isString(phone)`), ensuring that non-string values are not processed incorrectly.

Resolves INT-3641

N/A

N/A

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist

### DIFF
--- a/src/v0/destinations/google_adwords_offline_conversions/utils.js
+++ b/src/v0/destinations/google_adwords_offline_conversions/utils.js
@@ -272,7 +272,9 @@ const getAddConversionPayload = (message, Config, eventLevelConsentsData) => {
         ? sha256(email.trim()).toString()
         : email,
     phone:
-      hashUserIdentifier && isDefinedAndNotNull(phone) ? sha256(phone.trim()).toString() : phone,
+      hashUserIdentifier && isString(phone) && isDefinedAndNotNull(phone)
+        ? sha256(phone.trim()).toString()
+        : phone,
     address: buildAndGetAddress(message, hashUserIdentifier),
   };
 
--- a/src/v0/destinations/google_adwords_offline_conversions/utils.js
+++ b/src/v0/destinations/google_adwords_offline_conversions/utils.js
@@ -272,7 +272,7 @@ const getAddConversionPayload = (message, Config, eventLevelConsentsData) => {
         ? sha256(email.trim()).toString()
         : email,
     phone:
-      hashUserIdentifier && isString(phone) && isDefinedAndNotNull(phone)
+      hashUserIdentifier && isDefinedAndNotNull(phone) && isString(phone)
         ? sha256(phone.trim()).toString()
         : phone,
     address: buildAndGetAddress(message, hashUserIdentifier),
--- a/src/v0/destinations/google_adwords_offline_conversions/utils.js
+++ b/src/v0/destinations/google_adwords_offline_conversions/utils.js
@@ -461,4 +461,5 @@ module.exports = {
   getConsentsDataFromIntegrationObj,
   getCallConversionPayload,
   updateConversion,
+  getAddConversionPayload,
 };
--- a/src/v0/destinations/google_adwords_offline_conversions/utils.test.js
+++ b/src/v0/destinations/google_adwords_offline_conversions/utils.test.js
@@ -1,9 +1,11 @@
+const sha256 = require('sha256');
 const {
   getClickConversionPayloadAndEndpoint,
   buildAndGetAddress,
   getExisitingUserIdentifier,
   getConsentsDataFromIntegrationObj,
   getCallConversionPayload,
+  getAddConversionPayload,
 } = require('./utils');
 
 const API_VERSION = 'v18';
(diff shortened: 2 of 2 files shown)


devops-github-rudderstack:
- Allure Report: [View Report](https://rudderlabs.github.io/test-reports/rudder-transformer/4311/index.html)

src/v0/destinations/google_adwords_offline_conversions/utils.js
```diff
@@ -272,7 +272,9 @@ const getAddConversionPayload = (message, Config, eventLevelConsentsData) => {
         ? sha256(email.trim()).toString()
         : email,
     phone:
-      hashUserIdentifier && isDefinedAndNotNull(phone) ? sha256(phone.trim()).toString() : phone,
+      hashUserIdentifier && isString(phone) && isDefinedAndNotNull(phone)
```
manish339k, line 275: what will be the API response if we send phone as number or object? If it throws error, can we remove this field?

yashasvibajpai: I think passing one out of email or phone is mandatory in click/store conversions. Also silently removing event data should not be done ideally as this can cause confusion among customers on why a particular field is not going through. Hence, i am adding validation for it to be a string, as I see in the docs, this should be a string is not mentioned, we can update docs.

Thoughts @ItsSudip ?

manish339k: We should be careful while adding any extra validation.

vinayteki95: What if both email and phone number do not satisfy these conditions? If the destination expects atleast one of them to be available shouldn't we validate that and throw an error?

We should verify if either phone or email is necessary, and if that is the case, we should validate that as well.

yashasvibajpai: Yes, it is true that either of email/phone is mandatory for click/store conversions.
We have a validation in [L 333](https://github.com/rudderlabs/rudder-transformer/blob/4c6bcc4c62cd760eea0109b4253502a6d46e3815/src/v0/destinations/google_adwords_offline_conversions/utils.js#L335) already

vinayteki95: I think the validation is happening on the message directly and not before the clean up we are doing here. Can you verify if that is the case?

Fundamentally what we want to verify here is, whether we are validating if either of email or phone are  available after we conditionally remove invalid email and phone.

yashasvibajpai: Since we are hashing data to send to google, we need it in string format,
If the customer does not pass email/phone, the API will complain as it is needed and the error should show up in Errors, rather than now when there call is going as success and we are receiving alerts.
CC @ItsSudip

src/v0/destinations/google_adwords_offline_conversions/utils.js
```diff
@@ -272,7 +272,9 @@ const getAddConversionPayload = (message, Config, eventLevelConsentsData) => {
...
+        ? sha256(phone.trim()).toString()
+        : phone,
```
vinayteki95, line 277: Can we add test cases as well for the changes we made?

devops-github-rudderstack:
- Allure Report: [View Report](https://rudderlabs.github.io/test-reports/rudder-transformer/4311/index.html)

src/v0/destinations/google_adwords_offline_conversions/utils.js
yashasvibajpai: added

src/v0/destinations/google_adwords_offline_conversions/utils.js
ItsSudip, line 275:
```suggestion
      hashUserIdentifier && isDefinedAndNotNull(phone) && isString(phone)
```
We should check the existence of the phone, then check if that is a string or not.

devops-github-rudderstack:
- Allure Report: [View Report](https://rudderlabs.github.io/test-reports/rudder-transformer/4311/index.html)

ItsSudip (approved):

src/v0/destinations/google_adwords_offline_conversions/utils.test.js
```diff
@@ -394,3 +379,174 @@ describe('getCallConversionPayload', () => {
...
+    const config = {
+      validateOnly: false,
+      hashUserIdentifier: true,
+      defaultUserIdentifier: 'email',
+    };
+
+    const result = getAddConversionPayload(message, config, {});
+
+    // Should have empty userIdentifiers
+    expect(result.operations.create.userIdentifiers[0]).toEqual({
```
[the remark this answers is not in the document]
yashasvibajpai: Yes, it should be fine to send this to google

vinayteki95 (approved):

PR #4455, opened 2025-07-04
fix: process memory usage method high cpu utilization

atzoum:
## Description

Updating memory usage every 100ms to avoid increased cpu utilization

### DIFF
--- a/src/middlewares/memoryFencing.ts
+++ b/src/middlewares/memoryFencing.ts
@@ -11,6 +11,7 @@ import stats from '../util/stats';
  *   Defaults to 503 (Service Unavailable).
  */
 interface MemoryFenceOptions {
+  memoryUsageRefreshPeriod?: number; // default 100ms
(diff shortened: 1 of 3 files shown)


src/middlewares/memoryFencing.ts
```diff
@@ -29,17 +30,21 @@ interface MemoryFenceOptions {
...
   }
   const limit = v8.getHeapStatistics().heap_size_limit;
   stats.gauge('memory_heap_size_limit', limit);
-
+  let { heapUsed }: { heapUsed: number } = process.memoryUsage();
+  let lastMemoryCheck = Date.now();
   return async (ctx, next) => {
-    const { heapUsed } = process.memoryUsage();
+    // Check memory usage periodically
+    if (Date.now() - lastMemoryCheck >= memoryUsageRefreshPeriod) {
```
koladilip, line 43: can we use setInterval for this to keep updating heapUsed, they we don't see to really check the condition everytime.

PR #4491, opened 2025-07-13
fix: add mapping of advertiser_tracking_enabled and enable extInfo with empty string

ItsSudip:

Fix the logic of populating extInfo, setting an empty string for null values, apart from dropping the entire extInfo and use advertiser_tracking_enabled.

Resolves INT-3910

N/A

N/A

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist

### DIFF
--- a/src/v0/destinations/snapchat_conversion/utilsV3.js
+++ b/src/v0/destinations/snapchat_conversion/utilsV3.js
@@ -128,8 +128,7 @@ const getExtInfo = (message) => {
     deviceInfo.freeStorage,
     environmentInfo.timezone,
   ];
-
-  return extInfo.some((value) => value == null) ? null : extInfo;
+  return extInfo.map((value) => (value == null ? '' : value));
 };
 
 module.exports = {
--- a/src/v0/destinations/snapchat_conversion/transformV3.js
+++ b/src/v0/destinations/snapchat_conversion/transformV3.js
@@ -220,6 +220,12 @@ const trackResponseBuilder = (message, { Config }, mappedEvent) => {
       ...commonProperties.custom_data,
     };
   }
+  if (commonProperties.app_data) {
+    payload.data[0].app_data = {
+      ...payload.data[0]?.app_data,
+      ...commonProperties.app_data,
+    };
+  }
 
   // Process and validate payload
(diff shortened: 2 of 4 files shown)


devops-github-rudderstack:
- Allure Report: [View Report](https://rudderlabs.github.io/test-reports/rudder-transformer/4491/index.html)

src/v0/destinations/snapchat_conversion/utilsV3.js
```diff
@@ -128,8 +128,7 @@ const getExtInfo = (message) => {
     deviceInfo.freeStorage,
     environmentInfo.timezone,
   ];
-
-  return extInfo.some((value) => value == null) ? null : extInfo;
+  return extInfo.map((value) => (value == null ? '' : value));
```
krishna2020, line 131: I think the expectation will be to not send a value when it's not present.
So you also need to use filter for dropping the fields that are null. There's difference between the field being empty vs null vs 'null' vs non-existent field

yashasvibajpai (approved):

src/v0/destinations/snapchat_conversion/utilsV3.js
ItsSudip: https://developers.snap.com/api/marketing-api/Conversions-API/Parameters As per the doc, I don't see sending an empty string, but the customer suggested sending empty strings.

ItsSudip:
> I don't see any usage of `advertiser_tracking_enabled` though PR title says it

Actually, it is there in the mapping, but the populated value is not being used. So I added the use of it through commonProperties.app_data. I have updated one test case regarding this.

src/v0/destinations/snapchat_conversion/utilsV3.js
krishna2020: Extinfo requires all values and needs them to be in the order indexed below in the extinfo object specs.
It's assigning specific field [based on index](https://developers.snap.com/api/marketing-api/Conversions-API/Parameters#extinfo) - that's the reason for customer asking to send empty values to maintain the order. Then this change looks good

PR #4496, opened 2025-07-14
chore: remove logger for cdk service

utsabc:

This pull request includes a small change to the `getErrorInfo` function in `src/cdk/v2/utils.ts`. The change removes a logging statement that previously logged workflow step execution errors.

### DIFF
--- a/src/cdk/v2/utils.ts
+++ b/src/cdk/v2/utils.ts
@@ -86,8 +86,6 @@ export function getErrorInfo(err: CatchErr, isProd: boolean, defTags) {
   const message = isProd ? getWorkflowEngineErrorMessage(err) : err.message;
 
   if (err instanceof WorkflowExecutionError) {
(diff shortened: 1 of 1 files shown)


src/cdk/v2/utils.ts
```diff
@@ -86,8 +86,6 @@ export function getErrorInfo(err: CatchErr, isProd: boolean, defTags) {
   const message = isProd ? getWorkflowEngineErrorMessage(err) : err.message;
 
   if (err instanceof WorkflowExecutionError) {
-    logger.error(`Error occurred during workflow step execution: ${message}`, err);
```
koladilip, line 89: move this to debug

koladilip (changes requested):

src/cdk/v2/utils.ts
utsabc: done

PR #4579, opened 2025-08-21
fix: replace --json flag with grep for gh pr create compatibility

koladilip:

Fixed GitHub CLI compatibility issue in the prepare-for-prod-ut-deploy.yml workflow by replacing the --json flag with grep for extracting PR URLs.

N/A - Bug fix for workflow compatibility

The --json flag is not available for write operations like gh pr create in older versions of GitHub CLI, causing workflow failures. This fix ensures compatibility across different GitHub CLI versions.

**Change**: Replaced gh pr create --fill --json url --jq '.url' with gh pr create --fill | grep -o 'https://github.com/[^/]*/[^/]*/pull/[0-9]*'

**Reason**: The --json flag is only available for read operations (like gh release view) but not for write operations (like gh pr create) in older GitHub CLI versions.

N/A

N/A

The new approach uses grep with regex to extract PR URLs, which is more compatible but slightly less robust than JSON parsing. However, it works reliably for GitHub PR URLs.

@coderabbitai review

### DIFF
--- a/.github/workflows/prepare-for-prod-ut-deploy.yml
+++ b/.github/workflows/prepare-for-prod-ut-deploy.yml
@@ -135,7 +135,7 @@ jobs:
           git commit -m "chore: upgrade shared user-transformers to $UT_TAG_NAME"
           git push -u origin shared-user-transformer-$UT_TAG_NAME
 
-          PR_URL=$(gh pr create --fill --json url --jq '.url')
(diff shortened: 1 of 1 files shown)


devops-github-rudderstack:
- Allure Report: [View Report](https://rudderlabs.github.io/test-reports/rudder-transformer/4579/index.html)

PR #4712, opened 2025-10-14
chore(release): pull hotfix-release/v1.111.0 into main

devops-github-rudderstack:
:crown: *An automated PR*


src/v1/destinations/am/networkHandler.ts
```diff
@@ -0,0 +1,107 @@
...
+    );
+  }
+  if (isHttpStatusThrottled(status)) {
+    const { throttled_users: throttledUsers, throttled_devices: throttledDevices } = response;
+
+    const hasThrottledItems = (obj: Record<string, number>) =>
+      typeof obj === 'object' && obj !== null && Object.keys(obj).length > 0;
+
+    if (hasThrottledItems(throttledUsers) || hasThrottledItems(throttledDevices)) {
+      throw new RetryableError(
```
ItsSudip: Add a comment to explain why we need to return 500 instead of 429. The reason is that user-level throttling shouldn't affect the whole destination.

PR #4958, opened 2026-02-04
chore: version upgrade google ad destinations from 19 to 22

achettyiitr:

- Upgrade Google Ads integrations to use API v22 while documenting v23 as the latest available version, and align Google Ads version audit metadata accordingly.

- Resolves INT-5594

### DIFF
--- a/.github/scripts/versions.json
+++ b/.github/scripts/versions.json
@@ -112,7 +112,7 @@
     "Sunset Date Current Version used in backend": "Not Available"
   },
   {
-    "Destination": "GAEC",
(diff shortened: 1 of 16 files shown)


.github/scripts/versions.json
```diff
@@ -113,24 +113,24 @@
   },
   {
     "Destination": "GAEC",
```
ItsSudip: Can we rename it to GOOGLE_ADWORDS_ENHANCED_CONVERSIONS

ItsSudip (dismissed):

PR #5030, opened 2026-03-09
chore: batch framework

ItsSudip:

Write a brief explainer on your code changes.

Resolves INT-XXX

N/A

N/A

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist


krishna2020:
### Code review

Found 4 issues:

1. **Race condition: `BrazeIntegration` singleton has mutable instance state reset across concurrent async calls**

`userStore` and `failedLookupIdentifiers` are instance fields on the module-level singleton (`export const integration = new BrazeIntegration()`). `batchTransform()` resets them at lines 86–87 and then suspends at `await BrazeDedupUtility.doLookup()`. A second concurrent request hitting the same singleton will reset these fields mid-flight, corrupting the first request's dedup state. The legacy `processRouterDest` deliberately used function-local variables to avoid this.

https://github.com/rudderlabs/rudder-transformer/blob/212cda3b04c9f10db789c30bca7d214464294185/src/v0/destinations/braze/routerTransform.ts#L74-L96

2. **Debug `maxChunkSize: 2` left in PostHog production code — comment explicitly says production value is 250**

This will send one HTTP request per every 2 events to PostHog instead of per 250, causing a ~125x increase in API call volume and likely hitting rate limits.

https://github.com/rudderlabs/rudder-transformer/blob/212cda3b04c9f10db789c30bca7d214464294185/src/v0/destinations/posthog/routerTransform.ts#L195-L199

3. **Validation errors are miscounted as successes in metrics**

`processBatchedDestination` returns a mixed array of success and error responses, all passed to `handleRouterTransformSuccessEvents`. That function only increments `event_transform_failure` if `statTags` is present and non-empty. Validation failures from `validate()` (lines 344–351) have no `statTags`, so they fall through and incorrectly increment `event_transform_success`.

https://github.com/rudderlabs/rudder-transformer/blob/212cda3b04c9f10db789c30bca7d214464294185/src/services/destination/routerIntegration.ts#L343-L352

4. **`reqMetadata` hardcoded as `{}` in Braze — disables the `isNewStatusCodesAccepted` feature flag**

`batchTransform()` calls `process(input, processParams, {})` with a hardcoded empty object. The actual `reqMetadata` (carrying `features.FEATURE_FILTER_CODE`) is available in `doRouterTransformation` but never passed down. This causes `isNewStatusCodesAccepted(reqMetadata)` to always return `false`, silently breaking the 298 vs 400 filtered-events behavior for all Braze traffic on the new framework path.

https://github.com/rudderlabs/rudder-transformer/blob/212cda3b04c9f10db789c30bca7d214464294185/src/v0/destinations/braze/routerTransform.ts#L112-L116

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

ItsSudip:
Thanks for the thorough review! Addressing each point:

**1. Race condition — singleton mutable instance state**

Fixed by switching from a module-level singleton to a per-call instantiation pattern. Instead of exporting an instance (`export const integration = new BrazeIntegration()`), each destination now exports the class itself (`export const Integration = BrazeIntegration`). The framework caches the constructor in `FetchHandler` and calls `new IntegrationClass()` inside `processBatchedDestination` for each request. This gives each request a fully isolated instance, so `userStore` and `failedLookupIdentifiers` are never shared across concurrent calls while still being available as class-level fields across `batchTransform`, `postTransform`, and any future methods.

**2. Debug `maxChunkSize: 2`**

This is intentionally hardcoded to `2` for testing purposes only to make it easy to verify chunking behavior in tests. The production value will be `250` before this PR is merged.

**3. Validation errors miscounted as successes in metrics**

Fixed by adding `statTags` to validation error responses in `validate()`:
```ts
statTags: {
  errorCategory: tags.ERROR_CATEGORIES.DATA_VALIDATION,
  errorType: tags.ERROR_TYPES.INSTRUMENTATION,
}
```
`handleRouterTransformSuccessEvents` checks `'error' in resp && isObject(resp.statTags) && !isEmpty(resp.statTags)` to route to `event_transform_failure`. With `statTags` present, validation failures now correctly increment `event_transform_failure`.

**4. `reqMetadata` hardcoded as `{}`**

Fixed by threading `reqMetadata` through the full call chain: `doRouterTransformation` → `processBatchedDestination` → `batchTransform` → `process()`. Added as an optional parameter at each level so existing integrations (e.g. PostHog) are unaffected. Braze now passes `reqMetadata ?? {}` to `process()`, restoring the `isNewStatusCodesAccepted` behavior.

PR #5060, opened 2026-03-23
fix(mp): engage/groups API errors silently treated as success

shekhar-rudder:
## Summary
- Without `verbose=2` query param, MixPanel's `/engage` and `groups` endpoint return a non-standard response format that the transformer incorrectly parses as success (HTTP 200)
- Errors like missing `$distinct_id` were silently swallowed instead of propagating the actual 400 error from MixPanel
- Adding `verbose=2` makes MixPanel return the proper error response, e.g. `{"error": "Engage API error: $distinct_id, missing or empty", "status": 0}`

## Before / After
**Before** (error silently treated as success):
```json
{ "statusCode": 200, "error": "success" }
```
**After** (error correctly propagated):
```json
{ "statusCode": 400, "error": "Engage API error: $distinct_id, missing or empty" }
```

## Test plan
- [ ] Updated processor test expectations to include `params: { verbose: 2 }` (50 cases)
- [ ] Updated router test expectations to include `params: { verbose: 2 }` (8 cases)
- [ ] Run `jest mp` integration tests locally

### DIFF
--- a/src/v0/destinations/mp/transform.js
+++ b/src/v0/destinations/mp/transform.js
@@ -110,6 +110,7 @@ const responseBuilderSimple = (payload, message, eventType, destConfig) => {
       response.headers = {
         'Content-Type': 'application/json',
       };
+      response.params = { verbose: 2 };
   }
   return response;
 };
--- a/test/integrations/destinations/mp/processor/data.ts
+++ b/test/integrations/destinations/mp/processor/data.ts
@@ -690,7 +690,7 @@ export const data = [
               headers: {
                 'Content-Type': 'application/json',
               },
-              params: {},
+              params: { verbose: 2 },
               body: {
                 JSON: {},
                 JSON_ARRAY: {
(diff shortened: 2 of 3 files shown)


src/v0/destinations/mp/transform.js
```diff
@@ -110,6 +110,7 @@ const responseBuilderSimple = (payload, message, eventType, destConfig) => {
       response.headers = {
         'Content-Type': 'application/json',
       };
+      response.params = { verbose: 2 };
```
ItsSudip, line 113: It should be either 0 or 1 as per [this](https://developer.mixpanel.com/reference/profile-set) doc.

shekhar-rudder: From my local testing
`verbose=0`
`0`

`verbose=1`
`{
  "error": "$distinct_id, missing or empty",
  "status": 0
}`

`verbose=2`
Adds 2 extra fields in verbose=1 - `errors` and `num_good_events`

ItsSudip (approved):

src/v0/destinations/mp/transform.js
maheshkutty: If things are not documented. In the future, it may be possible that they remove support for it

PR #5066, opened 2026-03-23
fix: validate IP address in geolocation handler to prevent path traversal

itsmihir:

The geolocation function exposed to user transformations interpolates user input directly into the URL path without validation. An attacker could pass path traversal strings (e.g. '../../metrics') to reach arbitrary endpoints on the internal geolocation service.
Add ipaddr.isValid(ip) validation in all three geolocation handler copies (customTransformer.js, ivmFactory.js, contextReset.js) to reject any input that is not a valid IPv4/IPv6 address before the fetch call.

Resolves PIPE-2841

N/A

N/A

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist


test/__tests__/ivmCache/contextReset.test.js
```diff
@@ -536,6 +536,23 @@ describe('Context Reset Utilities', () => {
...
+      const mockReject = {
+        applyIgnored: jest.fn(),
+      };
+
+      await geolocationCallback(mockResolve, mockReject, '../../metrics');
+
+      expect(mockFetch).not.toHaveBeenCalled();
+      expect(mockReject.applyIgnored).toHaveBeenCalledWith(
+        undefined,
+        [expect.any(Object)] // ExternalCopy error
```
fracasula: Should we add an assertion on the error message being `"invalid ip address"` as well?

src/util/customTransformer.js
```diff
@@ -102,6 +103,9 @@ async function runUserTransform(
         if (args.length < 1) {
           throw new Error('ip address is required');
         }
+        if (!net.isIP(args[0])) {
```
fracasula: [optional] I don't like that we keep spreading logic/validation for this. For example we check the args length, if it is an IP and then we throw an exception if the geolocation url env var is empty.
What do you think about moving these into a common function? It could be an easy change.

src/util/ivmCache/contextReset.js
```diff
@@ -111,6 +112,9 @@ async function injectFreshApis(jail, cachedIsolate, credentials) {
         if (args.length === 0) {
           throw new Error('ip address is required');
         }
+        if (!net.isIP(args[0])) {
```
fracasula: This might fail if there is a port, right? Like `net.isIP('192.168.1.1:8080')`.
If that's the case it might be worth adding support for it? WDYT?

ktgowtham (dismissed):

itsmihir:
@fracasula @ktgowtham Using ipaddr to check if IP is valid or not to be consistent with geolocation service.

src/util/ivmCache/contextReset.js
itsmihir: Using `ipaddr` now to be consistent with geolocation

fracasula (dismissed):

src/util/geolocation.js
```diff
@@ -0,0 +1,17 @@
+const ipaddr = require('ipaddr.js');
+
+/**
+ * Validates an IP address argument for use in the geolocation API path.
+ * Uses ipaddr.js for consistency with the geolocation service.
+ * Throws if the IP is missing or invalid.
+ */
+function validateGeolocationIp(ip) {
```
ktgowtham: couldn't this have been a generic util instead of being geo location specific?

ktgowtham (dismissed):

ktgowtham (approved):

maheshkutty (approved):

PR #5074, opened 2026-03-24
chore: hashing pii batching and validations

achettyiitr:

- **API version** `202509` → **`202603`** (aligned with CDK workflow `LinkedIn-Version` header).
- **User identifiers**
  - Allow only configured trait keys; reject unknown identifier keys with a clear error.
  - **`validateHashingConsistency`** so plaintext vs pre-hashed values match `isHashRequired`.
  - When hashing is required: normalize email-like traits (strip whitespace, lowercase), validate with `validator.isEmail`, optionally drop invalid values when reject-invalid-fields is enabled (`LINKEDIN_AUDIENCE_REJECT_INVALID_FIELDS`); then **SHA-256** / **SHA-512** as appropriate.
  - **`InstrumentationError`** if no valid identifiers remain after processing.
- **Company identifiers**
  - Validate keys against the company trait allowlist.
  - Require at least one non-empty identifier after cleanup.
- **Batching**: split each action/group into chunks of **`MAX_BATCH_SIZE`** (default 5000, override via **`LINKEDIN_AUDIENCE_MAX_BATCH_SIZE`**).
- **`utils.ts`**: remove `hashIdentifiers` / `prepareUserIds`; keep endpoint + `prepareNonNullRecord` helpers.

- Resolves INT-5845, INT-5844, INT-5840, INT-5847, INT-5601


src/v0/destinations/linkedin_audience/transform.ts
```diff
@@ -25,10 +46,83 @@ function validateLinkedinAudienceEvent(event: unknown): LinkedinAudienceRecordRe
...
+      }
+    });
+    return userIds;
+  };
+
+  for (const [fieldName, value] of Object.entries(identifiers)) {
+    if (!USER_TRAITS_SET.has(fieldName)) {
+      throw new Error(`Invalid identifier key ${fieldName} for LinkedIn Audience.`);
+    }
+    if (value) {
```
shekhar-rudder: Lets use the `processAudienceRecord` util. The hashing function is hardcoded right now, and LinkedIn seems to use a different hashing approach than Google/FB—so we might need to tweak/refactor the util a bit.

shekhar-rudder (commented):
A few questions and suggestions across the hashing/batching/validation changes.

src/v0/destinations/linkedin_audience/config.ts
```diff
@@ -1,23 +1,40 @@
...
+  'organizationUrn',
+  'companyWebsiteDomain',
+  'companyEmailDomain',
+  'companyPageUrl',
+];
+export const COMPANY_TRAITS_SET = new Set(COMPANY_TRAITS);
+
 const BASE_URL = 'https://api.linkedin.com/rest';
 export const USER_ENDPOINT = (audienceId: string) => `${BASE_URL}/dmpSegments/${audienceId}/users`;
 export const USER_ENDPOINT_PATH = '/dmpSegments/<audienceId>/users';
```
shekhar-rudder, line 31: nit: `COMPANY_TRAITS_SET` seems unnecessary here. The array is only used for membership checks in one place, and with 5 items the O(1) vs O(n) difference is negligible. Consider just exporting `COMPANY_TRAITS` as an array and using `.includes()` directly — it avoids the extra export and keeps the config simpler.

src/v0/util/audienceUtils.ts
```diff
@@ -29,6 +30,21 @@ const PHONE_NUMBER_REGEX = /^\+?\d+$/;
  */
 export const isValidPhoneNumber = (value: string): boolean => PHONE_NUMBER_REGEX.test(value);
 
+function hashValue(value: string, hashingType: HashingType): string {
+  switch (hashingType) {
+    case HashingType.SHA256:
+      return sha256(value);
+    case HashingType.SHA512:
```
shekhar-rudder: `hashValue` and `HashingTypeToRegex` are two separate structures that both key off `HashingType` and need to stay in sync. Consider merging them into a single config map:

```ts
const HASHING_CONFIG: Partial<Record<HashingType, { regex: RegExp; hash: (v: string) => string }>> = {
  [HashingType.SHA256]: { regex: /^[\dA-Fa-f]{64}$/, hash: (v) => sha256(v) },
  [HashingType.SHA512]: { regex: /^[\dA-Fa-f]{128}$/, hash: (v) => createHash('sha512').update(v).digest('hex') },
  [HashingType.MD5]:    { regex: /^[\dA-Fa-f]{32}$/,  hash: (v) => createHash('md5').update(v).digest('hex') },
};
```

This eliminates the duplication and makes it impossible for the two to drift apart when a new hashing type is added.

src/v0/destinations/linkedin_audience/transform.ts
```diff
@@ -29,11 +45,82 @@ function validateLinkedinAudienceEvent(event: unknown): LinkedinAudienceRecordRe
...
+
+  const fieldConfigs: Record<string, AudienceField> = {};
+
+  Object.keys(identifiers).forEach((fieldName) => {
+    switch (fieldName) {
+      case 'sha256Email':
+        fieldConfigs[fieldName] = {
+          hashingType: HashingType.SHA256,
+          normalize: (value: string) => value.replace(REMOVE_SPACES_REGEX, '').toLowerCase(),
+          validate: (normalized: string) => validator.isEmail(normalized),
```
shekhar-rudder: This `switch` block building `fieldConfigs` could follow the `FB_FIELD_CONFIG` pattern from `fb_custom_audience/util.ts` — a `USER_FIELD_CONFIG: Record<string, AudienceField>` constant defined at module level (or in `config.ts`). This would:
- Keep `prepareUserTypePayload` lean
- Make the field configs easier to extend without touching the function body
- Stay consistent with the pattern already established in the codebase

src/v0/destinations/linkedin_audience/transform.ts
```diff
@@ -29,11 +45,82 @@ function validateLinkedinAudienceEvent(event: unknown): LinkedinAudienceRecordRe
...
-  const userIds = prepareUserIds(hashedIdentifiers);
+  const { message, connection, destination, metadata } = event;
+  const { isHashRequired } = connection.config.destination;
+  const { action, identifiers, fields } = message;
+
+  const prepareUserIds = (
+    ids: Record<string, string>,
+  ): {
+    idType: (typeof USER_IDENTIFIER_MAP)[keyof typeof USER_IDENTIFIER_MAP];
+    idValue: string;
```
shekhar-rudder: `prepareUserIds` is defined as an inner closure inside `prepareUserTypePayload`, which means it's recreated on every invocation. It doesn't close over any variables from the outer scope, so it could simply be a module-level function. This would also make it independently unit-testable.

src/v0/destinations/linkedin_audience/transform.ts
```diff
@@ -29,11 +45,82 @@ function validateLinkedinAudienceEvent(event: unknown): LinkedinAudienceRecordRe
...
+      workspaceId: metadata.workspaceId,
+      id: destination.ID,
+      type: DESTINATION_TYPE,
+      config: { isHashRequired },
+    },
+  }) as Record<string, string>;
+
+  if (Object.keys(processedIdentifiers).length === 0) {
+    throw new InstrumentationError('No identifiers found, aborting event.');
+  }
```
shekhar-rudder, line 114: `processAudienceRecord` returns `Record<string, unknown>`, so this `as Record<string, string>` cast silently bypasses TypeScript's type safety. Since the values are always strings at this point, the cleaner fix is to update `processAudienceRecord`'s return type to `Record<string, string>` directly, which would eliminate the need for the cast here.

src/v0/destinations/linkedin_audience/transform.ts
```diff
@@ -47,8 +134,18 @@ function prepareCompanyTypePayload(
   event: LinkedinAudienceRecordRequest,
 ): LinkedinAudienceCompanyPayload {
   const { action, identifiers, fields } = event.message;
-  const nonNullFields = removeNullValues(fields);
+  for (const fieldName of Object.keys(identifiers)) {
+    if (!COMPANY_TRAITS_SET.has(fieldName)) {
+      throw new Error(`Invalid identifier key ${fieldName} for LinkedIn Audience.`);
+    }
+  }
+
```
shekhar-rudder, line 136: `prepareCompanyTypePayload` manually validates keys and strips nulls but doesn't go through `processAudienceRecord`. Is this intentional? If it's by design (e.g. company fields have no normalization rules), a brief comment explaining the omission would help future readers.

src/v0/destinations/linkedin_audience/transform.ts
```diff
@@ -145,17 +242,20 @@ const processRouterDest = async (events: unknown[]): Promise<RouterTransformatio
...
-      const elementsPayload = {
-        elements: payloads.map((payload) => payload.payload),
-      };
-      const metadataList = payloads.map((payload) => payload.event.metadata);
-      const response = preparePayloadForProcessTransformation(elementsPayload, configParams);
+      const chunkedPayloads = lodash.chunk(payloads, MAX_BATCH_SIZE);
+      for (const chunk of chunkedPayloads) {
+        const elementsPayload = {
+          elements: chunk.map((payload) => payload.payload),
+        };
```
shekhar-rudder, line 245: Can you explain the motivation for chunking here? Is there a LinkedIn API limit on the number of elements per request that drives `MAX_BATCH_SIZE = 5000`? Is this pattern used in other record-type destinations (e.g. FB Custom Audience)? A brief comment on why this limit exists would help future maintainers.

test/integrations/destinations/linkedin_audience/router/data-native.ts
```diff
@@ -2706,4 +2708,512 @@ export const nativeData = [
       },
     },
   },
+  {
+    id: 'linkedin_audience-version-header-regression-test',
+    name: 'linkedin_audience',
+    description: 'Record call : LinkedIn-Version header uses 202603 (regression)',
+    scenario: 'Business',
+    successCriteria: 'should pass with 200 and include LinkedIn-Version 202603 header',
```
shekhar-rudder: The description says this is a regression test for the `LinkedIn-Version: 202603` header, but that header is already asserted in multiple existing tests in this file. What additional scenario does this test cover that the existing tests don't? If it's purely a version bump assertion, it may not need to be a separate test case.

src/v0/destinations/linkedin_audience/transform.ts
achettyiitr: > Can you explain the motivation for chunking here? Is there a LinkedIn API limit on the number of elements per request that drives MAX_BATCH_SIZE = 5000?

Yes. 
https://learn.microsoft.com/en-us/linkedin/marketing/matched-audiences/create-and-manage-segment-users?view=li-lms-2025-11&tabs=http#add-or-remove-multiple-users
https://learn.microsoft.com/en-us/linkedin/marketing/matched-audiences/create-and-manage-segment-companies?view=li-lms-2025-11&tabs=http#add-or-remove-multiple-companies

achettyiitr: Added comment.

src/v0/destinations/linkedin_audience/transform.ts
achettyiitr: For the company, there is no normalisation and Hashing requirements.

src/v0/destinations/linkedin_audience/transform.ts
shekhar-rudder: This comment looks unadressed

src/v0/destinations/linkedin_audience/transform.ts
```diff
@@ -29,11 +63,60 @@ function validateLinkedinAudienceEvent(event: unknown): LinkedinAudienceRecordRe
...
+    Object.keys(ids).forEach((key) => {
+      const value = ids[key];
+      if (value) {
+        userIds.push({ idType: USER_IDENTIFIER_MAP[key], idValue: value });
+      }
+    });
+    return userIds;
+  };
+
+  const fieldConfigs: Record<string, AudienceField> = {};
```
shekhar-rudder: Isn't value of `USERS_IDENTIFIER_CONFIG` and `fieldConfigs` same?

achettyiitr: Yeah, it's not needed. We only need this validation.

  ```
Object.keys(identifiers).forEach((fieldName) => {
    if (!USERS_IDENTIFIER_CONFIG[fieldName]) {
      throw new Error(`Invalid identifier key ${fieldName} for LinkedIn Audience.`);
    }
  });
```

shekhar-rudder (approved):

PR #5082, opened 2026-03-26
fix(facebook_lead_ads_native): add actual response to error log for debugging

shekhar-rudder:
## Summary
- Renamed `response` field to `responseSchema` in the non-OK Facebook API error log to clarify it contains the JSON schema
- Added new `response` field with `processedResponse.response` to log the actual API response body for better debugging

## Test plan
- [ ] Verify error logs now include the actual response body when Facebook API returns a non-OK status

### DIFF
--- a/src/sources/facebook_lead_ads_native/hydrate.ts
+++ b/src/sources/facebook_lead_ads_native/hydrate.ts
@@ -112,7 +112,8 @@ async function fetchLeadData(
   }
   logger.error(`[facebook_lead_ads_native] Non-OK response from Facebook API`, {
     status: processedResponse.status,
-    response: JSON.stringify(JsonSchemaGenerator.generate(processedResponse)),
+    responseSchema: JSON.stringify(JsonSchemaGenerator.generate(processedResponse)),
+    response: processedResponse.response,
   });
   // This should never be reached since errorResponseHandler always throws for errors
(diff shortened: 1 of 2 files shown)


achettyiitr (dismissed):

src/sources/facebook_lead_ads_native/hydrate.ts
```diff
@@ -112,7 +112,8 @@ async function fetchLeadData(
   }
   logger.error(`[facebook_lead_ads_native] Non-OK response from Facebook API`, {
     status: processedResponse.status,
-    response: JSON.stringify(JsonSchemaGenerator.generate(processedResponse)),
+    responseSchema: JSON.stringify(JsonSchemaGenerator.generate(processedResponse)),
+    response: processedResponse.response,
```
krishna2020, line 116: do we expect any sensitive data in the response ?

shekhar-rudder: This log won’t be triggered for every error scenario, only for cases that our code doesn’t currently handle. So far, for typical error scenarios, I haven’t seen Facebook return PII in error responses. I don’t expect it here either, but can’t be completely sure. We’ll need to add this to help debug 500 errors in production.

achettyiitr (approved):

PR #5091, opened 2026-03-30
fix(http): server fails to unmarshal response when headers contain non-string values

shekhar-rudder:
## Summary
- rudder-server unmarshals headers into `map[string]string`; non-string header values (numbers, booleans) from custom mappings caused unmarshalling errors for the HTTP destination
- Added `validateHeaders()` to detect and throw an `InstrumentationError` when a header value is not a string
- Converted `utils.js` / `utils.test.js` to TypeScript and extracted shared types into `types.ts`

## Test plan
- [ ] Unit tests updated in `utils.test.ts` covering valid and invalid header scenarios
- [ ] Integration test cases in `test/integrations/destinations/http/processor/configuration.ts` cover non-string header values
- [ ] Manual verification: configure HTTP destination with a numeric header value and confirm `InstrumentationError` is returned

### DIFF
--- a/src/cdk/v2/destinations/http/types.ts
+++ b/src/cdk/v2/destinations/http/types.ts
@@ -1,6 +1,6 @@
 export interface Mapping {
   from: string;
-  to: string
+  to: string;
 }
 
 export interface PathParam {
--- a/src/cdk/v2/destinations/http/types.ts
+++ b/src/cdk/v2/destinations/http/types.ts
@@ -1,5 +1,6 @@
 export interface Mapping {
   from: string;
+  to: string
 }
 
 export interface PathParam {
--- a/src/cdk/v2/destinations/http/types.ts
+++ b/src/cdk/v2/destinations/http/types.ts
@@ -1,0 +1,21 @@
+export interface Mapping {
+  from: string;
+}
+
+export interface PathParam {
+  path: string;
+}
+
+export interface BatchEvent {
+  batchedRequest: {
+    body: {
+      JSON: Record<string, unknown>;
+    };
+    method: string;
+    endpoint: string;
+    headers: Record<string, string>;
+    params: Record<string, unknown>;
+  };
+  metadata: unknown[];
+  destination: unknown;
+}
--- a/src/cdk/v2/destinations/http/types.ts
+++ b/src/cdk/v2/destinations/http/types.ts
@@ -5,17 +5,3 @@ export interface Mapping {
 export interface PathParam {
   path: string;
 }
-
-export interface BatchEvent {
-  batchedRequest: {
-    body: {
-      JSON: Record<string, unknown>;
-    };
-    method: string;
-    endpoint: string;
-    headers: Record<string, string>;
-    params: Record<string, unknown>;
-  };
-  metadata: unknown[];
-  destination: unknown;
-}
--- a/src/cdk/v2/destinations/http/utils.ts
+++ b/src/cdk/v2/destinations/http/utils.ts
@@ -1,15 +1,16 @@
-const { XMLBuilder } = require('fast-xml-parser');
-const { groupBy } = require('lodash');
-const { createHash } = require('crypto');
-const { ConfigurationError } = require('@rudderstack/integrations-lib');
-const { BatchUtils } = require('@rudderstack/workflow-engine');
-const jsonpath = require('rs-jsonpath');
-const {
+import { XMLBuilder } from 'fast-xml-parser';
+import { groupBy } from 'lodash';
+import { createHash } from 'crypto';
+import { ConfigurationError, InstrumentationError, isDefined } from '@rudderstack/integrations-lib';
+import { BatchUtils } from '@rudderstack/workflow-engine';
+import jsonpath from 'rs-jsonpath';
+import {
   base64Convertor,
   applyCustomMappings,
   isEmptyObject,
   removeUndefinedAndNullRecurse,
-} = require('../../../../v0/util');
+} from '../../../../v0/util';
+import type { Mapping, PathParam, BatchEvent } from './types';
 
 const CONTENT_TYPES_MAP = {
   JSON: 'JSON',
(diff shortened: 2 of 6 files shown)


src/cdk/v2/destinations/http/utils.ts
```diff
@@ -237,19 +250,35 @@ const batchSuccessfulEvents = (events, batchSize) => {
...
-module.exports = {
+// rudder-server unmarshals headers into map[string]string, so all values must be strings.
+// Non-string values (including numbers and booleans) will cause unmarshalling to fail.
+const validateHeaders = (headers: Record<string, unknown>): void => {
+  if (!headers || typeof headers !== 'object' || Array.isArray(headers)) {
+    return;
+  }
+  Object.entries(headers).forEach(([key, value]) => {
+    if (isDefined(value) && typeof value !== 'string') {
+      throw new InstrumentationError(
```
maheshkutty, line 269: I am inclined towards making it as ConfigurationError

maheshkutty: cc: @krishna2020 @saikumarrs

src/cdk/v2/destinations/http/utils.ts
```diff
@@ -237,19 +250,35 @@ const batchSuccessfulEvents = (events, batchSize) => {
...
-    batches.forEach((batch) => {
+    batches.forEach((batch: BatchEvent[]) => {
       response.push(buildBatchedRequest(batch));
     });
   });
   return response;
 };
 
-module.exports = {
+// rudder-server unmarshals headers into map[string]string, so all values must be strings.
```
maheshkutty, line 261: I am not sure in the future. Customers will complain about support of number and boolean values. In UI, we convert all non-dynamic values to strings; ideally, the same should happen in the backend for dynamic values

shekhar-rudder: > In UI, we convert all non-dynamic values to strings

By converting in the UI, do you mean that we persist all the values as string in the database?

> Customers will complain about support of number and boolean values.

They can stringify the number/boolean in the event itself, rt?

src/cdk/v2/destinations/http/utils.ts
saikumarrs: This situation is a bit on the fence. It could be due to either of the mistakes.

For the benefit of doubt, let's make it a configuration error, the events can at least be retried.

shekhar-rudder: > I am inclined towards making it as ConfigurationError

This is a runtime behavior, not a configuration error. A configuration issue would occur regardless of the event, but here the error depends on the incoming event. So `InstrumentationError` feels more apt.

shekhar-rudder: > For the benefit of doubt, let's make it a configuration error, the events can at least be retried.

Configuration error is currently a non-retryable error

saikumarrs: Is it only a small % of events that were failing or all of them? If it is former, InstrumentationError otherwise, ConfigurationError.

src/cdk/v2/destinations/http/utils.ts
maheshkutty: > By converting in the UI, do you mean that we persist all the values as string in the database?

Yes

maheshkutty: > They can stringify the number/boolean in the event itself, rt?

Some customers try to avoid conversion from their end because of their own reasons.

src/cdk/v2/destinations/http/utils.ts
shekhar-rudder: Till now we have seen this issue for only 1 destination. They had 33 success and 7 failures

src/cdk/v2/destinations/http/utils.ts
shekhar-rudder: To clarify, this doesn’t change any existing behavior. I'm only adding validations aligned with server behavior. Even prior to this PR, if the transformer returned a `number` or `boolean`, rudder-server would fail the event.

maheshkutty: I think we should handle this case where value can be a number and a boolean, not in this PR, maybe in some future PR.

maheshkutty: cc: @krishna2020

src/cdk/v2/destinations/http/types.ts
```diff
@@ -0,0 +1,21 @@
+export interface Mapping {
+  from: string;
```
maheshkutty, line 2: add `to` field

src/cdk/v2/destinations/http/utils.ts
```diff
@@ -1,24 +1,25 @@
...
+import { BatchUtils } from '@rudderstack/workflow-engine';
+import jsonpath from 'rs-jsonpath';
+import {
   base64Convertor,
   applyCustomMappings,
   isEmptyObject,
   removeUndefinedAndNullRecurse,
-} = require('../../../../v0/util');
+} from '../../../../v0/util';
+import type { Mapping, PathParam, BatchEvent } from './types';
```
maheshkutty: Can we use [BatchedRequestBody](https://github.com/rudderlabs/rudder-transformer/blob/fix/int-6134-fix-header-payload-format-validation/src/types/destinationTransformation.ts#L33) instead BatchEvent type?

shekhar-rudder: Addressed

src/cdk/v2/destinations/http/types.ts
shekhar-rudder: `to` is not being used anywhere in the code so I've skipped it.

maheshkutty (dismissed):

src/cdk/v2/destinations/http/types.ts
maheshkutty: When we call a template engine function, it uses the `to` field

shekhar-rudder: Added `to`

maheshkutty (approved):

PR #5093, opened 2026-03-30
refactor: onboard batching framework

ItsSudip:

Introduces a generic **Native Batching Framework** for the `/routerTransform` route that standardises validation, chunking, metadata resolution, and response formatting. Destinations can opt in incrementally via a registry map or per-destination environment variable. The new code path is gated behind a feature flag for incremental rollout.

Resolves INT-6025

Currently 50+ destinations each implement their own `processRouterDest` with duplicated batching logic — chunking, size limits, metadata alignment, and error handling. This framework extracts that into a reusable `BatchDestination<TBody>` abstract class so destinations only implement:

- `transformEvent()` — per-event transformation
- `getBatchStrategy()` — returns a `ChunkBatchStrategy` or `CustomBatchStrategy` instance
- `getInputSchema()` — Zod schema for integration-specific input validation

The framework handles iteration, validation, grouping by composite key `{endpoint, method, headers, params}`, strategy application, multiplexed event merging, error taxonomy, and REST envelope formatting.

### Architecture

```
POST /routerTransform
  → doRouterTransformation()
    → isBatchingFrameworkEnabled(destType, workspaceId)?
      YES → processBatchedDestination(events, IntegrationClass)
              1. Validate inputs (integration-specific Zod schema)
              2. Split on dontBatch flag
              3. transformEvents() → transformEvent() per event
              4. Filter failed jobIds from payloads (multiplexing safety)
              5. Group batchable payloads by {endpoint, method, headers, params}
              6. Apply BatchStrategy (ChunkBatchStrategy / CustomBatchStrategy)
              7. Convert to server format (BodyFormat enum: JSON/JSON_ARRAY/XML/FORM)
              8. Merge multiplexed events (combineBatchRequestsWithSameJobIds)
      NO  → legacy processRouterDest() path (unchanged)
```

### Feature flag

Per-destination env var: `{DEST}_BATCHING_FRAMEWORK_ENABLED_WORKSPACE_IDS`

- Comma-separated workspace IDs: `"ws-123,ws-456"` — enables for specific workspaces
- `"ALL"` — enables for all workspaces
- Not set / empty — disabled (legacy path)

OR logic with `batchedDestinationsMap`:
- **Pre-GA**: env var controls rollout per workspace
- **GA**: add destination to `batchedDestinationsMap` — always uses new path regardless of env var

- `FetchHandler` gains `batchDestinationHandlerMap` and `getBatchDestinationHandler()` for caching integration class constructors
- `MiscService` gains `getBatchDestinationHandler()` for dynamic module resolution
- `doRouterTransformation()` in `nativeIntegration.ts` checks the feature flag per destination group — when disabled, behaviour is identical to before

No. Uses existing dependencies: `zod`, `fast-json-stable-stringify`.

| Utility | File | Purpose |
|---------|------|---------|
| `BatchDestination<TBody>` | `src/services/destination/nativeBatching/batchDestination.ts` | Abstract class destinations extend |
| `ChunkBatchStrategy` | `src/services/destination/nativeBatching/chunkBatchStrategy.ts` | Strategy class for array chunking with maxItems/maxPayloadSize |
| `CustomBatchStrategy` | `src/services/destination/nativeBatching/customBatchStrategy.ts` | Strategy class for custom merge/dedup logic |
| `processBatchedDestination()` | `src/services/destination/nativeBatching/processBatchedDestination.ts` | Orchestration: validate → transform → group → batch → format |
| `isBatchingFrameworkEnabled()` | `src/constants/batchedDestinationsMap.ts` | Per-destination feature flag (env var OR registry) |
| `BodyFormat` enum | `src/services/destination/nativeBatching/types.ts` | Controls HTTP body envelope format (JSON/JSON_ARRAY/XML/FORM) |

Reuses existing: `combineBatchRequestsWithSameJobIds`, `groupRouterTransformEvents`, `handleRouterTransformSuccessEvents`, `generateErrorObject`.

- **Per-call instantiation**: `new IntegrationClass(destination)` per request to avoid race conditions with mutable state (important for destinations with dedup stores)
- **Prospective byte sizing**: `maxPayloadSize` measures the wrapped body size (`wrapBody([...current, candidate])`) before appending, ensuring accurate size limits
- **Multiplexing safety**: if a multiplexed event (one jobId producing payloads at multiple endpoints) partially fails, the failed jobId is removed from all success payloads before batching — preventing partial delivery
- **Error taxonomy**: uses `generateErrorObject` for consistent error classification with proper `statTags` (errorCategory, errorType)
- **Metrics**: emits `dont_batch_events`, `batch_size` (histogram), and `output_batches` for monitoring batching efficiency

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist


src/services/destination/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+export type ChunkStrategy<TBody extends Record<string, unknown> = Record<string, unknown>> = {
+  type: 'chunk';
+  maxSize?: number;
+  maxBytes?: string;
+  wrapBody: (bodies: TBody[]) => Record<string, unknown>;
+};
+
+export type CustomBatchResult = {
+  body: Record<string, unknown>;
+  jobIds: Set<number>;
```
shekhar-rudder: test comment 1

shekhar-rudder (commented):
test

src/services/destination/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+        }
+      } catch (error: any) {
+        errorEvents.push({
+          error: error.message || 'Unknown error during transformation',
+          statusCode: error.statusCode || 500,
+          jobId,
+          statTags: error.statTags,
+        });
+      }
+    }
```
shekhar-rudder: test comment 2

shekhar-rudder (commented):
test

src/services/destination/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+  // --- Framework-provided (not meant to be overridden) ---
+
+  validate(
+    inputs: RouterTransformationRequestData[],
+    errorResults: RouterTransformationResponse[],
+  ): RouterTransformationRequestData[] {
+    const integrationSchema = this.getIntegrationSchema();
+    const schema = integrationSchema ? baseInputSchema.and(integrationSchema) : baseInputSchema;
+
+    const validInputs: RouterTransformationRequestData[] = [];
```
shekhar-rudder: test comment 4

shekhar-rudder (commented):
test

shekhar-rudder (commented):
test

shekhar-rudder (commented):
test pos 233

shekhar-rudder (commented):
test pos 229

shekhar-rudder (commented):
test pos 230

shekhar-rudder (commented):
test pos 231

shekhar-rudder (commented):
test pos 232

shekhar-rudder (commented):
test pos 233

src/services/destination/nativeBatchingFramework/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+export type TransformResult<TBody extends Record<string, unknown> = Record<string, unknown>> = {
+  payloads: TransformedPayload<TBody>[];
+  errorEvents: TransformedErrorEvent[];
+};
+
+export type ChunkStrategy<TBody extends Record<string, unknown> = Record<string, unknown>> = {
+  type: 'chunk';
+  maxSize?: number;
+  maxBytes?: string;
+  wrapBody: (bodies: TBody[]) => Record<string, unknown>;
```
shekhar-rudder: **`ChunkStrategy.wrapBody` return type will block real destinations**

```ts
wrapBody: (bodies: TBody[]) => Record<string, unknown>;
```

Many REST APIs expect a top-level array body (Mixpanel `/import`, Braze `/users/track`, Amplitude `/batch`). The return type forces everything through an object wrapper. A destination needing `JSON_ARRAY` output has to do:

```ts
// hack: smuggle the array through an object key
wrapBody: (bodies) => ({ _items: bodies }),
buildRequestBody(body) { return { ..., JSON_ARRAY: { batch: body._items }, ... }; }
```

The `wrapBody` concern (how to combine N individual payloads into one) is conflated with the `buildRequestBody` concern (which HTTP body format to use). These should be separate — `wrapBody` should be allowed to return either `Record<string, unknown>` or `TBody[]`, and `buildRequestBody` decides the HTTP format from that.

src/services/destination/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+        const results = Array.isArray(transformedPayload)
+          ? transformedPayload
+          : [transformedPayload];
+        for (const result of results) {
+          payloads.push({ ...result, jobId });
+        }
+      } catch (error: any) {
+        errorEvents.push({
+          error: error.message || 'Unknown error during transformation',
+          statusCode: error.statusCode || 500,
```
shekhar-rudder: **Error handling bypasses the error taxonomy — stats will be degraded as more destinations onboard**

```ts
} catch (error: any) {
  errorEvents.push({
    statusCode: error.statusCode || 500,
    statTags: error.statTags,
  });
}
```

The legacy path goes through `handleRouterTransformFailureEvents`, which enriches `statTags` with `errorCategory`, `errorType`, `destType`, and emits `event_transform_failure`. The new path only passes through `error.statTags` if the thrown error already carries it — integrations throwing a plain `Error` will produce responses with no stat tags at all.

Consider using `getErrorStatusCode` (already in the util lib) or mirroring how `handleRtTfSingleEventError` works, so all onboarded destinations get consistent error classification.

src/services/destination/processBatchedDestination.ts
```diff
@@ -0,0 +1,233 @@
...
+          metadataMap,
+          destination,
+          integration,
+        ),
+      );
+    }
+  }
+
+  // 10. Merge responses that share jobIds (multiplexed events across groups)
+  return combineBatchRequestsWithSameJobIds(results);
```
shekhar-rudder: **`combineBatchRequestsWithSameJobIds` is called on mixed success + error results — latent bug for multiplexed events**

`results` at this point contains both success responses and error responses from `convertErrorEventsToResponses`. `combineBatchRequestsWithSameJobIds` merges entries by jobId overlap via `findExistingBatch`. For a multiplexed event (one jobId producing payloads at two endpoints) where one endpoint succeeds and the other fails, both responses share the same jobId — `findExistingBatch` matches them and merges the error's metadata into the success batch.

Error responses should be excluded from the merge:

```ts
const successResults = results.filter(r => !('error' in r));
const errorResults = results.filter(r => 'error' in r);
return [...combineBatchRequestsWithSameJobIds(successResults), ...errorResults];
```

src/services/destination/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+
+  getIntegrationSchema(): ZodType | null {
+    return null;
+  }
+
+  // --- Framework-provided (not meant to be overridden) ---
+
+  validate(
+    inputs: RouterTransformationRequestData[],
+    errorResults: RouterTransformationResponse[],
```
shekhar-rudder: **`validate` mutates an external array — surprising contract for a framework API**

```ts
validate(
  inputs: RouterTransformationRequestData[],
  errorResults: RouterTransformationResponse[],  // mutated in-place
): RouterTransformationRequestData[]
```

Callers must know that `errorResults` is modified as a side effect. If a destination eventually calls `super.validate()` inside an overridden `validate`, the mutation semantics are hard to reason about. A pure return signature would be a cleaner contract as the framework grows:

```ts
validate(inputs): { valid: RouterTransformationRequestData[]; errors: RouterTransformationResponse[] }
```

The caller then composes the results explicitly.

shekhar-rudder (commented):
Design review focused on extensibility ahead of onboarding multiple destinations.

src/constants/batchedDestinationsMap.ts
```diff
@@ -0,0 +1,28 @@
+export const batchedDestinationsMap: Record<string, true> = {};
+
+// Env var: BATCHING_FRAMEWORK_ENABLED_WORKSPACE_IDS
+// Values: comma-separated workspace IDs, or 'ALL' for all workspaces
+// Example: "workspace1,workspace2" or "ALL"
+// If not set or empty → disabled for all (legacy path)
+// Read at call time (not module load) to support runtime overrides and testing.
+const getEnabledWorkspaceIds = (): string[] =>
```
shekhar-rudder: **`isBatchingFrameworkEnabled` parses the env var on every call**

`getEnabledWorkspaceIds()` does a `split` + `map` + `filter` on `process.env.BATCHING_FRAMEWORK_ENABLED_WORKSPACE_IDS` on each invocation, and `isBatchingFrameworkEnabled` is called per destination group inside `mapInBatches`. With high throughput and many concurrent groups, this repeated string parsing compounds unnecessarily.

A module-level cached value (or reading once at startup) would avoid the overhead without sacrificing the testability benefit of not reading at module load time.

src/services/destination/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+      FORM: {},
+    };
+  }
+
+  async batchTransform(
+    inputs: RouterTransformationRequestData[],
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    _reqMetadata?: NonNullable<unknown>,
+  ): Promise<TransformResult<TBody>> {
+    const payloads: TransformedPayload<TBody>[] = [];
```
shekhar-rudder: **Naming suggestion:** Consider renaming `batchTransform` → `transformEvents`.

This is the most misleading name in the abstraction. The default implementation loops over `transformEvent` for each input individually — it doesn't do any batching. The actual batching is handled by the framework in `processBatchedDestination`.

Someone onboarding will see `batchTransform` and assume this is where they define batching logic, when in reality they should be looking at `getBatchStrategy` for that.

`transformEvents` accurately describes what the method does: transform multiple events.

src/services/destination/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+};
+
+export type ChunkStrategy<TBody extends Record<string, unknown> = Record<string, unknown>> = {
+  type: 'chunk';
+  maxSize?: number;
+  maxBytes?: string;
+  wrapBody: (bodies: TBody[]) => Record<string, unknown>;
+};
+
+export type CustomBatchResult = {
```
shekhar-rudder: **Naming suggestion:** Consider renaming `maxSize` → `maxItems` (or `maxEventsPerBatch`).

`maxSize` is ambiguous — size in bytes? count? character length? Since `maxBytes` already handles byte-based limits right below, this field should clearly indicate it's a **count** of events per chunk.

```ts
maxItems?: number;
```

shekhar-rudder (commented):
## Naming Review

Suggestions to improve naming clarity across the batching framework abstraction, focused on making the API more self-documenting for engineers onboarding new destinations.

src/services/destination/nativeBatchingFramework/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+};
+
+export type CustomBatchResult = {
+  body: Record<string, unknown>;
```
shekhar-rudder: **Naming suggestion:** Consider renaming `maxBytes` → `maxPayloadSize`.

This field accepts a **string** like `"4MB"`, not a number of bytes. The name `maxBytes` reads like it should be a `number`. `maxPayloadSize` better reflects the string-based format and avoids the type expectation mismatch.

```ts
maxPayloadSize?: string; // e.g. "4MB", "512KB"
```

src/services/destination/nativeBatchingFramework/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+
+    return { payloads, errorEvents };
+  }
+
+  getIntegrationSchema(): ZodType | null {
+    return null;
+  }
+
+  // --- Framework-provided (not meant to be overridden) ---
+
```
shekhar-rudder: **Naming suggestion:** Consider renaming `getIntegrationSchema` → `getInputSchema` or `getValidationSchema`.

"Integration schema" sounds like it describes the integration's configuration. It actually provides additional Zod validation for **input events**. Making that explicit helps users understand when and why to override it.

src/services/destination/nativeBatchingFramework/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+  /**
+   * Wraps the batched body into the BatchedRequestBody envelope.
+   * Default places body in JSON. Override for JSON_ARRAY, XML, FORM, etc.
+   */
+  buildRequestBody(body: Record<string, unknown>): BatchedRequestBody {
+    return {
+      JSON: body,
+      JSON_ARRAY: {},
+      XML: {},
+      FORM: {},
```
shekhar-rudder: **Naming suggestion:** Consider renaming `buildRequestBody` → `wrapRequestBody` or `buildRequestEnvelope`.

This method doesn't build the body — it wraps an already-built body into the transport envelope format (`JSON` / `JSON_ARRAY` / `XML` / `FORM`). "Build" implies construction from scratch. `wrapRequestBody` makes the intent immediately obvious.

src/services/destination/nativeBatchingFramework/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+// ---------------------------------------------------------------------------
+// Types
+// ---------------------------------------------------------------------------
+
+export type TransformedPayload<TBody extends Record<string, unknown> = Record<string, unknown>> = {
+  body: TBody;
+  endpoint: string;
+  method: string;
+  headers?: Record<string, unknown>;
+  params?: Record<string, unknown>;
```
shekhar-rudder: **Naming suggestion:** Consider renaming `TransformedPayload` → `TransformedEvent`.

"Payload" is overloaded in this codebase (HTTP payload? event payload? batch payload?). This type represents a single event after transformation, ready to be grouped and batched. `TransformedEvent` is more precise and aligns naturally with the `transformEvent` method that produces it.

src/services/destination/nativeBatchingFramework/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+  jobId: number;
+};
+
+export type TransformedErrorEvent = {
+  error: string;
+  statusCode: number;
+  jobId: number;
+  statTags?: Record<string, unknown>;
+};
```
shekhar-rudder: **Naming suggestion:** Consider renaming `TransformedErrorEvent` → `TransformError`.

Shorter and clearer. The "Event" suffix is confusing here since this isn't a RudderStack event — it's an error that occurred *during* transformation.

src/services/destination/nativeBatchingFramework/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+  }
+  return { batchable, nonBatchable };
+}
+
+export function resolveMetadatas(
+  jobIds: Set<number>,
+  metadataMap: Map<number, Partial<Metadata>>,
+): Partial<Metadata>[] {
+  return Array.from(jobIds).map((id) => {
+    const metadata = metadataMap.get(id);
```
shekhar-rudder: **Naming suggestion:** Consider renaming `resolveMetadatas` → `resolveMetadata`.

"Metadata" is already grammatically plural (metadatum → metadata). `resolveMetadatas` reads awkwardly.

src/services/destination/nativeBatchingFramework/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+  maxBytes?: string;
+  wrapBody: (bodies: TBody[]) => Record<string, unknown>;
+};
+
+export type CustomBatchResult = {
+  body: Record<string, unknown>;
+  jobIds: Set<number>;
+};
+
+export type CustomBatchStrategy<TBody extends Record<string, unknown> = Record<string, unknown>> = {
```
shekhar-rudder: **Naming suggestion:** Consider renaming `CustomBatchResult` → `BatchGroup`.

This type represents a group of events that should be sent as one request (a body + the jobIds that contributed to it). "Result" is vague — `BatchGroup` immediately communicates what it holds.

src/services/destination/nativeBatchingFramework/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+export type RouterIntegrationConstructor<
+  TBody extends Record<string, unknown> = Record<string, unknown>,
+> = new (destination: Destination) => RouterIntegration<TBody>;
+
+export abstract class RouterIntegration<
+  TBody extends Record<string, unknown> = Record<string, unknown>,
+> {
+  protected destination: Destination;
+
+  constructor(destination: Destination) {
```
shekhar-rudder: **Naming suggestion:** Consider renaming `RouterIntegration` → `BatchDestination` or `RouterDestinationAdapter`.

"Router" is RudderStack-internal jargon for the batching pipeline. A new engineer onboarding won't immediately know what "router" means in this context. `BatchDestination` makes the purpose self-evident: *"extend this to define how your destination handles batched events."*

src/services/destination/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+    return metadata;
+  });
+}
+
+type PayloadGroup<TBody extends Record<string, unknown>> = {
+  endpoint: string;
+  method: string;
+  headers: Record<string, unknown> | undefined;
+  params: Record<string, unknown> | undefined;
+  payloads: TransformedPayload<TBody>[];
```
shekhar-rudder: **Naming suggestion:** Consider renaming `PayloadGroup` → `RequestGroup`.

These are grouped by request-level properties (endpoint, method, headers, params). `RequestGroup` better reflects the grouping criteria.

src/constants/batchedDestinationsMap.ts
```diff
@@ -0,0 +1,40 @@
...
+    cachedWorkspaceIds =
+      envValue
+        ?.split(',')
+        ?.map((s) => s.trim())
+        ?.filter((s) => s) ?? [];
+  }
+  return cachedWorkspaceIds;
+};
+
+export const isBatchingFrameworkEnabled = (destType: string, workspaceId?: string): boolean => {
```
shekhar-rudder: workspaceId should not be optional

src/constants/batchedDestinationsMap.ts
```diff
@@ -0,0 +1,40 @@
...
+// Values: comma-separated workspace IDs, or 'ALL' for all workspaces
+// Example: "workspace1,workspace2" or "ALL"
+// If not set or empty → disabled for all (legacy path)
+// Cached to avoid repeated string parsing on every call.
+// Re-parses automatically if the env var value changes (e.g. in tests).
+let cachedWorkspaceIds: string[] | null = null;
+let cachedEnvValue: string | undefined;
+
+const getEnabledWorkspaceIds = (): string[] => {
+  const envValue = process.env.BATCHING_FRAMEWORK_ENABLED_WORKSPACE_IDS;
```
shekhar-rudder: BATCHING_FRAMEWORK_ENABLED_WORKSPACE_IDS should be at a destination type level

src/constants/batchedDestinationsMap.ts
shekhar-rudder: Lets make the logic either instead of AND

src/helpers/fetchHandlers.ts
```diff
@@ -9,6 +10,8 @@ export class FetchHandler {
 
   private static deletionHandlerMap: Map<string, any> = new Map();
 
+  private static routerTransformHandlerMap: Map<string, BatchDestinationConstructor> = new Map();
```
shekhar-rudder: See of names should be updated at other places too

src/services/destination/nativeIntegration.ts
```diff
@@ -102,12 +104,21 @@ export class NativeIntegrationDestinationService implements DestinationService {
...
-    const destHandler = FetchHandler.getDestHandler(destinationType, version);
     const groupedEvents: RouterTransformationRequestData[][] =
       await groupRouterTransformEvents(events);
+
     const response: RouterTransformationResponse[][] = await mapInBatches(
       groupedEvents,
       async (destInputArray: RouterTransformationRequestData[]) => {
+        const workspaceId = destInputArray[0].metadata?.workspaceId;
+        const useBatchingFramework = isBatchingFrameworkEnabled(destinationType, workspaceId);
+        const destHandler = useBatchingFramework
```
shekhar-rudder: See if the 3 ternary checks can be reduced to 1 check

src/services/destination/processBatchedDestination.ts
```diff
@@ -0,0 +1,265 @@
...
+// ---------------------------------------------------------------------------
+// Main orchestration function
+// ---------------------------------------------------------------------------
+
+export async function processBatchedDestination<
+  TBody extends Record<string, unknown> = Record<string, unknown>,
+>(
+  events: RouterTransformationRequestData[],
+  IntegrationClass: BatchDestinationConstructor<TBody>,
+  reqMetadata?: NonNullable<unknown>,
```
shekhar-rudder: `reqMetadata` should not optional

src/services/destination/processBatchedDestination.ts
```diff
@@ -0,0 +1,265 @@
...
+    allErrors.push(...batchResult.errorEvents);
+  }
+
+  // 5. Transform nonBatchable events individually and convert each to its own response
+  for (const single of nonBatchable) {
+    // eslint-disable-next-line no-await-in-loop
+    const singleResult = await integration.transformEvents([single], reqMetadata);
+    allErrors.push(...singleResult.errorEvents);
+
+    for (const payload of singleResult.payloads) {
```
shekhar-rudder: Push `singleResult.payloads` to `batchablePayloads`

src/services/destination/nativeBatchingFramework/processBatchedDestination.ts
```diff
@@ -0,0 +1,265 @@
...
+
+  // 1. Build metadata map
+  const metadataMap = new Map<number, Partial<Metadata>>();
+  for (const event of events) {
+    const metadata = event.metadata as Partial<Metadata>;
+    metadataMap.set(metadata.jobId!, metadata);
+  }
+
+  // 2. Validate inputs
+  const { valid: validInputs, errors: validationErrors } = integration.validate(events);
```
shekhar-rudder: Lets change the format of `validationErrors` to be same as `errorEvents`

src/services/destination/nativeBatchingFramework/processBatchedDestination.ts
```diff
@@ -0,0 +1,246 @@
+import type { Destination } from '../../../types/controlPlaneConfig';
```
shekhar-rudder: nativeBatchingFramework -> nativeBatching ?

src/services/destination/nativeBatchingFramework/processBatchedDestination.ts
```diff
@@ -0,0 +1,246 @@
...
+  reqMetadata: NonNullable<unknown>,
+): Promise<RouterTransformationResponse[]> {
+  if (events.length === 0) {
+    return [];
+  }
+  const { destination } = events[0];
+  const integration = new IntegrationClass(destination);
+
+  // 1. Build metadata map
+  const metadataMap = new Map<number, Partial<Metadata>>();
```
shekhar-rudder: Check if we can remove `Partial`

src/services/destination/nativeBatchingFramework/processBatchedDestination.ts
```diff
@@ -0,0 +1,246 @@
...
+  // 3. Split on dontBatch flag
+  const { batchableEvents, nonBatchableEvents } = groupByDontBatchDirective(validInputs);
+
+  // 4. Transform all events (batchable + nonBatchable)
+  let batchablePayloads: TransformedEvent<TBody>[] = [];
+  let nonBatchablePayloads: TransformedEvent<TBody>[] = [];
+
+  if (batchableEvents.length > 0) {
+    const batchResult = await integration.transformEvents(batchableEvents, reqMetadata);
+    batchablePayloads.push(...batchResult.payloads);
```
shekhar-rudder: (payloads, errorEvents) -> (successPayloads, errorPayloads)

src/services/destination/nativeBatchingFramework/processBatchedDestination.ts
```diff
@@ -0,0 +1,246 @@
...
+  // event's body would still appear in the batched request for the other endpoint.
+  const failedJobIds = new Set(allErrors.map((e) => e.jobId));
+  if (failedJobIds.size > 0) {
+    batchablePayloads = batchablePayloads.filter((p) => !failedJobIds.has(p.jobId));
+    nonBatchablePayloads = nonBatchablePayloads.filter((p) => !failedJobIds.has(p.jobId));
+  }
+
+  // 7. Convert nonBatchable payloads to individual responses
+  const successResponses: RouterTransformationResponse[] = [];
+  for (const payload of nonBatchablePayloads) {
```
shekhar-rudder: We are duplicating logic for batchable and non batchable payloads

src/services/destination/nativeBatchingFramework/processBatchedDestination.ts
```diff
@@ -0,0 +1,246 @@
...
+  // 8. Group batchable payloads by composite key and apply batch strategy
+  if (batchablePayloads.length > 0) {
+    const groups = groupPayloadsByCompositeKey(batchablePayloads);
+
+    for (const group of groups) {
+      const strategy = integration.getBatchStrategy(group.endpoint);
+
+      const batchResults =
+        strategy.type === 'chunk'
+          ? applyChunkStrategy(group.payloads, strategy)
```
shekhar-rudder: Lets implement `batch` method for chunk strategy as well so that we can get rid of the ternary logic

src/services/destination/nativeBatchingFramework/processBatchedDestination.ts
```diff
@@ -0,0 +1,246 @@
...
+    batched: true,
+    statusCode: 200,
+  };
+}
+
+// ---------------------------------------------------------------------------
+// Error event conversion
+// ---------------------------------------------------------------------------
+
+function convertErrorEventsToResponses(
```
shekhar-rudder: mapErrorPayloadToServerFormat, mapSuccessPayloadToServerFormat

src/services/destination/nativeBatchingFramework/processBatchedDestination.ts
```diff
@@ -0,0 +1,246 @@
...
+        jobIds: new Set(currentJobIds),
+      });
+      currentBodies = [];
+      currentJobIds = new Set();
+    }
+  };
+
+  for (const payload of payloads) {
+    // Check maxItems before adding
+    if (currentBodies.length > 0 && currentBodies.length >= maxItems) {
```
shekhar-rudder: `currentBodies.length > 0` can be removed from line 50 and 55

src/services/destination/nativeBatchingFramework/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+// ---------------------------------------------------------------------------
+// Abstract class: BatchDestination<TBody>
+// ---------------------------------------------------------------------------
+
+// Constructor type for BatchDestination subclasses — used by the framework to instantiate per request
+export type BatchDestinationConstructor<
+  TBody extends Record<string, unknown> = Record<string, unknown>,
+> = new (destination: Destination) => BatchDestination<TBody>;
+
+export abstract class BatchDestination<
```
shekhar-rudder: File should be renamed to `batchDestination.ts`

src/services/destination/nativeBatchingFramework/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+
+  abstract getBatchStrategy(endpoint: string): BatchStrategy<TBody>;
+
+  // --- MAY override ---
+
+  /**
+   * Wraps the batched body into the BatchedRequestBody envelope.
+   * Default places body in JSON. Override for JSON_ARRAY, XML, FORM, etc.
+   */
+  wrapRequestBody(body: Record<string, unknown>): BatchedRequestBody {
```
shekhar-rudder: We are splitting the final payload logic in 2 functions - wrapBody and wrapRequestBody. Lets update wrapBody to also have the capability to override body type. Then we don't need `wrapRequestBody` method in this class

src/services/destination/nativeBatchingFramework/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+> {
+  protected destination: Destination;
+
+  constructor(destination: Destination) {
+    this.destination = destination;
+  }
+
+  // --- MUST implement ---
+
+  abstract transformEvent(
```
shekhar-rudder: Lets also add `_reqMetadata` argument

src/services/destination/nativeBatchingFramework/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+          jobId,
+          statTags: errObj.statTags,
+        });
+      }
+    }
+
+    return { payloads, errorEvents };
+  }
+
+  getInputSchema(): ZodType | null {
```
shekhar-rudder: Lets remove `null` return type and make it a abstract method

src/services/destination/nativeBatchingFramework/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+    return null;
+  }
+
+  // --- Framework-provided (not meant to be overridden) ---
+
+  validate(inputs: RouterTransformationRequestData[]): {
```
shekhar-rudder: this method should be moved to framework

src/services/destination/nativeBatchingFramework/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+        const errorMessage = [
+          ...new Set(
+            parseResult.error.issues.map((issue) => {
+              const path = issue.path.length > 0 ? `${issue.path.join('.')}: ` : '';
+              return `${path}${issue.message}`;
+            }),
+          ),
+        ].join('; ');
+        errors.push({
+          jobId: (input.metadata as Partial<Metadata>)?.jobId ?? 0,
```
shekhar-rudder: typecasting can be removed

src/services/destination/nativeBatchingFramework/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+
+type RequestGroup<TBody extends Record<string, unknown>> = {
+  endpoint: string;
+  method: string;
+  headers: Record<string, unknown> | undefined;
+  params: Record<string, unknown> | undefined;
+  payloads: TransformedEvent<TBody>[];
+};
+
+export function groupPayloadsByCompositeKey<
```
shekhar-rudder: Any type/method not being used by `BatchDestination` class should not be in this file

src/services/destination/nativeBatchingFramework/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+  method: string;
+  headers?: Record<string, unknown>;
+  params?: Record<string, unknown>;
+  jobId: number;
+};
+
+export type TransformError = {
+  error: string;
+  statusCode: number;
+  jobId: number;
```
shekhar-rudder: Similar to success case, lets remove jobId

src/services/destination/nativeBatchingFramework/routerIntegration.ts
```diff
@@ -0,0 +1,291 @@
...
+
+export type BatchStrategy<TBody extends Record<string, unknown> = Record<string, unknown>> =
+  | ChunkStrategy<TBody>
+  | CustomBatchStrategy<TBody>;
+
+// ---------------------------------------------------------------------------
+// Factory functions
+// ---------------------------------------------------------------------------
+
+export function chunk<TBody extends Record<string, unknown> = Record<string, unknown>>(opts: {
```
shekhar-rudder: We should create classes for strategy which integrations can instantiate

src/services/destination/nativeIntegration.ts
```diff
@@ -102,12 +104,16 @@ export class NativeIntegrationDestinationService implements DestinationService {
...
-    const destHandler = FetchHandler.getDestHandler(destinationType, version);
     const groupedEvents: RouterTransformationRequestData[][] =
       await groupRouterTransformEvents(events);
+
     const response: RouterTransformationResponse[][] = await mapInBatches(
       groupedEvents,
       async (destInputArray: RouterTransformationRequestData[]) => {
+        const workspaceId = destInputArray[0].metadata?.workspaceId;
+        const useBatchingFramework =
+          workspaceId && isBatchingFrameworkEnabled(destinationType, workspaceId);
```
shekhar-rudder: non-null assertion on workspaceId

src/services/destination/nativeIntegration.ts
```diff
@@ -116,10 +122,24 @@ export class NativeIntegrationDestinationService implements DestinationService {
         );
         try {
           metaTO.metadata = destInputArray[0].metadata;
-          const doRouterTransformationResponse: RouterTransformationResponse[] =
-            await destHandler.processRouterDest(destInputArray, requestMetadata);
+          let transformedResponse: RouterTransformationResponse[];
+          let destHandler: any = null;
+          if (useBatchingFramework) {
```
shekhar-rudder, line 128: Lets add a comment calling it out explicitly that `destHandler` is null for batching case

shekhar-rudder (commented):
chore -> refactor

src/services/destination/nativeBatching/processBatchedDestination.ts
```diff
@@ -0,0 +1,302 @@
...
+            strategy.bodyFormat,
+          ),
+        );
+      }
+    }
+  }
+
+  // Metric: output batches — counted before merge since each entry is one HTTP request.
+  // After merge, a single response can contain multiple batchedRequests (multiplexed events).
+  stats.counter('output_batches', successResponses.length, metricTags);
```
shekhar-rudder: This metric is counting the number of successful events. Is this a useful metric?

src/services/destination/nativeBatching/processBatchedDestination.ts
```diff
@@ -0,0 +1,302 @@
...
+  const allPayloads = successPayloads.filter((p) => !failedJobIds.has(p.jobId));
+
+  // 7. Group all payloads by composite key and apply batch strategy
+  const successResponses: RouterTransformationResponse[] = [];
+  if (allPayloads.length > 0) {
+    const groups = groupPayloadsByCompositeKey(allPayloads);
+
+    for (const group of groups) {
+      const strategy = integration.getBatchStrategy(group.endpoint);
+      const batchResults = strategy.batch(group.payloads);
```
shekhar-rudder: I thought we decided that we will call batch method separately for batchable and non-batchable events. With this implementation, won't non-batchable events get merged with the rest?

src/services/destination/nativeBatching/processBatchedDestination.ts
```diff
@@ -0,0 +1,302 @@
...
+  }
+
+  return { valid, errors };
+}
+
+// ---------------------------------------------------------------------------
+// Helper functions
+// ---------------------------------------------------------------------------
+
+export function groupByDontBatchDirective(inputs: RouterTransformationRequestData[]): {
```
shekhar-rudder, line 64: Some of the methods in this file need not be exported anymore

src/services/destination/nativeBatching/processBatchedDestination.ts
```diff
@@ -0,0 +1,302 @@
...
+
+// ---------------------------------------------------------------------------
+// Validation
+// ---------------------------------------------------------------------------
+
+export function validateInputs<TBody extends Record<string, unknown>>(
+  inputs: RouterTransformationRequestData[],
+  integration: BatchDestination<TBody>,
+): { valid: RouterTransformationRequestData[]; errors: (TransformError & { jobId: number })[] } {
+  const schema = baseInputSchema.and(integration.getInputSchema());
```
shekhar-rudder: We should use the `baseInputSchema` in `processBatchedDestination` and not here. `events` in that function should be typed as `unknown` and only after validating with `baseInputSchema`, they should be considered as `RouterTransformationRequestData[]`

src/services/destination/nativeBatching/__tests__/processBatchedDestination.test.ts
```diff
@@ -0,0 +1,385 @@
...
+  }
+}
+
+function makeInput(
+  jobId: number,
+  data: string,
+  opts?: { type?: string; dontBatch?: boolean; shouldFail?: boolean },
+): RouterTransformationRequestData {
+  return {
+    message: { data, type: opts?.type || 'track', shouldFail: opts?.shouldFail } as any,
```
shekhar-rudder: Lets remove `any` from all the test code

src/services/destination/nativeBatching/__tests__/batchingFrameworkFeatureFlag.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    } else {
+      process.env[envKey] = originalEnv;
+    }
+    delete batchedDestinationsMap['TEST_DEST'];
+  });
+
+  describe('pre-GA: env var based rollout', () => {
+    it('returns false when env var is not set', () => {
+      delete process.env[envKey];
+      expect(isBatchingFrameworkEnabled('TEST_DEST', 'ws-1')).toBe(false);
```
shekhar-rudder: Consider using table driven tests to reduce boiler plate code

src/services/destination/nativeBatching/processBatchedDestination.ts
ItsSudip: That is fine. We can use this to understand the batching efficiency with the input events metric. Input events metric is already getting captured [here](https://github.com/rudderlabs/rudder-transformer/blob/412de48d3b4de69eb506556b9ffc5e942ff1701d/src/controllers/destination.ts#L91-L96).

src/services/destination/nativeBatching/processBatchedDestination.ts
ItsSudip: We need to export the functions for tests, right?

src/services/destination/nativeBatching/processBatchedDestination.ts
ItsSudip: The processBatchedDestination function is only called from nativeIntegration.ts, where the events are already typed as RouterTransformationRequestData[] by the HTTP layer and Koa middleware. Typing the parameter as unknown[] would push the type assertion upstream to nativeIntegration.ts (which would need to cast), rather than adding real safety.

The baseInputSchema validation inside validateInputs already catches malformed events at runtime (missing message.type, missing metadata.jobId, etc.) and produces proper error responses. So we get runtime validation without losing compile-time type safety at the call site.

If we later expose processBatchedDestination as a public API with external callers, we can revisit typing the parameter as unknown[]. For now the current approach keeps the internal contract clean.

src/services/destination/processBatchedDestination.ts
```diff
@@ -0,0 +1,265 @@
...
+      if (cleanedMetadata.length > 0) {
+        cleanedSuccessResults.push({ ...resp, metadata: cleanedMetadata });
+      }
+    }
+  } else {
+    cleanedSuccessResults.push(...successResults);
+  }
+
+  // 12. Merge success responses that share jobIds (multiplexed events across groups)
+  return [...combineBatchRequestsWithSameJobIds(cleanedSuccessResults), ...errorResults];
```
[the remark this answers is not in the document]
ItsSudip: This is already addressed can you check?

src/services/destination/processBatchedDestination.ts
```diff
@@ -0,0 +1,233 @@
...
+          payload.endpoint,
+          payload.method,
+          payload.headers,
+          payload.params,
+          metadataMap,
+          destination,
+          integration,
+        ),
+      );
+    }
```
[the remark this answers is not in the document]
ItsSudip: This is removed and merged with the batchable events. It has already been addressed.

src/services/destination/nativeBatchingFramework/routerIntegration.ts
```diff
@@ -0,0 +1,288 @@
...
+        for (const result of results) {
+          payloads.push({ ...result, jobId });
+        }
+      } catch (error: any) {
+        errorEvents.push({
+          error: error.message || 'Unknown error during transformation',
+          statusCode: error.statusCode || 500,
+          jobId,
+          statTags: error.statTags,
+        });
```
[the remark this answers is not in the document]
ItsSudip: The transformEvent will throw error with 4xx status code. 500 status code is for unhandled scenarios.

src/services/destination/nativeBatching/__tests__/batchDestination.test.ts
```diff
@@ -0,0 +1,308 @@
...
+  Config: { apiKey: 'test-key' },
+  DestinationDefinition: { Name: 'TEST' },
+} as unknown as Destination;
+
+class TestIntegration extends BatchDestination<TestBody> {
+  transformEvent(
+    input: RouterTransformationRequestData,
+  ): Omit<TransformedEvent<TestBody>, 'jobId'> {
+    return {
+      body: { value: (input.message as any).data },
```
shekhar-rudder: This file still has `any`

src/services/destination/nativeBatching/processBatchedDestination.ts
shekhar-rudder: Shouldn't we use the same name for the metric (`output_batches` Vs `dest_transform_input_events`) irrespective of the underlying framework being used?

src/services/destination/nativeBatching/processBatchedDestination.ts
```diff
@@ -0,0 +1,314 @@
...
+  const successBatchablePayloads = batchablePayloads.filter((p) => !failedJobIds.has(p.jobId));
+  const successNonBatchablePayloads = nonBatchablePayloads.filter(
+    (p) => !failedJobIds.has(p.jobId),
+  );
+
+  // 7. Build request groups:
+  // - Batchable payloads are grouped by composite key (endpoint, method, headers, params)
+  // - NonBatchable payloads each form their own group to prevent merging with other events
+  const batchableGroups = groupPayloadsByCompositeKey(successBatchablePayloads);
+  const nonBatchableGroups = successNonBatchablePayloads.map((payload) => ({
```
shekhar-rudder: How about `const nonBatchableGroups = successNonBatchablePayloads.map((payload) => groupPayloadsByCompositeKey([payload])).flat();` to avoid code duplication

src/services/destination/nativeBatching/processBatchedDestination.ts
shekhar-rudder: > Typing the parameter as unknown[] would push the type assertion upstream to nativeIntegration.ts

I didn't understand this part. If you type the parameter as `unknown`, it doesn't impact `nativeIntegration`. Upstream can continue to send events as it is without any code changes or typecasting

> The baseInputSchema validation inside validateInputs already catches malformed events at runtime

This is happening too late. Assuming that the input is malformed, the code will throw an error because of statements like `const { destination } = events[0];` and `const { workspaceId } = events[0].metadata;`.

I'm fine with not doing `baseInputSchema` at all. But with the current implementation, I don't think its helpful

src/services/destination/nativeBatching/processBatchedDestination.ts
ItsSudip: I wanted to use a different name to differentiate among frameworks. But I think that is already done by the destType. I mean at a time one destType will be on one framework only after GA.

src/services/destination/nativeBatching/__tests__/batchDestination.test.ts
```diff
@@ -0,0 +1,287 @@
...
+    });
+  }
+  return map;
+}
+
+// ---------------------------------------------------------------------------
+// Tests
+// ---------------------------------------------------------------------------
+
+describe('parseSizeToBytes', () => {
```
shekhar-rudder, line 106: Since some of the methods were moved out of batchDestination, same should be done for tests

src/constants/__tests__/batchedDestinationsMap.test.ts
```diff
@@ -0,0 +1,111 @@
+import {
+  batchedDestinationsMap,
+  isBatchingFrameworkEnabled,
+} from '../../../../constants/batchedDestinationsMap';
+
+describe('isBatchingFrameworkEnabled', () => {
```
shekhar-rudder, line 3: This file should be in constants directory and not here. Or otherwise we should `batchedDestinationsMap` file from `constants` to `nativeBatching` directory

shekhar-rudder (approved):

PR #5095, opened 2026-03-30
feat: onboard posthog to the batching framework

ItsSudip:

Onboard posthog to the new batching framework.

Resolves INT-6026

N/A

N/A

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist


src/v0/destinations/posthog/routerTransform.ts
```diff
@@ -0,0 +1,55 @@
...
+  BatchDestination,
+  TransformedEvent,
+  ChunkBatchStrategy,
+} from '../../../services/destination/nativeBatching/batchDestination';
+import type { BatchStrategy } from '../../../services/destination/nativeBatching/types';
+import { processEvent } from './transform';
+import type { PostHogPayload, PostHogProcessorRequest } from './types';
+
+class PostHogIntegration extends BatchDestination<PostHogPayload> {
+  transformEvent(input: PostHogProcessorRequest): TransformedEvent<PostHogPayload> {
```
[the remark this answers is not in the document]
ItsSudip: Updated with PostHogRouterRequest.

src/v0/destinations/posthog/routerTransform.ts
```diff
@@ -0,0 +1,55 @@
...
+  transformEvent(input: PostHogRouterRequest): TransformedEvent<PostHogPayload> {
+    const result = processEvent(input.message, input.destination);
+    // Strip api_key from the body — it belongs in the wrapBody wrapper
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    const { api_key, ...eventBody } = result.body.JSON;
+    return {
+      body: eventBody,
+      endpoint: result.endpoint,
+      method: result.method,
+      headers: result.headers,
```
shekhar-rudder, line 21: **Two issues with the `api_key` destructuring**

**1. `processEvent` lacks an explicit return type, so `api_key` is inferred from `{}`**

`processEvent` in `transform.ts` has no return type annotation. That means `result.body.JSON` resolves to `{}` here, and destructuring gives you `api_key` without a meaningful type. Adding an explicit return type (e.g. `ProcessorTransformationOutput`) to `processEvent` would make this strongly typed.

**2. `PostHogPayload` doesn't include `type`, but every batch item has one**

Only `api_key` is stripped here — `type` (e.g. `"capture"`, `"alias"`) stays in `eventBody` and lands in the `batch` array. But `PostHogPayload` is documented as *"Payload shape before `api_key` and `type` are added"*, so the generic `BatchDestination<PostHogPayload>` doesn't match what's actually in each batch item at runtime. Consider adding `type: string` to `PostHogPayload` (or introducing a `PostHogBatchItem` type) to close the gap.

src/v0/destinations/posthog/routerTransform.ts
```diff
@@ -0,0 +1,55 @@
...
+        message: z
+          .object({
+            userId: z.union([z.string(), z.number()]).optional(),
+            anonymousId: z.union([z.string(), z.number()]).optional(),
+            type: z.string().refine((val) => val !== 'record', {
+              message: 'message type should not be record',
+            }),
+          })
+          .passthrough()
+          .refine((msg) => !!msg.userId || !!msg.anonymousId, {
```
shekhar-rudder, line 45: **Use an allowlist instead of a blocklist for `type` validation**

Blocking only `"record"` means unrecognised types (e.g. `"audienceList"`) pass Zod validation silently and fall through to `processEvent`, which throws a generic `InstrumentationError`. An allowlist gives a clear, early failure for anything outside the known set:
```ts
type: z.enum(['track', 'page', 'screen', 'identify', 'alias', 'group'])
```
This also makes the supported event types self-documenting in the schema.

test/integrations/destinations/posthog/router/data.ts
```diff
@@ -1,4 +1,406 @@
+const batchingFrameworkTests = [
+  {
+    name: 'posthog',
+    description: 'Batching Framework: Single alias event produces batched output with wrapper',
+    feature: 'router',
```
shekhar-rudder, line 5: **No need for a separate `batchingFrameworkTests` constant**

The intermediate constant is only used immediately as `...batchingFrameworkTests` in `data`. Just add the new test cases directly into the `data` array — it keeps the file structure consistent with other destination test files and avoids the indirection.

src/v0/destinations/posthog/routerTransform.test.ts
```diff
@@ -0,0 +1,199 @@
...
+// ---------------------------------------------------------------------------
+// Tests
+// ---------------------------------------------------------------------------
+
+describe('PostHog Batching Framework Integration', () => {
+  describe('single event', () => {
+    it('produces a batch of 1 with wrapper envelope', async () => {
+      const inputs = [makeEvent(1, 'track')];
+      const results = await processBatchedDestination(inputs, PostHogIntegration, {});
+
```
shekhar-rudder: **Two structural concerns with this test file**

**1. Overlap with `data.ts`**
The integration tests in `test/integrations/destinations/posthog/router/data.ts` already cover the batching-framework path (single alias, multiple tracks, validation error). Several cases here look similar — worth checking whether there's meaningful duplication and if some tests belong in one place or the other.

**2. The subject under test should be `PostHogIntegration`, not `processBatchedDestination`**
Calling `processBatchedDestination(inputs, PostHogIntegration, {})` exercises the whole framework pipeline. That's what integration tests are for. This unit test file should test `PostHogIntegration` in isolation — e.g. `transformEvent`, `getBatchStrategy`, `getInputSchema` — so failures here point directly at the integration class, not the framework.

shekhar-rudder (commented):
Overall clean implementation — workspace-based rollout, `api_key` stripping, and the batch wrapper envelope are all correct. A few comments below on type safety, input validation, and test structure.

src/v0/destinations/posthog/transform.ts
```diff
@@ -142,22 +142,32 @@ const responseBuilderSimple = (
...
     type: category.type,
   };
-  const response = defaultRequestConfig();
-  response.endpoint = `${
-    stripTrailingSlash(destination.Config.yourInstance) || DEFAULT_BASE_ENDPOINT
-  }/batch`;
-  response.method = defaultPostRequestConfig.requestMethod;
-  response.headers = {
-    'Content-Type': JSON_MIME_TYPE,
+  return {
```
shekhar-rudder: You need not change this piece of code to satisfy Typescript since `defaultRequestConfig` already takes care of it. Just typecast the output of `defaultRequestConfig` method

src/v0/destinations/posthog/routerTransform.test.ts
```diff
@@ -0,0 +1,161 @@
...
+      (input.message as Record<string, unknown>).type = type;
+      const schema = integration.getInputSchema();
+      const parseResult = schema.safeParse(input);
+
+      expect(parseResult.success).toBe(false);
+    });
+
+    it('rejects events missing both userId and anonymousId', () => {
+      const input = makeInput(1, 'track');
+      Reflect.deleteProperty(input.message, 'userId');
```
shekhar-rudder: Why use `Reflect.deleteProperty` instead of deleting property directly?

src/v0/destinations/posthog/routerTransform.test.ts
```diff
@@ -0,0 +1,161 @@
...
+      const input = makeInput(1, type, overrides);
+      const result = integration.transformEvent(input);
+
+      expect(result.endpoint).toBe('https://app.posthog.com/batch');
+      expect(result.method).toBe('POST');
+      expect(result.headers).toEqual({ 'Content-Type': 'application/json' });
+      // api_key should be stripped from the event body
+      expect(result.body).not.toHaveProperty('api_key');
+      // type (e.g. 'capture', 'alias') should remain in the event body
+      expect(result.body.type).toBeDefined();
```
shekhar-rudder: Instead of separate assertion for each field, you can have a single assertion for the whole object `result`. This way you ensure that you don't miss any fields

src/v0/destinations/posthog/routerTransform.test.ts
ItsSudip: Used Reflect.deleteProperty as delete was showing typescript error of deleting a required property.

ItsSudip: Anyway used typecasting.

src/v0/destinations/posthog/transform.ts
```diff
@@ -151,13 +152,16 @@ const responseBuilderSimple = (
     'Content-Type': JSON_MIME_TYPE,
   };
   response.body.JSON = responseBody;
-  return response;
+  return response as PostHogTransformOutput;
```
shekhar-rudder: We should typecast output of `defaultRequestConfig` method only because it is in Javascript. We shouldn't be typecasting here

src/v0/destinations/posthog/routerTransform.test.ts
```diff
@@ -0,0 +1,162 @@
...
+    it.each<{ type: RudderMessage['type']; overrides?: Partial<PostHogMessage> }>([
+      { type: 'track' },
+      { type: 'page' },
+      { type: 'screen' },
+      { type: 'alias', overrides: { previousId: 'prev-1' } },
+    ])('transforms $type event correctly', ({ type, overrides }) => {
+      const input = makeInput(1, type, overrides);
+      const result = integration.transformEvent(input);
+
+      expect(result).toMatchObject({
```
shekhar-rudder: Lets use `toEqual` instead of `toMatchObject` to match all the fields

src/v0/destinations/posthog/routerTransform.test.ts
shekhar-rudder: In that case I'm okay with the options

ItsSudip: Didn't get you. You are okay with typeCasting or reflect.deleteProperty?

shekhar-rudder: My bad, I missed "both". I'm okay with both the options - typeCasting and reflect.deleteProperty

shekhar-rudder (approved):

PR #5103, opened 2026-04-01
feat(fb-custom-audience): send payload in request body to support larger audiences

shekhar-rudder:
## Summary
- FB's query param size limit (~60-70KB) causes API errors well before the official 10,000 record limit; moving the payload to `body.JSON` removes this constraint
- Added `src/util/featureFlags.ts` — a shared utility for env-var-based feature flags (supports `ALL`, `NONE`, or comma-separated workspace IDs)
- Refactored `google_adwords_enhanced_conversions` to use the same utility, removing its bespoke `utils.ts`

## Test plan
- [ ] Unit tests in `util.test.ts` cover flag enabled for all workspaces
- [ ] Integration test in `router/data.ts` covers end-to-end payload-in-body behavior with `FB_CUSTOM_AUDIENCE_PAYLOAD_IN_BODY=ALL`
- [ ] Verify GAEC behavior unchanged with `DEST_GAEC_ADJUSTMENT_TYPE_SUPPORTED_WORKSPACE_IDS` flag


src/util/featureFlags.ts
```diff
@@ -0,0 +1,24 @@
...
+  const val = process.env[flagName];
+  if (!val) {
+    return false;
+  }
+
+  const upper = val.trim().toUpperCase();
+  if (upper === 'ALL') {
+    return true;
+  }
+  if (upper === 'NONE') {
```
ItsSudip: I don't think it is needed. This is the same as the default scenario.

shekhar-rudder: Addressed

ItsSudip (approved):

PR #5126, opened 2026-04-10
fix(gaoc): wrong conversion action used when MCC returns duplicate actions

shekhar-rudder:
## Summary
- When using MCC (sub-account), the Google Ads SearchStream API can return conversion actions with the same name from both the target account (`customerId`) and the manager account (`loginCustomerId`). The last result would overwrite the first, potentially using the wrong conversion action for uploads.
- Added `conversion_action.owner_customer` to the GAQL query and prefer the action owned by `customerId` when duplicates exist. Falls back to MCC-owned action if no customer-owned action is found.
- Added table-driven unit tests for `getConversionActionIds` covering: customerId preference regardless of order, MCC fallback, and empty input.

## Test plan
- [x] Unit tests pass (`npx jest src/v0/destinations/google_adwords_offline_conversions/utils.test.js -t "getConversionActionIds"`)
- [ ] Verify with affected customer's account that the correct conversion action ID is selected

### DIFF
--- a/src/v0/destinations/google_adwords_offline_conversions/utils.js
+++ b/src/v0/destinations/google_adwords_offline_conversions/utils.js
@@ -591,6 +591,26 @@ const getListCustomVariable = ({ properties, conversionCustomVariableMap, custom
 };
 
 /**
+ * Build a name→resourceName map from SearchStream results, preferring entries owned by customerId.
+ * When using MCC (sub-account), the API may return resources from both the target account
+ * (customerId) and the manager account. We prefer the one owned by customerId; if none exists
+ * we fall back to the first (MCC-owned) entry.
+ */
+const buildOwnerPreferredMap = (results, resultKey, customerId) => {
+  const map = {};
+  for (const resultItem of results) {
+    const resource = resultItem[resultKey];
+    if (resource?.name && resource?.resourceName) {
+      const ownedByCustomer = resource.ownerCustomer?.endsWith(`/${customerId}`);
+      if (!map[resource.name] || ownedByCustomer) {
+        map[resource.name] = resource.resourceName;
+      }
+    }
+  }
+  return map;
+};
+
+/**
  * Batch fetch multiple conversion actions in a single API call
  * Returns a map of conversion names to resource names (does not cache - caller should handle caching)
  * @param {string} customerId - The customer ID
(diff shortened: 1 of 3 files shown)


src/v0/destinations/google_adwords_offline_conversions/utils.js
```diff
@@ -648,11 +649,16 @@ const batchFetchConversionActions = async ({ Config, customerId, conversionNames
...
-  // Build result map from API response
+  // When using MCC (sub-account), the API may return conversion actions from both the
+  // target account (customerId) and the manager account (loginCustomerId).
+  // We prefer the action owned by customerId since conversions are uploaded to that account.
   for (const resultItem of results) {
     const { conversionAction } = resultItem;
     if (conversionAction?.name && conversionAction?.resourceName) {
-      conversionMap[conversionAction.name] = conversionAction.resourceName;
+      const ownedByCustomer = conversionAction.ownerCustomer?.endsWith(`/${customerId}`);
+      if (!conversionMap[conversionAction.name] || ownedByCustomer) {
```
achettyiitr: Do we still needs `!conversionMap[conversionAction.name]` check, since we are filtering based on ownedByCustomer, we should expect duplicates?
```suggestion
      if (ownedByCustomer) {
```

shekhar-rudder: This is meant to handle the scenario where the action exists only in the parent account and not in the customer account. In that case `ownedByCustomer` will never be `true`. I’m not sure whether this is practically possible or not.

maheshkutty:
@shekhar-rudder we also need to add a filter for custom variables.
https://github.com/rudderlabs/rudder-transformer/blob/int-6206-google-conversion-id/src/v0/destinations/google_adwords_offline_conversions/utils.js#L746

shekhar-rudder:
@maheshkutty Done — added `conversion_custom_variable.owner_customer` to the GAQL query and the same MCC owner-preference logic in `batchFetchConversionCustomVariablesMap`. Updated network mocks too. See e40580d.

maheshkutty (approved):

maheshkutty:
@shekhar-rudder As an enhancement task, we can add a status filter also to conversion_actions and conversions_custom_variabale

achettyiitr (approved):

PR #5132, opened 2026-04-13
fix: add validation to posthog to fail events more than 1mb

ItsSudip:

Add validation to discard events with a size of more than 1mb. Though in the doc it is mentioned 1MB, but they are also failing with the events with ~1MB, so we kept the limit to 900000 bytes to have some headroom.

Resolves INT-6222

N/A

N/A

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist

### DIFF
--- a/src/v0/destinations/posthog/config.ts
+++ b/src/v0/destinations/posthog/config.ts
@@ -42,4 +42,8 @@ const PROPERTY = {
 
 const MAPPING_CONFIG = getMappingConfig({ ...CONFIG_CATEGORIES, PROPERTY }, __dirname);
 
-export { DEFAULT_BASE_ENDPOINT, CONFIG_CATEGORIES, MAPPING_CONFIG, PROPERTY };
+// PostHog's Kafka producer defaults to message.max.bytes = 1,000,000. To keep headroom of 100kb we kept the limit to 900_000
+// https://posthog.com/docs/data/ingestion-warnings
+const MAX_EVENT_SIZE_BYTES = 900_000;
+
+export { DEFAULT_BASE_ENDPOINT, CONFIG_CATEGORIES, MAPPING_CONFIG, PROPERTY, MAX_EVENT_SIZE_BYTES };
--- a/src/v0/destinations/posthog/config.ts
+++ b/src/v0/destinations/posthog/config.ts
@@ -43,9 +43,8 @@ const PROPERTY = {
 const MAPPING_CONFIG = getMappingConfig({ ...CONFIG_CATEGORIES, PROPERTY }, __dirname);
 
 // PostHog documents a 1 MB event size limit, but through trial and error we found
-// that events under 1 MB are also rejected. PostHog's Kafka producer defaults to
-// message.max.bytes = 1,000,000 and Kafka adds framing overhead on top of the payload.
-// To keep headroom we set the limit to 900,000 bytes (~100 KB buffer).
+// that events under 1 MB are also rejected. To keep headroom we set the limit to
+// 900,000 bytes (~100 KB buffer).
 // https://posthog.com/docs/data/ingestion-warnings
 const MAX_EVENT_SIZE_BYTES = 900_000;
 
--- a/src/v0/destinations/posthog/config.ts
+++ b/src/v0/destinations/posthog/config.ts
@@ -42,7 +42,10 @@ const PROPERTY = {
 
 const MAPPING_CONFIG = getMappingConfig({ ...CONFIG_CATEGORIES, PROPERTY }, __dirname);
 
-// PostHog's Kafka producer defaults to message.max.bytes = 1,000,000. To keep headroom of 100kb we kept the limit to 900_000
+// PostHog documents a 1 MB event size limit, but through trial and error we found
+// that events under 1 MB are also rejected. PostHog's Kafka producer defaults to
+// message.max.bytes = 1,000,000 and Kafka adds framing overhead on top of the payload.
+// To keep headroom we set the limit to 900,000 bytes (~100 KB buffer).
 // https://posthog.com/docs/data/ingestion-warnings
 const MAX_EVENT_SIZE_BYTES = 900_000;
 
--- a/src/v0/destinations/posthog/routerTransform.ts
+++ b/src/v0/destinations/posthog/routerTransform.ts
@@ -1,4 +1,5 @@
 import { z, ZodType } from 'zod';
+import { InstrumentationError } from '@rudderstack/integrations-lib';
 import {
   BatchDestination,
   TransformedEvent,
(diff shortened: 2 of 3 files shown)


src/v0/destinations/posthog/routerTransform.ts
```diff
@@ -7,13 +8,22 @@ import {
...
     const result = processEvent(input.message, input.destination);
     // Strip api_key from the body — it belongs in the wrapBody wrapper
     // eslint-disable-next-line @typescript-eslint/naming-convention
     const { api_key, ...eventBody } = result.body.JSON!;
+
+    const eventSize = Buffer.byteLength(JSON.stringify(eventBody), 'utf8');
+    if (eventSize > MAX_EVENT_SIZE_BYTES) {
+      throw new InstrumentationError(
+        `Event size (${eventSize} bytes) exceeds PostHog's 1 MB limit. PostHog will reject this event with "maximum event size exceeded". Reduce the properties payload.`,
+      );
```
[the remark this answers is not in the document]
ItsSudip: Actually we came to know that from trial and error method that posthog is discarding events less than 1mb as well. So kept the limit to 900000 bytes. But to have consistency with the doc we kept it 1mb.

src/v0/destinations/posthog/config.ts
```diff
@@ -42,4 +42,8 @@ const PROPERTY = {
 
 const MAPPING_CONFIG = getMappingConfig({ ...CONFIG_CATEGORIES, PROPERTY }, __dirname);
 
-export { DEFAULT_BASE_ENDPOINT, CONFIG_CATEGORIES, MAPPING_CONFIG, PROPERTY };
+// PostHog's Kafka producer defaults to message.max.bytes = 1,000,000. To keep headroom of 100kb we kept the limit to 900_000
```
shekhar-rudder: > Actually we came to know that from trial and error method that posthog is discarding events less than 1mb as well. So kept the limit to 900000 bytes. But to have consistency with the doc we kept it 1mb.

This is useful context. We should add it as comment

src/v0/destinations/posthog/routerTransform.ts
```diff
@@ -7,13 +8,22 @@ import {
...
   transformEvent(input: PostHogRouterRequest): TransformedEvent<PostHogPayload> {
     const result = processEvent(input.message, input.destination);
     // Strip api_key from the body — it belongs in the wrapBody wrapper
     // eslint-disable-next-line @typescript-eslint/naming-convention
     const { api_key, ...eventBody } = result.body.JSON!;
+
+    const eventSize = Buffer.byteLength(JSON.stringify(eventBody), 'utf8');
+    if (eventSize > MAX_EVENT_SIZE_BYTES) {
+      throw new InstrumentationError(
+        `Event size (${eventSize} bytes) exceeds PostHog's 1 MB limit. PostHog will reject this event with "maximum event size exceeded". Reduce the properties payload.`,
```
shekhar-rudder: Reduce the properties payload -> Sounds incomplete

src/v0/destinations/posthog/routerTransform.ts
```diff
@@ -7,13 +8,22 @@ import {
...
+import { MAX_EVENT_SIZE_BYTES } from './config';
 
 class PostHogIntegration extends BatchDestination<PostHogPayload> {
   transformEvent(input: PostHogRouterRequest): TransformedEvent<PostHogPayload> {
     const result = processEvent(input.message, input.destination);
     // Strip api_key from the body — it belongs in the wrapBody wrapper
     // eslint-disable-next-line @typescript-eslint/naming-convention
     const { api_key, ...eventBody } = result.body.JSON!;
+
+    const eventSize = Buffer.byteLength(JSON.stringify(eventBody), 'utf8');
```
shekhar-rudder, line 20: Instead of us failing the event, should we handle it in proxy?

ItsSudip: It is already documented why we would handle it in proxy?

shekhar-rudder: As discussed earlier, introducing integration-specific validation on our side could lead to duplication. If PostHog updates their limits in the future, we would need to revisit and update our implementation accordingly.

ItsSudip: That can also happen with the overall limit of 10mb.

ItsSudip: Adding transformerProxy will increase processing time. I don't think that for such documented limits, we should add transformerProxy. If Posthog updates its limit, we can increase that as well.

shekhar-rudder: Yeah, but that’s something we’ll need to handle on our side. If we send everything in one batch and it crosses the 10 MB limit, we’ll have to retry those events (maybe one by one) since they’re all valid. In the 1 MB case, the event itself is invalid anyway, so it doesn’t really matter whether we reject it or PostHog does.

ItsSudip: It is the same, right? We are preparing a batch considering the invalid event, and through the proxy, we need to retry the events one by one.

shekhar-rudder: > Adding transformerProxy will increase processing time.

This is a valid point

> We are preparing a batch considering the invalid event, and through the proxy, we need to retry the events one by one.

Based on yesterday’s conversation, I was under the impression that PostHog’s batch API returns which events succeeded and which failed. Based on that, we could simply mark the events exceeding 1 MB as failed. No need to retry one by one.

ItsSudip: I'm sorry for not telling you clearly, but PostHog is returning the following response.
`maximum event size exceeded: Event rejected by kafka during send`.
From this, it is not understandable which event succeeded and which failed.

shekhar-rudder: Ok then this looks good

src/v0/destinations/posthog/config.ts
```diff
@@ -42,4 +42,11 @@ const PROPERTY = {
 
 const MAPPING_CONFIG = getMappingConfig({ ...CONFIG_CATEGORIES, PROPERTY }, __dirname);
 
-export { DEFAULT_BASE_ENDPOINT, CONFIG_CATEGORIES, MAPPING_CONFIG, PROPERTY };
+// PostHog documents a 1 MB event size limit, but through trial and error we found
+// that events under 1 MB are also rejected. PostHog's Kafka producer defaults to
```
shekhar-rudder: This part seems unnecessary - PostHog's Kafka producer defaults to // message.max.bytes = 1,000,000 and Kafka adds framing overhead on top of the payload.

shekhar-rudder (dismissed):

shekhar-rudder (approved):

PR #5133, opened 2026-04-13
fix(fb_custom_audience): add accessToken config validation

shekhar-rudder:
## Summary
- Adds `accessToken` config validation for `fb_custom_audience` at the two entry points: `processRouterDest` (router) and `processEvent` (processor)
- Throws `ConfigurationError` with a clear message when `accessToken` is empty, instead of failing with a vague API error downstream
- Adds test cases for both processor and router paths covering the missing accessToken scenario

## Test plan
- [ ] Existing `fb_custom_audience` tests pass
- [ ] New processor test: empty `accessToken` returns 400 with `ConfigurationError`
- [ ] New router test: rETL record V2 with empty `accessToken` returns 400 with `ConfigurationError`

### DIFF
--- a/src/v0/destinations/fb_custom_audience/transform.ts
+++ b/src/v0/destinations/fb_custom_audience/transform.ts
@@ -251,6 +251,11 @@ const process = (event: {
 }) => processEvent(event.message, event.destination, event.metadata?.workspaceId as string);
 
 const processRouterDest = async (inputs: FbRecordEvent[], reqMetadata: unknown) => {
+  const { accessToken } = inputs[0].destination.Config;
+  if (!isDefinedAndNotNullAndNotEmpty(accessToken)) {
+    throw new ConfigurationError('Missing required configuration field: accessToken');
+  }
+
   const respList: unknown[] = [];
   const groupedInputs = await groupByInBatches(inputs, (input) =>
     (input.message.type ?? '').toLowerCase(),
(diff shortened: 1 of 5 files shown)


src/v0/destinations/fb_custom_audience/transform.ts
```diff
@@ -182,6 +182,10 @@ const processEvent = (
     throw new InstrumentationError(` ${message.type} call is not supported `);
   }
 
+  if (!isDefinedAndNotNullAndNotEmpty(accessToken)) {
+    throw new ConfigurationError('Missing required configuration field: accessToken');
```
achettyiitr: We can align the error message, similar to what we have for audience ID?
```suggestion
    throw new ConfigurationError('Access Token is a mandatory field');
```

src/v0/destinations/fb_custom_audience/transform.ts
```diff
@@ -247,6 +251,11 @@ const process = (event: {
 }) => processEvent(event.message, event.destination, event.metadata?.workspaceId as string);
 
 const processRouterDest = async (inputs: FbRecordEvent[], reqMetadata: unknown) => {
+  const { accessToken } = inputs[0].destination.Config;
+  if (!isDefinedAndNotNullAndNotEmpty(accessToken)) {
```
achettyiitr, line 255: Same as above comment.

src/v0/destinations/fb_custom_audience/transform.ts
shekhar-rudder: Done — updated the error message to `'Access Token is a mandatory field'` to align with the audienceId pattern.

src/v0/destinations/fb_custom_audience/transform.ts
shekhar-rudder: Done — updated here as well to match.

achettyiitr (approved):

maheshkutty (approved):

PR #5158, opened 2026-04-22
fix: intercom search api

achettyiitr:

- Refactored `searchContact` in `src/cdk/v2/destinations/intercom/utils.js` to simplify lookup construction.
- Added support for multi-lookup Intercom search queries (`email` + `external_id`) using `OR`.
- Preserved rETL mapped flow behavior to search only by `external_id` when email is absent.
- Kept strict lookup value validation and early-return behavior for invalid/missing lookups.

Resolves INT-6260


src/cdk/v2/destinations/intercom/utils.js
```diff
@@ -276,42 +276,67 @@ const filterCustomAttributes = (payload, type, destination, message) => {
...
+    const firstLookUpField = getLookUpField(message);
+    let firstLookUpFieldValue = getLookupValue(firstLookUpField);
+
+    if (firstLookUpField === 'email' && firstLookUpFieldValue) {
+      firstLookUpFieldValue = firstLookUpFieldValue.toLowerCase();
+    }
+    addLookup(firstLookUpField, firstLookUpFieldValue);
+
+    if (firstLookUpField !== 'external_id') {
+      addLookup('external_id', getLookupValue('external_id'));
```
ItsSudip: We need to call getLookUpValue with userId or anonymousId right? As we don't have external_id on the input payload.

src/cdk/v2/destinations/intercom/utils.js
```diff
@@ -276,42 +276,67 @@ const filterCustomAttributes = (payload, type, destination, message) => {
...
+  );
+  if (invalidLookups.length > 0) {
+    warn('[INTERCOM] Invalid lookups found in the message', {
+      invalidLookups,
+    });
+    return null;
+  }
+  if (lookups.length === 0) {
+    warn('[INTERCOM] No lookups found in the message', {
+      message,
```
ItsSudip: This test can go up before applying lookups.filter.

src/cdk/v2/destinations/intercom/utils.js
```diff
@@ -335,6 +360,12 @@ const searchContact = async (message, destination, metadata) => {
   );
   const processedUserResponse = processAxiosResponse(response);
   if (isHttpStatusSuccess(processedUserResponse.status)) {
+    if (processedUserResponse.response?.data.length > 1) {
+      warn('[INTERCOM] Multiple contacts found for the message', {
+        message,
+        contacts: processedUserResponse.response?.data,
+      });
+    }
```
ItsSudip, line 376: Let's add one metric to get the count of such scenarios.

ItsSudip (changes requested):

src/cdk/v2/destinations/intercom/utils.js
achettyiitr: No need for metrics. We can count based on logs if needed.

ItsSudip: Metrics will be faster to query. We can remove logs as well and keep the metric only. WDYT?

achettyiitr: Log-based metrics are better as we can have all different sets of information, we can't have those for metrics. Also, since the frequency is expected to be very low, I don't think fastness matters here.

achettyiitr: I mean, if you need contact Id, you can't keep those in metrics right.

ItsSudip: I missed what we are logging. We shouldn't log the entire data it can contain pii data like phone and email.

achettyiitr: Yeah, missed that. Fixint it.

src/cdk/v2/destinations/intercom/utils.js
```diff
@@ -276,42 +276,68 @@ const filterCustomAttributes = (payload, type, destination, message) => {
...
+  const isRetlExternalIdMapping =
+    message.context?.mappedToDestination && message.traits?.external_id && !message.traits?.email;
+  if (isRetlExternalIdMapping) {
+    addLookup('external_id', message.traits.external_id);
+  } else {
+    const lookupField = getLookUpField(message);
+    let lookupFieldValue = getFieldValueFromMessage(message, lookupField) || message?.context?.traits?.[lookupField];
+
+    if (lookupField === 'email' && lookupFieldValue) {
+      lookupFieldValue = lookupFieldValue.toLowerCase();
```
maheshkutty, line 303: This looks like behavior changes to me for search contact @achettyiitr are you sure for email we need to do a lowercase

achettyiitr: Yeah, INTERCOM stores email as lowercase. While doing the search API, if we search with mixed case, it doesn't work.

src/cdk/v2/destinations/intercom/utils.js
```diff
@@ -335,6 +361,14 @@ const searchContact = async (message, destination, metadata) => {
   );
   const processedUserResponse = processAxiosResponse(response);
   if (isHttpStatusSuccess(processedUserResponse.status)) {
+    if (processedUserResponse.response?.data.length > 1) {
```
maheshkutty, line 370: In case of more than 1 match, can we give preference like  
lookupField + externald high
only email medium
only externalId low

because we do not know in which order search api will give us a result

achettyiitr: No, we can't give preference. And, we don't expect multiple results. In case of multiple results we are logging it.

src/cdk/v2/destinations/intercom/utils.js
maheshkutty: If email other than string it will fails

achettyiitr: Fixed it.

src/cdk/v2/destinations/intercom/utils.js
maheshkutty: Another scenario can be if lookupField itself is externalId.

achettyiitr: Do you expect to get multiple results if lookupField itself is externalId?

maheshkutty: I am not sure you can check it once

ItsSudip: If lookUpField itself is externalId, it can return a maximum of 2 user details.

src/cdk/v2/destinations/intercom/utils.js
```diff
@@ -335,6 +367,14 @@ const searchContact = async (message, destination, metadata) => {
...
+      warn('[INTERCOM] Multiple contacts found for the message', {
+        sourceId: metadata.sourceId,
+        destinationId: metadata.destinationId,
+        workspaceId: metadata.workspaceId,
+        lookups,
```
ItsSudip: `lookups` can contain email, which is PII. Let's remove this.

test/integrations/destinations/intercom/router/data.ts
```diff
@@ -541,172 +541,6 @@ const routerRequest2: RouterTransformationRequest = {
 };
 
 const routerRequest3: RouterTransformationRequest = {
-  input: [
```
ItsSudip, line 544: Why is the entire test being removed?

achettyiitr: This test was not being used.

src/cdk/v2/destinations/intercom/utils.js
achettyiitr: It would be hard to debug the issue, but for now removing it.

ItsSudip (approved):

PR #5167, opened 2026-04-25
fix: add handler to detect silent failures on hubspot

ItsSudip:

When a batch payload sent to a HubSpot batch endpoint is malformed (or otherwise rejected without proper error reporting), HubSpot returns **2xx** with `results: []` and `errors: []` — a silent no-op where nothing is actually written, but our handler previously marked all events as successfully delivered.

This change adds defensive detection for that pattern in the HubSpot network handler.

Resolves INT-6253

### Affected endpoints

Detection scoped to **new API v3 batch endpoints** (URLs containing `/crm/v3/` and `/batch/`):

- `POST /crm/v3/objects/contacts/batch/upsert`
- `POST /crm/v3/objects/contacts/batch/create`
- `POST /crm/v3/objects/contacts/batch/update`
- `POST /crm/v3/objects/:objectType/batch/create` (rETL)
- `POST /crm/v3/objects/:objectType/batch/update` (rETL)
- `POST /crm/v3/associations/:from/:to/batch/create` (rETL)

Excluded (empty body is expected by design):
- `POST /events/v3/send` (track) — 204 No Content
- `POST /contacts/v1/contact/batch/` (legacy batch) — 202 Accepted
- `track.hubspot.com/v1/event` (legacy track) — 200 with empty body

### Detection logic

```
If status is 2xx (including 207) AND endpoint matches /crm/v3/.../batch/...:
  If response.results is empty AND response.errors is empty:
    → SILENT FAILURE → mark all events as 400 (Aborted)
```

When `errors` are present, the existing `handle207MultiStatus` path runs and preserves specific HubSpot error messages (e.g., `"Property value \"invalid-email\" is not a valid email"`). The silent-failure check fires only when both arrays are empty.

### Why 400 (not 500/dontBatch retry)

Retrying with the same payload would produce the same silent no-op. Marking as 400 surfaces the issue in delivery error logs for operator visibility.

Previously, the network handler trusted any 2xx response (including 207) as successful delivery, even when the response body indicated no records were processed. With this change, batch endpoint responses with empty `results` and `errors` are now flagged as failures.

N/A

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist


src/v1/destinations/hs/networkHandler.ts
```diff
@@ -88,6 +88,45 @@ type UpsertResponse = {
   errors?: UpsertError[];
 };
 
+const SILENT_FAILURE_ERROR_MESSAGE =
```
achettyiitr: It can be inline. It's being used in only 1 places, do we still need to declare it as constant?

src/v1/destinations/hs/networkHandler.ts
```diff
@@ -88,6 +88,45 @@ type UpsertResponse = {
...
+};
+
+const buildSilentFailureResponse = (
+  rudderJobMetadata: ProxyMetdata[],
+  status: number,
+): DeliveryV1Response => ({
+  status,
+  message: '[HUBSPOT Response V1 Handler] - Silent failure detected',
+  response: rudderJobMetadata.map((metadata) => ({
+    statusCode: 400,
```
achettyiitr:
```suggestion
    statusCode: HTTP_STATUS_CODES.BAD_REQUEST,
```

src/v1/destinations/hs/networkHandler.ts
```diff
@@ -88,6 +88,45 @@ type UpsertResponse = {
...
 };
 
+const SILENT_FAILURE_ERROR_MESSAGE =
+  '[HUBSPOT] Silent failure: HubSpot returned 2xx but the response indicates no records were processed (empty results and errors).';
+
+// Only new API v3 batch endpoints (e.g., /crm/v3/objects/contacts/batch/upsert)
+// require the silent-failure check. Legacy /contacts/v1/contact/batch/ returns
+// empty body by design (202 Accepted) and must be excluded.
+const isBatchEndpoint = (endpoint?: string): boolean =>
+  typeof endpoint === 'string' && endpoint.includes('/crm/v3/') && endpoint.includes('/batch/');
```
achettyiitr: Any URL containing both `/crm/v3/` and `/batch/` matches. That includes batch read (and potentially other batch verbs) where an empty results array can be legitimate. If those routes ever go through this handler with a JSON body that normalizes to empty arrays, you could false-positive into silent failure.
Can we be more deterministic here on what our batch endpoints are?

src/v1/destinations/hs/networkHandler.ts
```diff
@@ -88,6 +88,45 @@ type UpsertResponse = {
   errors?: UpsertError[];
 };
 
+const SILENT_FAILURE_ERROR_MESSAGE =
+  '[HUBSPOT] Silent failure: HubSpot returned 2xx but the response indicates no records were processed (empty results and errors).';
+
+// Only new API v3 batch endpoints (e.g., /crm/v3/objects/contacts/batch/upsert)
+// require the silent-failure check. Legacy /contacts/v1/contact/batch/ returns
+// empty body by design (202 Accepted) and must be excluded.
+const isBatchEndpoint = (endpoint?: string): boolean =>
```
achettyiitr: It's misleading from the function name that we are not excluding legacy endpoints. Can we have a more appropriate naming?
```suggestion
const isNewBatchEndpoint = (endpoint?: string): boolean =>
```

src/v1/destinations/hs/networkHandler.ts
```diff
@@ -88,6 +88,45 @@ type UpsertResponse = {
   errors?: UpsertError[];
 };
 
+const SILENT_FAILURE_ERROR_MESSAGE =
+  '[HUBSPOT] Silent failure: HubSpot returned 2xx but the response indicates no records were processed (empty results and errors).';
+
+// Only new API v3 batch endpoints (e.g., /crm/v3/objects/contacts/batch/upsert)
+// require the silent-failure check. Legacy /contacts/v1/contact/batch/ returns
+// empty body by design (202 Accepted) and must be excluded.
```
achettyiitr: We can move this comment to where it is being used. Having a generic function with a reference for silent failure is misleading.

src/v1/destinations/hs/networkHandler.ts
```diff
@@ -156,6 +195,14 @@ const responseHandler = (responseParams: {
   const responseWithIndividualEvents: DeliveryJobState[] = [];
   const { response, status } = destinationResponse;
 
+  // Detect silent failures on batch endpoints: HubSpot returned 2xx but the
+  // response indicates no records were processed (empty results and errors).
+  // Mark all events as 400 since retrying with the same payload would produce
+  // the same silent no-op.
+  if (isHttpStatusSuccess(status) && isSilentFailure(response, destinationRequest?.endpoint)) {
```
achettyiitr: Will destinationRequest be nil here? `destinationRequest?.endpoint`

src/v1/destinations/hs/networkHandler.ts
achettyiitr: We already defined endpoint as a string, do we still need a type check here? Can't we do something like this?
```suggestion
const isBatchEndpoint = (endpoint?: string): boolean => {
  if (!endpoint) {
    return false;
  }
  return endpoint.includes('/crm/v3/') && endpoint.includes('/batch/');
}
```

test/integrations/destinations/hs/dataDelivery/silentFailure.ts
```diff
@@ -0,0 +1,296 @@
+import { authHeader1 } from '../maskedSecrets';
+import { generateMetadata, generateProxyV1Payload } from '../../../testUtils';
+
+const SILENT_FAILURE_ERROR =
```
achettyiitr: Use the one which is present in networkHandler.ts

test/integrations/destinations/hs/dataDelivery/silentFailure.ts
```diff
@@ -0,0 +1,296 @@
...
+  '[HUBSPOT] Silent failure: HubSpot returned 2xx but the response indicates no records were processed (empty results and errors).';
+
+const SILENT_FAILURE_MESSAGE = '[HUBSPOT Response V1 Handler] - Silent failure detected';
```
achettyiitr: Move this to networkHandler and reference it here. Also, use it for buildSilentFailureResponse?

src/v1/destinations/hs/networkHandler.ts
ItsSudip: I can't think of any other deterministic way here. Can you suggest any? If we need to use the same handler in reading, we can introduce a different variable to identify if the request is for delivery or pre-transformation lookup.

test/integrations/destinations/hs/dataDelivery/silentFailure.ts
ItsSudip: We shouldn't use any imported variable in tests as it defeats the purpose.

test/integrations/destinations/hs/dataDelivery/silentFailure.ts
ItsSudip: Same as above

test/integrations/destinations/hs/dataDelivery/silentFailure.ts
achettyiitr: Didn't get what purpose you are talking about.

achettyiitr: Instead of duplicating things, we should reference them. As these are constants, they should be reused. Also, in future, if we change the error message, we would need to modify the tests as well, which shouldn't be necessary.

test/integrations/destinations/hs/dataDelivery/silentFailure.ts
achettyiitr: Instead of duplicating things, we should reference them. As these are constants, they should be reused. Also, in future, if we change the error message, we would need to modify the tests as well, which shouldn't be necessary.

test/integrations/destinations/hs/dataDelivery/silentFailure.ts
ItsSudip: In testing we should assert the output. So that if anything get changed in processing that can be caught. But if we refer the same things that will defeats the purpose of testing.

achettyiitr: I don't think we are testing the error messages; we are rather testing the scenario, right?

achettyiitr: If the test is to compare the error message, may be your point make sense. SInce we are not doing that, it doesn't make sense.

src/v1/destinations/hs/networkHandler.ts
achettyiitr: Don't we know the list of endpoints that will end up here? Or is it un-deterministic?

ItsSudip: '/crm/v3/objects/contacts/batch/upsert',
  '/crm/v3/objects/contacts/batch/create',
  '/crm/v3/objects/contacts/batch/update',
  '/crm/v3/objects/:objectType/batch/create',  // matched via regex/replace
  '/crm/v3/objects/:objectType/batch/update',
  '/crm/v3/associations/:from/:to/batch/create',
  These are the endpoints. Should we use these then?

achettyiitr: Yeah, let's have a deterministic list rather than a regex so as to have control. Later, if something changes, we can control it.

ItsSudip: Some endpoints are dynamic. We need to use regex there to compare.

achettyiitr: Ok. If there are dynamic lists, there is no other way. Let's have a regex then.

test/integrations/destinations/hs/dataDelivery/silentFailure.ts
achettyiitr: Does these need to be uppercase? https://www.typescriptlang.org/docs/handbook/variable-declarations.html#const-declarations

achettyiitr: We can move these to networkHandler and use it in both places.

achettyiitr (approved):

PR #5185, opened 2026-05-05
feat(custom_audience): sandbox template parsing in isolated-vm

shekhar-rudder:
## Summary
- Run `parseTemplate` inside per-workspace **isolated-vm** isolates so user-supplied templates cannot affect the host process
- Add `IvmScriptRunner` — a cached isolate runner backed by `DisposableCache` (LRU with stats/logging/auto-disposal)
- Bundle the parser via esbuild into a single IIFE (`sandboxedParse.bundle.js`) loaded into each isolate context
- Make the `parse-template` endpoint async and require `workspaceId` for per-workspace isolate isolation
- Add configurable `cacheName` to `DisposableCache` so different caches emit distinct metric tags
- Fix race condition in `IvmScriptRunner.getOrCreate` via promise coalescing — concurrent calls for the same workspace now share a single in-flight isolate creation instead of each creating their own
- Fix CI workflow to restore `npm run prepare` (isolated-vm native rebuild) before `build:ci` — `.npmrc` has `ignore-scripts=true` so lifecycle hooks don't auto-run

## Test plan
- [x] `npm test -- --testPathPattern="custom_audience" --no-coverage` — unit tests for templateParser + sandbox
- [x] `npm test -- --testPathPattern="ivmScriptRunner" --no-coverage` — promise coalescing race condition tests
- [x] `npm test -- --testPathPattern="ivmCache" --no-coverage` — cacheName and metric tag tests
- [x] `npm test -- --testPathPattern="eventTest" --no-coverage` — controller endpoint tests
- [ ] `npm run build:custom-audience-sandbox` builds the bundle successfully
- [ ] Verify `--no-node-snapshot` in the start script doesn't break normal startup
- [ ] CI passes (`user_transformation*` test suites no longer crash on missing isolated-vm binary)


src/util/ivmCache/index.js
```diff
@@ -10,6 +10,7 @@ class DisposableCache {
   constructor(options = {}) {
     this.maxSize = Number.parseInt(options.maxSize ?? process.env.IVM_CACHE_MAX_SIZE ?? '10', 10);
     this.ttlMs = Number.parseInt(options.ttlMs ?? process.env.IVM_CACHE_TTL_MS ?? '300000', 10);
+    this.cacheName = options.name ?? 'ivm';
```
shekhar-rudder, line 13: To prevent conflict with user transformation metrics

src/controllers/eventTest.ts
```diff
@@ -34,13 +35,13 @@ export class EventTestController {
...
-  public static parseCustomAudienceTemplate(ctx: Context) {
+  public static async parseCustomAudienceTemplate(ctx: Context) {
     const parsed = parseTemplateBodySchema.safeParse(ctx.request.body);
     if (!parsed.success) {
       ctx.status = 400;
       ctx.body = { error: formatZodError(parsed.error) };
       return;
     }
-    ctx.body = parseTemplate(parsed.data.requestBody);
+    ctx.body = await sandboxedParseTemplate(parsed.data.requestBody, parsed.data.workspaceId);
```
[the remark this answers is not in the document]
shekhar-rudder: sandboxedParseTemplate method internally has a try catch block implemented

src/v0/destinations/custom_audience/template/templateParserSandbox.ts
```diff
@@ -0,0 +1,37 @@
...
+ * (dist/sandboxedParse.bundle.js) is produced at build time by esbuild
+ * and contains templateParser + json-template-engine as a single IIFE.
+ */
+import path from 'path';
+import { IvmScriptRunner } from './ivmScriptRunner';
+import logger from '../../../../logger';
+import type { ParseTemplateResult } from './templateParser';
+
+// process.cwd() is always the project root (start script runs without cd).
+const BUNDLE_PATH = path.join(process.cwd(), 'dist', 'sandboxedParse.bundle.js');
```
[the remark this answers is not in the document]
shekhar-rudder: Added comment

.github/workflows/dt-test-and-report-code-coverage.yml
```diff
@@ -57,8 +57,8 @@ jobs:
           node-version-file: '.nvmrc'
           cache: 'npm'
 
-      - name: Install Dependencies
-        run: npm ci && npm run prepare
+      - name: Install Dependencies and Build
+        run: npm ci && npm run prepare && npm run build:ci
```
saikumarrs: Let's move the build operation into its own step for clarity.

src/v0/destinations/custom_audience/template/templateParserSandbox.ts
```diff
@@ -0,0 +1,39 @@
...
+): Promise<ParseTemplateResult> {
+  try {
+    return await runner.execute<ParseTemplateResult>(
+      workspaceId,
+      `parseTemplateInSandbox(${JSON.stringify(template)})`,
+    );
+  } catch (err: unknown) {
+    const message = err instanceof Error ? err.message : String(err);
+    logger.error('Template sandbox error', { workspaceId, error: message });
+    return { valid: false, errors: ['Template parsing is temporarily unavailable.'] };
```
saikumarrs, line 36: Don't we want to include more details like the source `message`? If not raw, at least parse and normalise?

src/v0/destinations/custom_audience/template/sandboxedParse.entry.ts
saikumarrs, line 1: We could build the templateParser as IIFE directly, right? Any challenges?

src/v0/destinations/custom_audience/template/templateParserSandbox.ts
```diff
@@ -0,0 +1,39 @@
...
+  execTimeoutMs: 1_000,
+});
+
+export async function sandboxedParseTemplate(
+  template: string,
+  workspaceId: string,
+): Promise<ParseTemplateResult> {
+  try {
+    return await runner.execute<ParseTemplateResult>(
+      workspaceId,
```
saikumarrs, line 30: While workspace level isolation make sense, please double check to ensure no globals are polluted for subsequent runs.

src/v0/destinations/custom_audience/template/ivmScriptRunner.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  private async createEntry() {
+    const isolate = new ivm.Isolate({ memoryLimit: this.memoryLimitMb });
+    const context = await isolate.createContext();
+    const script = await isolate.compileScript(this.getBundleCode());
+    await script.run(context, { timeout: this.initTimeoutMs });
+    return {
+      isolate,
+      context,
+      destroy: async () => {
```
saikumarrs: Who consumes this `destroy` function? It's wired into `DisposableCache.handleDispose` (which calls `value.destroy()` on eviction/delete), but it only runs **after** the entry is successfully cached.

If `createContext`, `compileScript`, or `script.run` throws partway through `createEntry`, the `ivm.Isolate` has already been constructed but the returned object (with `destroy`) is never produced — so nothing cleans up the partially-initialized isolate. That's a native-memory leak per failed init (timeout on bundle eval, OOM during init, malformed bundle, etc.).

Suggest wrapping the body in a `try/catch` that disposes the isolate on error:

```ts
private async createEntry() {
  const isolate = new ivm.Isolate({ memoryLimit: this.memoryLimitMb });
  try {
    const context = await isolate.createContext();
    const script = await isolate.compileScript(this.getBundleCode());
    await script.run(context, { timeout: this.initTimeoutMs });
    return {
      isolate,
      context,
      destroy: async () => { /* ... existing ... */ },
    };
  } catch (err) {
    try { isolate.dispose(); } catch { /* already disposed */ }
    throw err;
  }
}
```

src/v0/destinations/custom_audience/template/templateParserSandbox.ts
```diff
@@ -0,0 +1,39 @@
...
+ * (dist/sandboxedParse.bundle.js) is produced at build time by esbuild
+ * and contains templateParser + json-template-engine as a single IIFE.
+ */
+import path from 'path';
+import { IvmScriptRunner } from './ivmScriptRunner';
+import logger from '../../../../logger';
+import type { ParseTemplateResult } from './templateParser';
+
+// process.cwd() is the project root in both production (npm start) and test (Jest).
+// __dirname is unsuitable here because it resolves to src/ under Jest (ts-jest) but
```
saikumarrs: `process.cwd()` here is fragile — any `process.chdir()` upstream, a non-root container `WORKDIR`, or a non-standard start command resolves `BUNDLE_PATH` to the wrong location and the first sandbox call fails with `ENOENT`. The freshness check in `templateParserSandbox.test.ts` already uses `__dirname`-relative resolution successfully, so the comment about `__dirname` being "unsuitable" only holds for the test-vs-runtime path divergence — which we can sidestep entirely.

Since the bundle lives at a fixed offset from the compiled `.js`, prefer:

```ts
// dist/v0/destinations/custom_audience/template/templateParserSandbox.js → dist/sandboxedParse.bundle.js
const BUNDLE_PATH = path.resolve(__dirname, '../../../../sandboxedParse.bundle.js');
```

This stays correct regardless of cwd, which matters once we start running this in containers or under process supervisors.

src/v0/destinations/custom_audience/template/ivmScriptRunner.ts
```diff
@@ -0,0 +1,126 @@
...
+   * Execute an expression inside a cached isolate.
+   *
+   * The isolate is looked up (or created) by `cacheKey`. The expression is
+   * compiled as a script and run with `copy: true` so the return value is
+   * transferred to the main thread via structured clone.
+   */
+  async execute<T>(cacheKey: string, expression: string): Promise<T> {
+    const entry = await this.getOrCreate(cacheKey);
+
+    try {
```
saikumarrs, line 68: nit: this `catch` evicts the cached isolate on **any** throw. For the intended cases (timeout, OOM, isolate-disposed) eviction is correct, but `parseTemplateInSandbox` already catches internal parser errors and returns `{ valid: false, ... }` — so the only way to reach this branch should be infra-level failures.

Worth narrowing to ivm-specific errors so a transient user-input-shaped error (if one ever slips through) doesn't churn the workspace's isolate:

```ts
} catch (err: unknown) {
  const message = err instanceof Error ? err.message : String(err);
  if (
    message.includes('Script execution timed out') ||
    message.includes('Isolate was disposed') ||
    message.includes('Array buffer allocation failed')
  ) {
    this.cache.delete(cacheKey);
  }
  throw err;
}
```

Not blocking — current behavior is safe (just slightly wasteful in edge cases).

src/v0/destinations/custom_audience/template/templateParserSandbox.ts
```diff
@@ -0,0 +1,39 @@
...
+// dist/ at runtime, requiring different relative paths for each context.
+const BUNDLE_PATH = path.join(process.cwd(), 'dist', 'sandboxedParse.bundle.js');
+
+const runner = new IvmScriptRunner({
+  bundlePath: BUNDLE_PATH,
+  memoryLimitMb: 8,
+  initTimeoutMs: 5_000,
+  execTimeoutMs: 1_000,
+});
```
saikumarrs, line 22: nit: sandbox config is hardcoded here, but `DisposableCache` already supports env overrides (`IVM_CACHE_MAX_SIZE`, `IVM_CACHE_TTL_MS`). Consider the same pattern for the runner so we can tune in prod without a redeploy if a workspace hits the timeout or memory wall:

```ts
const runner = new IvmScriptRunner({
  bundlePath: BUNDLE_PATH,
  memoryLimitMb: Number.parseInt(
    process.env.CUSTOM_AUDIENCE_SANDBOX_MEMORY_MB ?? '8',
    10,
  ),
  initTimeoutMs: Number.parseInt(
    process.env.CUSTOM_AUDIENCE_SANDBOX_INIT_TIMEOUT_MS ?? '5000',
    10,
  ),
  execTimeoutMs: Number.parseInt(
    process.env.CUSTOM_AUDIENCE_SANDBOX_EXEC_TIMEOUT_MS ?? '1000',
    10,
  ),
});
```

Defaults stay identical to today's behavior; nothing changes unless an env var is set.

shekhar-rudder: I've made this configurable in one of the upcoming PRs. https://github.com/rudderlabs/rudder-transformer/pull/5193/changes#diff-62fdf0f1f248195020c34ab8d7fbb528788446d44db27c1ae2c1218acfe316faR118

.github/workflows/dt-test-and-report-code-coverage.yml
shekhar-rudder: Done — split into separate "Install Dependencies" and "Build" steps.

src/v0/destinations/custom_audience/template/templateParserSandbox.ts
shekhar-rudder: Keeping the generic message. This catch handles infra-level failures (OOM, timeout, disposed isolate) — propagating the raw message would leak internal details to the caller. The error is already logged server-side via `logger.error` with workspaceId for debugging.

src/v0/destinations/custom_audience/template/sandboxedParse.entry.ts
shekhar-rudder: We can't merge `sandboxedParse.entry.ts` into `templateParser.ts` because `templateParser.ts` is imported by regular runtime code (tests, the non-sandboxed endpoint). Adding `globalThis.parseTemplateInSandbox = ...` there would execute that side effect on every import, polluting the main Node.js process globals. The entry file is the boundary between main-process code and code bundled for the isolate.

src/v0/destinations/custom_audience/template/templateParserSandbox.ts
shekhar-rudder: Confirmed — `parseTemplate` is a pure function: each call creates fresh local state (`ValidationContext`, `Set<string>`). The module-level `NODE_HANDLERS` is a read-only lookup map, never mutated. No mutable state persists between calls within the same isolate context.

src/v0/destinations/custom_audience/template/ivmScriptRunner.ts
shekhar-rudder: Good catch — refactored `createEntry` to clean up both `context` and `isolate` on init failure using try/catch. Extracted the cleanup into a shared `releaseIvmResources` helper (same pattern as user transformer's factory destroy).

src/v0/destinations/custom_audience/template/templateParserSandbox.ts
shekhar-rudder: Switched to `path.resolve(__dirname, ...)`. Works under both ts-jest and compiled runtime.

src/v0/destinations/custom_audience/template/ivmScriptRunner.ts
shekhar-rudder: Considered this, but matching on ivm error strings (`'Script execution timed out'`, etc.) is fragile — those messages aren't part of isolated-vm's public API. Current behavior (evict on any throw) is safe and simpler.

src/v0/destinations/custom_audience/template/templateParserSandbox.ts
shekhar-rudder: Addressed in #5193.

saikumarrs (approved):

PR #5187, opened 2026-05-06
feat(custom_audience): router transform pipeline

shekhar-rudder:
## Summary

Adds router transform support for the **Custom Audience** destination, built on the native batching framework.

### Custom Audience integration

Each incoming record event goes through: action lookup → custom-mapping injection → field processing (empty-stripping + optional SHA256/SHA512/MD5 hashing via `processAudienceRecord`). The result is a typed `TransformedEvent` carrying the cleaned record.

Key design choices:
- **Eager endpoint resolution** — endpoints and request headers are computed once per configured action in the constructor, not per event.
- **Action-aware grouping** — a new `internalGroupKey` field keeps insert/update/delete in separate batches even when their (endpoint, method, headers) happen to match. Each action carries its own `requestBody` template, so mixing them would produce wrong payloads.
- **Per-action chunking** — uses `ChunkBatchStrategy` with `maxItems: actionConfig.batchSize` and a `wrapBody` that evaluates the action's `requestBody` json-template against the batched records and connection config.
- **Authentication** — supports noAuth, Basic, Bearer token, and API key header styles, resolved from destination config.
- **Custom mappings** — connection-level `customMappings` inject literal values onto each record before template evaluation. The `from`/`to` naming follows the upstream schema convention (documented with JSDoc since semantics are inverted from the usual source→target pattern).

### Framework extensions

- **Typed generics** on `BatchDestination`, `BatchDestinationConstructor`, and `processBatchedDestination` — added `TConfig` / `TConnectionConfig` type parameters so subclasses get typed `Destination.Config` and `Connection<T>`. Defaults to `Record<string, unknown>` to preserve backward compatibility (PostHog compiles unchanged).
- **Connection injection** — `connection` is extracted from the first event and passed into the constructor by the framework. All events in a single router-transform call share the same connection (rudder-server groups by source+destination), so this is safe.
- **`internalGroupKey`** on `TransformedEvent` — an optional discriminator included in the composite grouping key. Lets destinations add grouping dimensions beyond the user-visible request fields without polluting headers or params.

### Wiring

- Added `CUSTOM_AUDIENCE: true` to `batchedDestinationsMap` and `features.routerTransform` so the dispatcher routes through `processBatchedDestination`.
- Template execution uses `JsonTemplateEngine.createAsSync` in-process — `TODO(INT-6295)` marks the swap point for sandboxed evaluation.

### Known behavior: single-element JSONata array flattening

When a `requestBody` template maps over `$.records` and there is exactly one record, the json-template-engine returns a plain object instead of a single-element array (standard JSONata semantics). This is tested explicitly and template authors should account for it if the downstream API requires an array.

## Test plan
- [ ] `npm run lint` — clean
- [ ] `npm test -- --testPathPattern="custom_audience" --no-coverage` — 29 unit tests pass (21 utils + 8 routerTransform, both table-driven; includes explicit single-element JSONata flattening assertion)
- [ ] `npm test -- --testPathPattern="batchedDestinationsMap|nativeBatching|posthog" --no-coverage` — framework + PostHog regression tests pass with no changes
- [ ] `npm run test:ts -- component --destination=custom_audience` — 5 router scenarios pass: action grouping + chunking, missing-action 400, all-fields-stripped 400, customMappings literal injection, and isHashRequired SHA256 hashing
- [ ] `npm run test:ts -- component --destination=posthog` — PostHog component tests pass (no framework regression)

## Dependencies
- Stacked on top of PR #5185 (`feat/int-6296-isolate-vm-integration-for-template-parsing`). Merge order matters.
- **INT-6295** (Template Evaluator) for sandboxed execution — stubbed with a TODO
- **INT-6290** for the rudder-integrations-config destination definition. That schema currently keys `actions` uppercase; this implementation expects lowercase per audience-destination convention.

### DIFF
--- a/src/services/destination/nativeBatching/processBatchedDestination.ts
+++ b/src/services/destination/nativeBatching/processBatchedDestination.ts
@@ -1,5 +1,5 @@
 import stableStringify from 'fast-json-stable-stringify';
-import type { Destination } from '../../../types/controlPlaneConfig';
+import type { Connection, Destination } from '../../../types/controlPlaneConfig';
 import type { Metadata } from '../../../types/rudderEvents';
 import type {
   RouterTransformationRequestData,
@@ -20,9 +20,13 @@ import { combineBatchRequestsWithSameJobIds } from '../../../v0/util';
 // Validation
 // ---------------------------------------------------------------------------
 
-export function validateInputs<TBody extends Record<string, unknown>>(
+export function validateInputs<
+  TBody extends Record<string, unknown>,
+  TConfig = Record<string, unknown>,
+  TConnectionConfig = Record<string, unknown>,
+>(
   inputs: RouterTransformationRequestData[],
-  integration: BatchDestination<TBody>,
+  integration: BatchDestination<TBody, TConfig, TConnectionConfig>,
 ): { valid: RouterTransformationRequestData[]; errors: (TransformError & { jobId: number })[] } {
   const schema = integration.getInputSchema();
 
@@ -109,6 +113,7 @@ export function groupPayloadsByCompositeKey<
       method: payload.method,
       headers: payload.headers ?? {},
       params: payload.params ?? {},
+      internalGroupKey: payload.internalGroupKey ?? '',
     });
 
     let group = map.get(key);
@@ -190,16 +195,21 @@ function mapErrorPayloadToServerFormat(
 
 export async function processBatchedDestination<
   TBody extends Record<string, unknown> = Record<string, unknown>,
+  TConfig = Record<string, unknown>,
+  TConnectionConfig = Record<string, unknown>,
 >(
   events: RouterTransformationRequestData[],
-  IntegrationClass: BatchDestinationConstructor<TBody>,
+  IntegrationClass: BatchDestinationConstructor<TBody, TConfig, TConnectionConfig>,
   reqMetadata: NonNullable<unknown>,
 ): Promise<RouterTransformationResponse[]> {
   if (events.length === 0) {
     return [];
   }
   const { destination } = events[0];
-  const integration = new IntegrationClass(destination);
+  const connection = events.find((event) => event.connection)?.connection as
+    | Connection<TConnectionConfig>
+    | undefined;
+  const integration = new IntegrationClass(destination as Destination<TConfig>, connection);
   const destType = destination.DestinationDefinition?.Name?.toUpperCase() ?? 'unknown';
   const { workspaceId } = events[0].metadata;
   const metricTags = { destType, workspaceId, destinationId: destination.ID };
--- a/src/v0/destinations/custom_audience/routerTransform.ts
+++ b/src/v0/destinations/custom_audience/routerTransform.ts
@@ -1,0 +1,118 @@
+import { z, ZodType } from 'zod';
+import {
+  BatchDestination,
+  ChunkBatchStrategy,
+  CustomBatchStrategy,
+  TransformedEvent,
+} from '../../../services/destination/nativeBatching/batchDestination';
+import type { BatchStrategy } from '../../../services/destination/nativeBatching/types';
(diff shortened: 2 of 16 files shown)


src/v0/destinations/custom_audience/routerTransform.ts
```diff
@@ -0,0 +1,118 @@
...
+  Action,
+  CustomAudienceConnectionConfig,
+  CustomAudienceDestConfig,
+  CustomAudienceRouterRequest,
+} from './types';
+
+class CustomAudienceIntegration extends BatchDestination<
+  Record<string, string>,
+  CustomAudienceDestConfig,
+  { destination: CustomAudienceConnectionConfig }
```
saikumarrs: Should it be called "CustomAudienceConnectionDestConfig"?

The current name sounds like it represents the whole connection config.

src/v0/destinations/custom_audience/routerTransform.ts
```diff
@@ -0,0 +1,118 @@
...
+    connection?: Connection<{ destination: CustomAudienceConnectionConfig }>,
+  ) {
+    super(destination, connection);
+    this.headers = buildRequestHeaders(destination.Config);
+    this.endpointByAction = Object.fromEntries(
+      Object.entries(destination.Config.actions).map(([action, actionConfig]) => [
+        action,
+        resolveEndpoint(actionConfig!.endpoint, destination.Config.baseUrl, this.connectionConfig),
+      ]),
+    ) as Partial<Record<Action, string>>;
```
saikumarrs: Let's refactor this to a function.

src/v0/destinations/custom_audience/routerTransform.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  private get connectionConfig(): CustomAudienceConnectionConfig {
+    return this.connection!.config.destination;
```
saikumarrs, line 51: We're assuming "connection" will always be available. Is the input contract validated for the presence at the outer layers?

src/v0/destinations/custom_audience/routerTransform.ts
shekhar-rudder: Renamed to `CustomAudienceConnectionDestConfig` across all files — types, routerTransform, utils, and both test files.

src/v0/destinations/custom_audience/routerTransform.ts
shekhar-rudder: Extracted into a private `buildEndpointsByAction()` method. Constructor now calls `this.buildEndpointsByAction()`.

src/v0/destinations/custom_audience/routerTransform.ts
shekhar-rudder: Added an explicit guard in the constructor that throws `InstrumentationError` if `connection` is missing. This is intentionally at the integration level (not the framework) — `connection` is optional in `BatchDestination` since not all destinations use rETL connections, but `custom_audience` requires it for audience ID, hashing config, and custom mappings.

saikumarrs (approved):

src/services/destination/nativeBatching/processBatchedDestination.ts
```diff
@@ -190,16 +195,21 @@ function mapErrorPayloadToServerFormat(
...
   if (events.length === 0) {
     return [];
   }
   const { destination } = events[0];
-  const integration = new IntegrationClass(destination);
+  const connection = events.find((event) => event.connection)?.connection as
+    | Connection<TConnectionConfig>
+    | undefined;
+  const integration = new IntegrationClass(destination as Destination<TConfig>, connection);
   const destType = destination.DestinationDefinition?.Name?.toUpperCase() ?? 'unknown';
```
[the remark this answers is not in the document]
shekhar-rudder: As part of the API contract, all the events in the batch will have the same connection object

PR #5189, opened 2026-05-11
fix: add validation to customerIo for undefined properties

ItsSudip:

Fixes an uncontrolled `TypeError` in the Customer.io router transformer that crashes anonymous SCREEN events when `message.properties` is `undefined`.

The crash occurs at [`src/v0/destinations/customerio/util.js:288`](https://github.com/rudderlabs/rudder-transformer/blob/develop/src/v0/destinations/customerio/util.js#L288), where `message.properties.name` is dereferenced without optional chaining inside the anonymous-event branch of `defaultResponseBuilder`. The equivalent expression in `transform.js:102` already uses `message.properties?.name` — this duplicate in `util.js` was missing the guard.

**Fix**:
```diff
- trimmedEvName = `Viewed ${truncate(message.event || message.properties.name, 86)} Screen`;
+ trimmedEvName = `Viewed ${truncate(message.event || message.properties?.name || '', 86)} Screen`;
```

The trailing `|| ''` is required because `truncate-utf8-bytes` itself throws `Input must be string` when both `message.event` and `message.properties?.name` are absent — without the fallback, optional chaining alone would just swap one Bugsnag alert for another.

**Trigger conditions** (all four must hold):
1. `messageType === 'screen'`
2. No `userId` / `email` (anonymous-event branch)
3. `message.event` is falsy
4. `message.properties` is `undefined`

**Test coverage**: Added a regression unit test in `util.test.js`. Verified that reverting the source fix makes the new test fail with the exact production error, and re-applying the fix makes it pass.

Resolves [INT-6351](https://linear.app/rudderstack/issue/INT-6351)

Eliminate a production Bugsnag alert (`alert group I3YVCG2ZRHG6K`, observed on v1.132.0 against Customer.io PROD) caused by malformed anonymous SCREEN events reaching the router transformer.

No behavioral change for well-formed events. For the previously-crashing edge case (anonymous SCREEN with both `event` and `properties` missing), the transformer now produces a payload with `name: "Viewed  Screen"` instead of throwing — matching the spirit of the analogous guard in `transform.js:102`.

N/A

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist

### DIFF
--- a/src/v0/destinations/customerio/util.js
+++ b/src/v0/destinations/customerio/util.js
@@ -285,7 +285,7 @@ const defaultResponseBuilder = (message, evName, userId, evType, destination, me
     // CustomerIO supports 100byte of event name for anonymous users
     if (messageType === EventType.SCREEN) {
       // 100 - len(`Viewed  Screen`) = 86
-      trimmedEvName = `Viewed ${truncate(message.event || message.properties.name, 86)} Screen`;
+      trimmedEvName = `Viewed ${truncate(message.event || message.properties?.name || '', 86)} Screen`;
     } else {
       if (typeof evName !== 'string') {
         throw new InstrumentationError('Event Name type should be a string');
--- a/src/v0/destinations/customerio/util.test.js
+++ b/src/v0/destinations/customerio/util.test.js
@@ -347,3 +347,32 @@ describe('Unit test cases for customerio defaultResponseBuilder with userId cont
     });
   });
 });
+
+describe('Unit test cases for customerio defaultResponseBuilder anonymous SCREEN edge cases', () => {
+  it('should not crash when properties is undefined for anonymous screen event', () => {
+    const expectedOutput = {
+      rawPayload: {
+        data: undefined,
+        name: 'Viewed  Screen',
+        type: 'event',
+        anonymous_id: 'anon123',
+      },
+      endpointDetails: {
+        endpoint: 'https://track.customer.io/api/v1/events',
+        path: 'v1/events',
+      },
+      requestConfig: { requestFormat: 'JSON', requestMethod: 'POST' },
+    };
+
+    const result = defaultResponseBuilder(
+      { anonymousId: 'anon123', type: 'screen' },
+      'Viewed undefined Screen',
+      undefined,
+      'event',
+      { Config: { apiKey: 'test-api-key', siteID: 'test-site-id' } },
+      'screen',
+    );
+
+    expect(result).toEqual(expectedOutput);
+  });
+});


src/v0/destinations/customerio/util.test.js
```diff
@@ -347,3 +347,32 @@ describe('Unit test cases for customerio defaultResponseBuilder with userId cont
     });
   });
 });
+
+describe('Unit test cases for customerio defaultResponseBuilder anonymous SCREEN edge cases', () => {
+  it('should not crash when properties is undefined for anonymous screen event', () => {
+    const expectedOutput = {
+      rawPayload: {
+        data: undefined,
+        name: 'Viewed  Screen',
```
achettyiitr, line 356: Why is there an extra space here? Can't we just do `Viewed Screen`, even [here](http://github.com/rudderlabs/rudder-transformer/blob/develop/src/v0/destinations/customerio/transform.js#L100-L103) as well.

src/v0/destinations/customerio/util.js
```diff
@@ -285,7 +285,7 @@ const defaultResponseBuilder = (message, evName, userId, evType, destination, me
     // CustomerIO supports 100byte of event name for anonymous users
     if (messageType === EventType.SCREEN) {
       // 100 - len(`Viewed  Screen`) = 86
-      trimmedEvName = `Viewed ${truncate(message.event || message.properties.name, 86)} Screen`;
```
achettyiitr, line 288: Also, Can we have a utility for this, so that it is not segmented in 2 places.

src/v0/destinations/customerio/util.test.js
ItsSudip: In the previous section we are not adding [Here](https://github.com/rudderlabs/rudder-transformer/blob/faf5623912bd3e016442911062586bcdcf5bed85/src/v0/destinations/customerio/transform.js#L100-L102), we are not using truncate, so the event will be `Viewed undefined Screen`. But on the fix we are using truncate so added empty string in place of undefined which is resulting extra space.

achettyiitr: Should we fix the code to send `Viewed Screen` instead of `Viewed  Screen`? Having an extra space in the middle is confusing.

src/v0/destinations/customerio/util.js
ItsSudip: We can't use same utility because in one place we are using truncate and in another place we are not. If we use common util method then there will be behaviour diff.

achettyiitr (approved):

PR #5193, opened 2026-05-11
feat(custom_audience): execute requestBody templates inside isolated-vm

shekhar-rudder:
## Summary

Moves user-supplied `requestBody` template **evaluation** into isolated-vm, completing the sandbox work started in #5185 (which sandboxed only the *parse* step). After this change, untrusted templates supplied via destination config never run in the Node process — both validation and execution happen inside a memory- and time-limited isolate.

### Why isolated-vm for evaluation?

The `requestBody` template is authored by customers via the dashboard and executed per-batch during router transforms. Running it in-process means a malicious or buggy template could access Node globals, block the event loop, or exhaust memory. The parse step was already sandboxed (#5185), but the evaluate step was still running in-process with a TODO comment. This PR closes that gap.

### Key design decisions

**One bundle, one isolate per workspace** — Parse and evaluate share a single esbuild bundle (`dist/sandboxedTemplate.bundle.js`) and a single `IvmScriptRunner` instance. The runner is sized for the heavier evaluate operation (32 MB heap, 500 ms exec timeout, configurable via `CUSTOM_AUDIENCE_IVM_*` env vars). This avoids maintaining two separate bundles/runners for operations that always coexist.

**Structured clone via `evalClosure`** — Data (template, chunks, connection config) is passed into the isolate using `context.evalClosure()` with `{ copy: true }`, which uses V8's structured clone. This avoids `JSON.stringify` embedding in script text, which silently drops `undefined`, converts `NaN`/`Infinity` to `null`, and requires V8 to re-parse serialized text as JS source. Args are scoped to the closure (referenced as `$0`, `$1`, `$2`), avoiding global mutation and race conditions between concurrent calls for the same workspace. This is consistent with how user transformers pass data into isolates via `ExternalCopy`.

**Chunking stays in Node, evaluation goes to the isolate** — The batch strategy chunks payloads by `batchSize` using the existing native-batching infrastructure, then passes all chunks to the isolate in a single call. The isolate evaluates the template once per chunk and returns an array of response bodies. This keeps the chunking logic reusable and limits isolate round-trips (≤ 3 per transform, one per action group).

**Async `BatchStrategy.batch()`** — The isolate call is inherently async, so `BatchStrategy.batch()` now returns `Promise<BatchGroup[]>`. This is a framework-level change that touches `ChunkBatchStrategy`, `CustomBatchStrategy`, and `processBatchedDestination`. Existing sync strategies (e.g. PostHog) are unaffected since returning a value from an async function wraps it in a resolved promise.

**Error categorization** — Template runtime errors (bad syntax, missing field) throw `InstrumentationError` (4xx, user's template is wrong). Isolate infrastructure failures (timeout, OOM, disposed isolate) throw `PlatformError` (5xx, retryable).

### What changed

- **`ivmScriptRunner.ts`** — `execute()` now uses `context.evalClosure()` with `{ copy: true }` (structured clone) instead of `isolate.compileScript()` with `JSON.stringify`-embedded data. Args are passed as `$0, $1, $2` scoped to the closure — no global mutation, no race conditions. Adds a shared singleton `templateSandboxRunner` at module level.

- **`sandboxedTemplate.entry.ts`** (new) — Replaces `sandboxedParse.entry.ts`. Exports two globals on `globalThis`: `parseTemplateInSandbox` (unchanged) and `evaluateTemplateInSandbox` (new). The evaluate function compiles the template via `JsonTemplateEngine`, maps it over all chunks, and returns `{ ok, bodies }` or `{ ok: false, error }`.

- **`templateSandbox.ts`** (new) — Replaces `templateParserSandbox.ts`. Exports `sandboxedParseTemplate` (unchanged behaviour) and `sandboxedEvaluateTemplate` (new). Both use `executeClosure` with positional args.

- **`routerTransform.ts`** — `getBatchStrategy()` switches from `ChunkBatchStrategy` wrapping an in-process `evaluateTemplate` to `CustomBatchStrategy` that chunks payloads via `chunkPayloadsBySize`, then calls `sandboxedEvaluateTemplate` in the isolate.

- **`chunkPayloads.ts`** (new) — Extracts the chunking loop from `ChunkBatchStrategy` into a standalone `chunkPayloadsBySize()` function. `wrapBody` is mandatory — callers that don't use size-based splitting pass a no-op that throws if invoked unexpectedly. `ChunkBatchStrategy` now delegates to it.

- **`utils.ts`** — Removes the in-process `evaluateTemplate` function. `resolveEndpoint` (its only remaining caller) inlines the template engine call directly, with a comment explaining why endpoint resolution intentionally stays in-process (no record data, no user-controlled template path — only `${$.connection.fieldName}` interpolation).

- **`types.ts` / `chunkBatchStrategy.ts` / `customBatchStrategy.ts` / `processBatchedDestination.ts`** — `BatchStrategy.batch()` signature becomes `async → Promise<BatchGroup[]>`. `processBatchedDestination` awaits the call.

- **Tests** — `templateSandbox.test.ts` covers parse (same cases as before) plus evaluate: multi-chunk evaluation, object-iteration templates, `InstrumentationError` on bad templates, `PlatformError` on isolate failure, and structured clone type preservation (undefined, NaN, Infinity). `assertBundleFreshness` is defined once and imported in `routerTransform.test.ts`. PostHog router tests updated to `await` the now-async `batch()`. `batchDestination.test.ts` adds an async `CustomBatchStrategy` test case.

## Test plan
- [x] `npm run build:custom-audience-sandbox` produces up-to-date `dist/sandboxedTemplate.bundle.js`
- [x] `npm run lint` clean
- [x] `npm test -- --testPathPattern="custom_audience|nativeBatching|posthog|eventTest" --no-coverage` — all tests pass
- [x] `npm run test:ts -- component --destination=custom_audience` — 5/5 router integration scenarios pass
- [x] PostHog router (sync `ChunkBatchStrategy`) still works through the now-async `batch()` interface
- [x] Bundle-freshness `beforeAll` guard in `routerTransform.test.ts` and `templateSandbox.test.ts` catches stale bundles in CI
- [x] Structured clone tests verify that `undefined`, `NaN`, `Infinity` survive the isolate boundary


src/v0/destinations/custom_audience/routerTransform.ts
```diff
@@ -95,20 +95,35 @@ class CustomAudienceIntegration extends BatchDestination<
...
-    return new CustomBatchStrategy<Record<string, string>>((payloads) => {
+    return new CustomBatchStrategy<Record<string, string>>(async (payloads) => {
       const action = payloads[0].internalGroupKey as Action;
       const actionConfig = actions[action]!;
-      return new ChunkBatchStrategy<Record<string, string>>({
+
+      // Chunk outside the isolate so the existing native-batching split logic
+      // is reused. Each chunk's records are passed through to the isolate
+      // together with the (untrusted) requestBody template.
+      const chunks = chunkPayloadsBySize(payloads, {
```
saikumarrs: "BySize" is ambiguous. Let's simply call it "chunkPayloads" maybe.

src/v0/destinations/custom_audience/routerTransform.ts
```diff
@@ -95,20 +95,40 @@ class CustomAudienceIntegration extends BatchDestination<
...
+      });
+      const bodies = await sandboxedEvaluateTemplate(
+        actionConfig.requestBody,
+        chunks.map((chunk) => chunk.bodies),
+        connectionConfig,
+        workspaceId,
+      );
+      if (bodies.length !== chunks.length) {
+        throw new InstrumentationError(
+          `Template evaluation returned ${bodies.length} bodies but expected ${chunks.length}`,
```
saikumarrs:
```suggestion
          `Template evaluation returned ${bodies.length} request bodies but expected ${chunks.length}`,
```

src/v0/destinations/custom_audience/template/templateSandbox.ts
```diff
@@ -0,0 +1,70 @@
...
+    return { valid: false, errors: ['Template parsing is temporarily unavailable.'] };
+  }
+}
+
+// ---------------------------------------------------------------------------
+// Evaluate
+// ---------------------------------------------------------------------------
+
+// Must stay in sync with the EvaluateResult definition in sandboxedTemplate.entry.ts.
+type EvaluateResult =
```
saikumarrs: `EvaluateResult` is defined twice — `sandboxedTemplate.entry.ts:25` and here — each carrying a `Must stay in sync` comment. That's drift waiting to happen. Define it once and `import type` it in the other — esbuild erases type-only imports, so the entry bundle stays clean.

src/v0/destinations/custom_audience/routerTransform.ts
saikumarrs: It seems like based on the current chunks processing logic, we might never hit it.
Either we should change the chunks processing logic to error handle individual chunk operation or get rid of this validation.

src/v0/destinations/custom_audience/routerTransform.ts
shekhar-rudder: Done — renamed to `chunkPayloads`.

src/v0/destinations/custom_audience/routerTransform.ts
shekhar-rudder: Applied the wording change.

shekhar-rudder: Keeping this as a post-condition assertion. The sandbox evaluates user-supplied templates inside an isolate — this check enforces the 1:1 contract at the boundary where data crosses back. It's unreachable under normal operation (zero runtime cost) but catches any bug in the sandbox workflow that breaks the expected mapping.

Changed it from `InstrumentationError` to `PlatformError(500)` since a count mismatch would be an internal issue, not a user input problem.

src/v0/destinations/custom_audience/template/templateSandbox.ts
shekhar-rudder: Good catch — exported `EvaluateResult` from `templateSandbox.ts` and replaced the duplicate in `sandboxedTemplate.entry.ts` with `import type`. esbuild erases type-only imports so the bundle stays clean.

saikumarrs (approved):

PR #5196, opened 2026-05-13
fix: remove grouping logic of iterable

ItsSudip:

Removed the `groupEventsByType` call from the Iterable router (`src/v0/destinations/iterable/transform.js`). Events are now batched directly by endpoint bucket inside `filterEventsAndPrepareBatchRequests`, so a single router invocation produces ~2 HTTP calls to Iterable (one `users/bulkUpdate`, one `events/trackBulk`) instead of one call per type-switch per user.

Test data for `routerTest 1` in `test/integrations/destinations/iterable/router/data.ts` updated to reflect the new merged batching (5 batches instead of 7).

Resolves [INT-6362](https://linear.app/rudderstack/issue/INT-6362/iterable-low-throughput-causing-data-backlog) — Iterable: low throughput causing data backlog.

Full investigation summary (problem, root cause, evidence, empirical ordering verification, fix) is in the [Linear comment](https://linear.app/rudderstack/issue/INT-6362#comment-c24b9ea3).

The Iterable workspace was running ~30:1 input/output batching ratio on the router, generating ~30× more HTTP calls than necessary and causing rt-pending-events backlog.

Root cause: `groupEventsByType` deliberately fragments events into many small chunks whenever a user alternates between `identify` and `track`, in order to preserve per-user causal ordering. The intent is that an `identify` lands at Iterable before its dependent `track`.

By removing the call, events batch directly by endpoint bucket. Per-item order within each bulk request is preserved (which is what Iterable honors on a best-effort basis); only the theoretical cross-endpoint sequencing at the transformer is dropped — and that wasn't load-bearing.

Yes — per-user identify→track sequencing across endpoints is no longer enforced at the transformer layer. Reason: empirically verified that Iterable's async ingestion doesn't preserve this order anyway, so the transformer's effort produced no actual benefit and significantly reduced throughput.

Within-endpoint per-item order in the request body is unchanged.

N/A

N/A — `groupEventsByType` in `src/v0/util/index.js` is untouched and other destinations that rely on it are unaffected.

- Expected production impact: input/output batching ratio recovers from ~30:1 to ~2:1 (the natural endpoint-bucket count per router invocation). Backlog drain rate scales by the same factor.
- Existing batch size guards remain in effect: `INITIAL_CHUNK_SIZE = 1000` (`lodash.chunk`) and `MAX_BODY_SIZE_IN_BYTES = 4MB` (size-based split) — neither was the limiting factor before (production batches were at 4–5% of the 4MB cap) and they still cap batch size correctly under the new flow.
- The controller's `groupRouterTransformEvents` (which pre-splits inputs by `destination.ID + sources.job_id + config`) is unchanged, so events from different destinations are still processed in separate router invocations.

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist


shekhar-rudder (commented):
Code change looks good — clean simplification with clear throughput wins. One concern about the test data.

test/integrations/destinations/iterable/router/data.ts
```diff
@@ -847,12 +786,22 @@ export const data: RouterTestData[] = [
                 endpointPath: 'events/trackBulk',
                 headers: {
                   'Content-Type': 'application/json',
-                  api_key: '12345',
+                  api_key: 'dummyApiKey',
                 },
                 params: {},
                 body: {
                   JSON: {
```
shekhar-rudder, line 796: The test inputs have event 4 with `apiKey: '12345'` and `dataCenter: 'USDC'`, but the updated expected output merges it into this `trackBulk` batch using `apiKey: 'dummyApiKey'` (from event 2's destination). This asserts that event 4's data would be sent to the wrong API key.

This happens because `batchTrackEvents` → `lodash.chunk` doesn't group by destination, and `processTrackBatch` uses `chunk[0].destination` for the entire batch.

In production this can't happen — `groupRouterTransformEvents` pre-splits by `stableStringify(destination.Config)`, so events with different configs never share a router invocation. But the test is now asserting incorrect behavior for an impossible scenario.

**Suggestion:** Normalize the test inputs so event 4 uses the same `apiKey: 'dummyApiKey'` and consistent config, matching what production actually delivers. This makes the test a clearer contract of real behavior.

shekhar-rudder (dismissed):

saikumarrs (dismissed):

shekhar-rudder (approved):

PR #5234, opened 2026-05-27
feat(custom_audience): support useInsertConfig toggle for UPDATE action

shekhar-rudder:
## Summary
- Adds `UpdateActionConfig` type with an optional `useInsertConfig` boolean, scoped to the update action slot only
- Extends `lookupActionConfig()` to substitute the insert config when `useInsertConfig` is true on the update action
- Routes `buildEndpointsByAction()` and `getBatchStrategy()` through `lookupActionConfig` so the substitution applies consistently to endpoint, method, requestBody template, batchSize, and fields
- No changes to INSERT or DELETE action processing

## Test plan
- [x] Unit tests in `utils.test.ts`: useInsertConfig true/false/absent, and missing insert config error
- [x] Unit test in `routerTransform.test.ts`: update events with useInsertConfig use insert config's endpoint, method, and template
- [x] Component test fixture (`useInsertConfigDestination`) and test case in `router/data.ts`
- [ ] CI passes

### DIFF
--- a/src/v0/destinations/custom_audience/utils.ts
+++ b/src/v0/destinations/custom_audience/utils.ts
@@ -22,6 +22,14 @@ export const lookupActionConfig = (
   if (!actionConfig) {
     throw new InstrumentationError(ERROR_MESSAGES.NO_ACTION_CONFIG(action));
   }
+  // When the update action opts into reusing the insert config, substitute it.
+  if (action === 'update' && destConfig.actions.update?.useInsertConfig) {
+    const insertConfig = destConfig.actions.insert;
+    if (!insertConfig) {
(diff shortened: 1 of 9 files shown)


src/v0/destinations/custom_audience/utils.ts
```diff
@@ -22,6 +22,14 @@ export const lookupActionConfig = (
   if (!actionConfig) {
```
ItsSudip, line 22: When useInsertConfig is true will then actions[update] will be defined as we need to use the insert action config?

shekhar-rudder: `useInsertConfig` is inside the update action so `actions[update]` will be defined. It is not a top level field

ItsSudip (approved):

PR #5251, opened 2026-06-03
feat: add support of recommended events to braze

ItsSudip:

Adds opt-in support for Braze's recommended ecommerce events (`ecommerce.*`) behind a new `useRecommendedEcommerceEvents` destination flag (default `false`). When enabled, 6 RudderStack ecommerce track events are mapped to Braze recommended event names:

| RudderStack event | Braze event | action |
|---|---|---|
| Product Viewed | `ecommerce.product_viewed` | — |
| Product Added | `ecommerce.cart_updated` | `add` |
| Product Removed | `ecommerce.cart_updated` | `remove` |
| Checkout Started | `ecommerce.checkout_started` | — |
| Order Completed | `ecommerce.order_placed` | — |
| Order Refunded | `ecommerce.order_refunded` | — |
| Order Cancelled | `ecommerce.order_cancelled` | — |

Implemented as a new `ecommerceUtil` module + per-event mapping JSONs, branched into `transform.ts` before the legacy `Order Completed` path.

Resolves INT-4302

Design and rationale: [Braze Ecommerce enhancement LLD](https://www.notion.so/36bf2b415dd0807193e4e7d2c636f702).

Only when `useRecommendedEcommerceEvents` is enabled:

1. `Order Completed` emits `ecommerce.order_placed` in `events[]` instead of the legacy `purchases[]`.
2. Unmapped event-level and per-product properties always flow to `metadata` — not gated by `sendPurchaseEventWithExtraProperties`.

Flag-off behavior is unchanged.

N/A

New Braze-specific module `src/v0/destinations/braze/ecommerceUtil.ts`:

- `getEcommerceMapping(eventName)` — RS event name → `{brazeEvent, action}`.
- `deriveSource(message)` — resolves to `'web' | 'ios' | 'android'`.
- `buildEcommerceEventProperties(...)` — runs `constructPayload` per-event, builds `products[]`, routes metadata, emits the validation counter.

No changes to shared `src/v0/util/`.

- Per-event mapping JSONs load at module init via `getMappingConfig`.
- Validation counter `braze_recommended_event_validation_warn` labels: `{destination_id, workspace_id, braze_event}` — bounded categorical. Fires at most once per event.

## Testing

- Unit: `ecommerceUtil.test.ts` covers mapping correctness, fallback chains, single-product wrapping, source derivation, metadata pass-through, and send-anyway validation.
- Integration: `test/integrations/destinations/braze/processor/data.ts` adds 8 fixtures — one per mapped event plus gap-field scenarios.
- Manual: validated against a local data plane → Braze test workspace via a Postman collection (happy paths, missing-required scenarios, new optional fields).

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist


src/v0/destinations/braze/types.ts
```diff
@@ -175,6 +175,10 @@ export interface BrazeDestinationConfig {
   enableIdentifyForAnonymousUser?: boolean;
   blacklistedEvents?: string[];
   whitelistedEvents?: string[];
+  // When true, RS ecommerce track events are mapped to Braze recommended
+  // `ecommerce.*` events instead of legacy custom/purchase events.
+  // Single toggle gates both cloud and device-mode SDKs.
```
achettyiitr, line 180: Can we use [useEcommerceRecommendedEvents](https://github.com/rudderlabs/rudder-integrations-config/pull/2511).

src/v0/destinations/braze/data/ecommerce/CartUpdated.json
achettyiitr, line 1: Why do we have files prefix with BrazeCartUpdatedConfig? Can we keep CartUpdated?

achettyiitr: Can't we have this kind of format which is more cleaner?
```
src/v0/destinations/braze/data/
  BrazeUserAttributesConfig.json        # legacy, untouched
  ecommerce/
    ProductViewed.json
    CartUpdated.json
    CheckoutStarted.json
    OrderPlaced.json
    OrderRefunded.json
    OrderCancelled.json
    Product.json                        # shared nested product object
```

achettyiitr: config.ts can be mofidified to something like this:

```
// config.ts
BRAZE_CART_UPDATED:     { name: 'ecommerce/CartUpdated' },
BRAZE_PRODUCT_VIEWED:   { name: 'ecommerce/ProductViewed' },
// ...
BRAZE_ECOMMERCE_PRODUCT:{ name: 'ecommerce/Product' },
```

src/v0/destinations/braze/data/ecommerce/CartUpdated.json
```diff
@@ -0,0 +1,32 @@
...
+  {
+    "destKey": "cart_id",
+    "sourceKeys": "properties.cart_id",
+    "brazeRequired": true
+  },
+  {
+    "destKey": "total_value",
+    "sourceKeys": ["properties.total", "properties.value"],
+    "brazeRequired": false
+  },
```
achettyiitr, line 13: For others we have `["properties.total", "properties.revenue", "properties.value"]`, why revenue is not included?

achettyiitr, line 13: Doc says, it conditional?
https://www.braze.com/docs/developer_guide/analytics/logging_ecommerce_events/?tab=cart_updated#event-schemas

src/v0/destinations/braze/data/ecommerce/Product.json
achettyiitr, line 1: Didn't got the schema for this, rest others are present [here](https://www.braze.com/docs/developer_guide/analytics/logging_ecommerce_events/?tab=product_viewed)? Can you share the spec?

src/v0/destinations/braze/transform.ts
```diff
@@ -350,6 +351,35 @@ function processTrackEvent(
     }
   }
 
+  // New recommended-events path. Gated by destination config; falls through to legacy
+  // for unmapped events. Preserves attributes[] + dedup built above.
+  if (messageType === EventType.TRACK && destination.Config.useRecommendedEcommerceEvents) {
+    const ecomMapping = getEcommerceMapping(eventName);
+    if (ecomMapping) {
```
achettyiitr, line 358: It's better to throw an error in case the mapping does not exist rather then fallback to old code?

src/v0/destinations/braze/transform.ts
```diff
@@ -350,6 +351,35 @@ function processTrackEvent(
...
+        ecomMapping.action,
+        destination,
+      );
+      let ecomEvent: Record<string, unknown> = {
+        name: ecomMapping.brazeEvent,
+        time: timestamp,
+        properties: ecomProperties,
+      };
+      ecomEvent = setExternalIdOrAliasObject(ecomEvent, message);
+      ecomEvent = addAppId(ecomEvent, message);
```
achettyiitr, line 372: Didn't saw [spec](https://www.braze.com/docs/developer_guide/analytics/logging_ecommerce_events/?tab=cart_updated&subtab=cart_updated) mentioning anything about external_id, _update_existing_only, user_alias, app_id?

src/v0/destinations/braze/transform.ts
```diff
@@ -350,6 +351,35 @@ function processTrackEvent(
     }
   }
 
+  // New recommended-events path. Gated by destination config; falls through to legacy
+  // for unmapped events. Preserves attributes[] + dedup built above.
+  if (messageType === EventType.TRACK && destination.Config.useRecommendedEcommerceEvents) {
+    const ecomMapping = getEcommerceMapping(eventName);
+    if (ecomMapping) {
+      const { timestamp } = message;
```
achettyiitr, line 359: Do we need to check if timestamp is valid?

src/v0/destinations/braze/transform.ts
```diff
@@ -350,6 +351,35 @@ function processTrackEvent(
...
+      const ecomProperties = buildEcommerceEventProperties(
+        message,
+        ecomMapping.brazeEvent,
+        ecomMapping.action,
+        destination,
+      );
+      let ecomEvent: Record<string, unknown> = {
+        name: ecomMapping.brazeEvent,
+        time: timestamp,
+        properties: ecomProperties,
```
achettyiitr, line 369: `handleReservedProperties` is bypassed on the ecom path. The legacy custom-event path calls handleReservedProperties(properties) before building the payload; the ecom branch returns before reaching it. The mapped recommended-event keys are fixed/safe, but the metadata passthrough is freeform user data routed straight into the event properties. If Braze reserves any event-property names, an unmapped source key could collide unsanitized. → Confirm whether reserved-property handling should apply to metadata, or document why it's intentionally skipped.

src/v0/destinations/braze/ecommerceUtil.ts
```diff
@@ -0,0 +1,388 @@
...
+  const properties = (message.properties || {}) as Record<string, unknown>;
+  const hasExplicitProductsArray = Array.isArray(properties.products);
+
+  // Step 1+2: event-level field mapping.
+  const payload = (constructPayload(message, eventMapping) || {}) as EcommerceEventProperties;
+
+  // Step 3: products[] (skipped for product_viewed — flat, single-product event).
+  if (productMapping !== null) {
+    payload.products = buildProductsArray(message, brazeEvent, productMapping);
+  }
```
achettyiitr, line 435: How are we making sure that for an event with products required, validation is happening since we don't have it in the config?

src/v0/destinations/braze/data/ecommerce/CartUpdated.json
ItsSudip: Yes, `total_value` is required only for `action=replace`. Since this PR only emits `add`/`remove` (`Cart Viewed` / `Cart Updated` are intentionally unmapped per LLD — `cart_updated(replace)` overwrite-on-view was deemed too risky for v1), `total_value` is optional for our case — current config has `brazeRequired: false`, which is correct.

src/v0/destinations/braze/data/ecommerce/Product.json
ItsSudip: The shared per-product schema is embedded in each event's tab on the recommended events doc — same fields across the board: `product_id`, `product_name`, `variant_id`, `quantity`, `price` (required), `image_url`, `product_url`, `metadata` (optional). See for example the [`cart_updated` tab](https://www.braze.com/docs/developer_guide/analytics/logging_ecommerce_events/?tab=cart_updated#event-schemas) under `products[]`.

src/v0/destinations/braze/transform.ts
ItsSudip: The fallback is intentional. Two classes of events hit this path:

1. **Non-ecommerce RS events** (Page View, Login, Identify, etc.) — must continue as generic Braze custom events even when the flag is on.
2. **Intentionally-unmapped ecommerce events** (`Cart Viewed`, `Cart Updated`) — per the LLD decision (`cart_updated(replace)` overwrite-on-view semantics deemed too risky for v1), these continue as custom events under their original RS names.

Throwing would drop both. The flag scopes the new path to mapped events only and leaves everything else untouched.

src/v0/destinations/braze/transform.ts
ItsSudip: `external_id`, `user_alias`, `_update_existing_only`, `app_id` are `/users/track` request-envelope fields per Braze's [data ingestion API](https://www.braze.com/docs/api/endpoints/user_data/post_user_track) — not part of the recommended-event payload itself. The recommended-events spec describes what goes *inside* each `events[]` entry; the envelope wraps the whole batch and is required regardless of event type.

src/v0/destinations/braze/transform.ts
ItsSudip: The legacy custom-event path doesn't validate `message.timestamp` either — it's set by the gateway and assumed valid. Mirroring that here for consistency. If timestamp validation is needed it'd be a cross-destination concern handled at the gateway/dataplane layer.

src/v0/destinations/braze/ecommerceUtil.ts
ItsSudip: `products[]` validation is handled structurally in `emitValidationCounter` (ecommerceUtil.ts), not via the config. For product-bearing events (`productMapping !== null`), it checks whether `payload.products` is missing, empty, or any item has unresolved required fields — the same single per-event counter fires. `products` isn't listed as a config entry because the array's existence is structural to the event type, not a field-mapping concern.

src/v0/destinations/braze/transform.ts
ItsSudip: Good catch — addressed. `handleReservedProperties` now runs on the event-level metadata pass-through inside `buildEcommerceEventProperties`, so reserved keys (`time`, `event_name`) are stripped before they reach `payload.metadata`. Mapped fields aren't affected (none of them are reserved). Per-product `metadata` doesn't go through it since `time`/`event_name` aren't reserved at the product-object level. Added a unit test asserting the scrub.

achettyiitr (approved):

src/v0/destinations/braze/ecommerceUtil.ts
```diff
@@ -0,0 +1,475 @@
...
+// stays narrowed to `'add' | 'remove'`.
+const ECOMMERCE_ROUTED_ENTRIES: readonly EcommerceRoutedEntry[] = Object.values(
+  ConfigCategory,
+).filter((entry): entry is EcommerceRoutedCategory => 'brazeEvent' in entry);
+
+// Case-insensitive RS event name (lowercased, trimmed) → Braze mapping.
+const EVENT_NAME_TO_BRAZE: Record<string, EcommerceMapping> = Object.fromEntries(
+  ECOMMERCE_ROUTED_ENTRIES.flatMap((entry) =>
+    entry.rsEvents.map((rs) => [rs.name, { brazeEvent: entry.brazeEvent, action: rs.action }]),
+  ),
```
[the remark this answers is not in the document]
achettyiitr: Seems like a stale comment. We are already doing lower case when getEcommerceMapping. cc @ItsSudip

src/v0/destinations/braze/data/ecommerce/CartUpdated.json
ItsSudip: Updated

src/v0/destinations/braze/data/ecommerce/ProductViewed.json
```diff
@@ -0,0 +1,50 @@
...
+    "brazeRequired": true
+  },
+  {
+    "destKey": "product_name",
+    "sourceKeys": "properties.name",
+    "type": "String",
+    "brazeRequired": true
+  },
+  {
+    "destKey": "variant_id",
```
krishna2020, line 15: what's difference between product_id & variant_id from braze ? we are using same fields to map them

PR #5253, opened 2026-06-03
fix: handle undefined google ads response

ItsSudip:

Fixes a 500 crash in the Google Ads Enhanced Conversions (GAEC) delivery `responseHandler`.

When the Google Ads API returns a `200` with an empty/undefined body (a fully-successful upload with no partial failures), the handler crashed while destructuring `partialFailureError`:

```
Cannot destructure property 'partialFailureError' of 'destinationResponse.response' as it is undefined.
```

The fix adds a null-guard before destructuring, so a `200` with no body is correctly treated as success:

```js
const { partialFailureError } = destinationResponse?.response || {};
```

Also adds co-located unit tests for `responseHandler` covering the regression and the existing partial-failure paths.

Resolves INT-6475

- Stop GAEC deliveries from failing with a retryable 500 when Google Ads responds `200` with an empty body. A `200` without a `partialFailureError` means the upload fully succeeded, so it is now returned as success instead of throwing.

The only behaviour change is for a `200` response with an undefined/empty body: previously it threw a `TypeError` (reported as a 500), now it returns success. All existing partial-failure (non-zero `code`) and error-status paths are unchanged.

N/A

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist

### DIFF
--- a/src/v0/destinations/google_adwords_enhanced_conversions/networkHandler.js
+++ b/src/v0/destinations/google_adwords_enhanced_conversions/networkHandler.js
@@ -126,7 +126,9 @@ const gaecResponseHandler = (responseParams) => {
   const { status } = destinationResponse;
   if (isHttpStatusSuccess(status)) {
     // for google ads enhance conversions the partialFailureError returns with status 200
-    const { partialFailureError } = destinationResponse.response;
+    // a successful 200 may also have an empty/undefined body (no partial failures),
+    // so guard against destructuring undefined before reading partialFailureError
+    const { partialFailureError } = destinationResponse?.response || {};
     // non-zero code signifies partialFailure
     // Ref - https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto
     if (partialFailureError && partialFailureError.code !== 0) {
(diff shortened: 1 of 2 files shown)


achettyiitr (approved):

src/v0/destinations/google_adwords_enhanced_conversions/networkHandler.js
```diff
@@ -126,7 +126,9 @@ const gaecResponseHandler = (responseParams) => {
   const { status } = destinationResponse;
   if (isHttpStatusSuccess(status)) {
     // for google ads enhance conversions the partialFailureError returns with status 200
-    const { partialFailureError } = destinationResponse.response;
+    // a successful 200 may also have an empty/undefined body (no partial failures),
```
maheshkutty, line 129: @ItsSudip Can you point out in a Google Doc for these comments 
In case of a 200 HTTP status code, we might also get an empty response body
When I was checking, I did not find any document

ItsSudip: I don't have any documentation. But from the bug we could say we received empty response.

PR #5257, opened 2026-06-04
fix: allowlist non-sensitive metadata in error notifier payloads

ItsSudip:

Stops the error notifier (Bugsnag + logger) from leaking the per-event `metadata` — which carries `secret.access_token`, `secret.refresh_token`, and `secret.developer_token` — in error alerts.

A previous fix (#5141) stripped only the **top-level** `metadata` key, but the proxy/delivery v1 failure path (`handlevV1DeliveriesFailureEvents`) nests the per-job metadata inside `resp.response[i].metadata`, so secrets still reached Bugsnag (observed in the GAEC incident, Grafana #1103740).

This PR sanitizes at the **single notifier choke point** (`client.notify`): it recursively walks the payload and reduces every `metadata` object (at any depth) to an **allowlist** of non-sensitive fields — `sourceId`, `destinationId`, `workspaceId` — dropping `secret` and the rest (`jobId`, `userId`, `attemptNum`, `dontBatch`, …).

- New `sanitizeMetadata` util (TypeScript) — recursive, **non-mutating** (the input is the same object reference returned as the delivery response), circular-ref-safe; handles `metadata` as a single object or an array of objects, and drops a non-object/primitive `metadata` value (could itself be a secret) to `{}`.
- `client.notify` now allowlists metadata before delegating to any notifier/logger.
- Removed the now-redundant per-call-site `metadata: undefined` strips in `errorReporting.ts` and `handleRtTfSingleEventError`, centralizing the policy in one place.

Result — the leaked metadata becomes:
```jsonc
"metadata": { "sourceId": "...", "destinationId": "...", "workspaceId": "..." }
```

Resolves INT-6476

- Close the secret-leak gap left by #5141 for the delivery (proxy v1) path, where metadata is nested one level deeper than the top-level strip handled.
- Centralize sanitization at the notifier boundary so no call site can leak — regardless of payload shape — fixing the root cause (scattered, per-call-site strips) rather than patching one more path.
- Allowlist (rather than drop the whole `metadata`) so alerts retain useful, non-sensitive debugging context — `sourceId` / `destinationId` / `workspaceId` — while credentials and noise are removed.
- Covers both leak surfaces: the Bugsnag alert payload **and** the internal `logger.debug`/`logger.error` lines in `bugsnag.js`, since both receive metadata via `client.notify`.

Error alerts/logs now include only the allowlisted metadata fields (`sourceId`, `destinationId`, `workspaceId`) instead of the full per-job metadata. Non-metadata context is unaffected: `status`, `message`, `statTags`, `response[].error`, `response[].statusCode` are retained, and Bugsnag's deploy metadata (`image`/`source`/`cluster`) and `opContext` are added separately so they remain. The data plane does not depend on the echoed-back metadata.

N/A

Adds `src/util/errorNotifier/sanitize.ts` exporting a single public function `sanitizeMetadata(value)` — a non-mutating, recursive, circular-ref-safe metadata allowlister used by the notifier client. The allowlist (`sourceId`, `destinationId`, `workspaceId`) is a module-level constant, and the recursion accumulator is kept private behind an internal helper so the public contract is just `sanitizeMetadata(value)`. Written in TypeScript.

Runs only on the error path (per failed event), over small payloads — negligible overhead.

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist


src/util/errorNotifier/sanitize.ts
```diff
@@ -0,0 +1,95 @@
+// Only these metadata fields are safe to forward to the error notifier (Bugsnag / logger).
+// Everything else on the per-job metadata — notably `secret` (access/refresh/developer
+// tokens), plus jobId/userId/attemptNum/dontBatch — is dropped.
+export const DEFAULT_ALLOWED_METADATA_FIELDS: string[] = [
+  'sourceId',
```
shekhar-rudder: **Suggestion: simplify this module (~95 → ~38 lines), behavior-preserving.**

Two things drive most of the complexity here and neither earns its keep:

1. **The `allowedFields` parameter / `DEFAULT_` prefix.** There's exactly one caller (`client.js`, `sanitizeMetadata(metadata)` with no second arg), and it never overrides the list. The param is exercised only by the `supports a custom allowlist` test — i.e. the test exists *because* the seam exists, not because anything needs it. `DEFAULT_` also implies a non-default case that doesn't exist. Collapse to a plain constant and drop the param; if a second policy ever appears, parameterize *then*.
2. **`pickFields` + `allowlistMetadata` split, plus a dead branch.** They merge into one helper, and the `if (!isRecord(value)) return value` inside `sanitizeMetadata` is unreachable (after the `typeof === 'object'` + not-null + not-array checks it's already a record).

Proposed:

```ts
// Only these metadata fields are safe to forward to the error notifier (Bugsnag / logger).
// Everything else on the per-job metadata — notably `secret` (access/refresh/developer
// tokens), plus jobId/userId/attemptNum/dontBatch — is dropped.
const ALLOWED_METADATA_FIELDS = ['sourceId', 'destinationId', 'workspaceId'];

const isRecord = (value: unknown): value is Record<string, unknown> =>
  typeof value === 'object' && value !== null && !Array.isArray(value);

// Reduce a `metadata` value to the allowlisted fields it actually has. An array is
// reduced element-by-element; a primitive (or null) can't be allowlisted and might
// itself be a secret, so it collapses to {}.
const allowlistMetadata = (value: unknown): unknown => {
  if (Array.isArray(value)) return value.map(allowlistMetadata);
  if (!isRecord(value)) return {};
  return Object.fromEntries(
    ALLOWED_METADATA_FIELDS.filter((field) =>
      Object.prototype.hasOwnProperty.call(value, field),
    ).map((field) => [field, value[field]]),
  );
};

/**
 * Returns a sanitized deep copy of an error payload: wherever a `metadata` key
 * appears (at any depth) its value is reduced to the allowlisted fields. The input
 * is never mutated — it doubles as the live delivery response — and circular
 * references are dropped rather than followed.
 */
export const sanitizeMetadata = (
  value: unknown,
  seen: WeakSet<object> = new WeakSet(),
): unknown => {
  if (value === null || typeof value !== 'object') return value;
  if (seen.has(value)) return undefined; // circular back-edge
  seen.add(value);
  try {
    if (Array.isArray(value)) return value.map((item) => sanitizeMetadata(item, seen));
    return Object.fromEntries(
      Object.entries(value).map(([key, val]) => [
        key,
        key === 'metadata' ? allowlistMetadata(val) : sanitizeMetadata(val, seen),
      ]),
    );
  } finally {
    seen.delete(value);
  }
};
```

Identical on every existing test except `supports a custom allowlist`, which should be removed along with the param. The `seen`/`finally` deep-copy machinery is genuinely irreducible (input is the live delivery response, so we must copy; Bugsnag payloads can self-reference, so we must guard cycles).

Minor nuance, not exercised and not realistic: for `metadata` being an *array of arrays*, the current code maps inner arrays to `{}` while the merged `allowlistMetadata` recurses into them. Metadata is only ever a single object or an array of objects, so this can't occur — flagging it as a conscious call rather than an accident.

shekhar-rudder (commented):
One non-blocking design/simplification suggestion on `sanitize.ts`. The change is correct and well-tested — sanitization is properly centralized at the `client.notify` choke point, genuinely non-mutating (important since the same `resp` is both reported and returned as the delivery response), and all `.notify` paths funnel through it.

shekhar-rudder (commented):
Solid fix — centralizing sanitization at the `client.notify` choke point is the right call, and I confirmed all notifier paths (`reportError` → `client.notify`, `errNotificationClient.notify`) funnel through it. Test coverage is thorough (non-mutation, shared refs, circular refs, array/primitive metadata). One minor API note below.

Note: the PR description still references `DEFAULT_ALLOWED_METADATA_FIELDS` and a `sanitizeMetadata(value, allowedFields = [...])` signature, but the final code hardcodes `ALLOWED_METADATA_FIELDS` and the second param is now `seen` — worth updating the description to match.

src/util/errorNotifier/sanitize.ts
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Returns a sanitized deep copy of an error payload: wherever a `metadata` key
+ * appears (at any depth) its value is reduced to the allowlisted fields. The input
+ * is never mutated — it doubles as the live delivery response — and circular
+ * references are dropped rather than followed.
+ */
+export const sanitizeMetadata = (
+  value: unknown,
+  seen: WeakSet<object> = new WeakSet(),
+): unknown => {
```
shekhar-rudder: `seen` is an internal recursion accumulator but it's exposed as the second public parameter. Consider wrapping the recursion in a private helper so the public contract is just `sanitizeMetadata(value)` — that prevents a caller from accidentally passing a stray second argument and makes the API intent clearer.

shekhar-rudder (approved):

PR #5263, opened 2026-06-05
fix(zoho): prevent COQL injection in legacy deletion lookup

ItsSudip:

Fixes a COQL injection vulnerability in the Zoho **legacy single-event deletion** path. `searchRecordIdV2` built the COQL `WHERE` clause by interpolating both keys and values from `event.message.identifiers` (untrusted input) directly into the query string, with no escaping or allowlisting. The matched record IDs then flow into a `DELETE ...?ids=` request, so a crafted identifier could match unintended records.

Two complementary defenses, mirroring the patterns already used in the safe batched path (`buildBatchedCOQLQueryWithIN`):

- **Value escaping** — added a shared `escapeCOQLStringValue` helper (escapes backslash, then single-quote) applied to string/array values in `generateWhereClause`, so a value cannot break out of its string literal. The batched path's inline copy was refactored to reuse this helper.
- **Key allowlisting** — `generateWhereClause`/`generateSqlQuery` now take an `allowedFields` list; `searchRecordIdV2` passes the configured `identifierMappings` field names. Identifier keys not in the mapping are dropped, so an attacker-controlled key can't be injected into the clause. If all keys drop, the query is empty and the existing "Identifier values are not provided" instrumentation error fires.

Both are required: escaping closes value-side injection; allowlisting closes key-side injection (which escaping alone cannot).

**Tests**
- Unit (`utils.test.ts`): value quote escaped; non-mapped key dropped; mixed payload keeps only allowlisted keys.
- Integration: new regression fixture `router/identifierEscaping.ts` whose network mock matches **only the escaped query** — any escaping regression makes the mock miss and fails the test.

Verified: 62 unit + 28 integration tests pass; lint clean.

Resolves INT-6478

Close the injection on the default deletion codepath with a minimal, low-risk change. Numeric values were already safe (`typeof value === 'number'`). No existing fixtures needed changes — real deletion traffic uses allowlisted identifiers with plain values, so escaping/allowlisting are no-ops for legitimate events.

Identifier keys not present in the destination's configured `identifierMappings` are now dropped from the deletion lookup query (previously they were interpolated as-is). This is a no-op for legitimate traffic, where identifiers are built from that same mapping.

N/A

Added `escapeCOQLStringValue` (module-local to the Zoho destination) to centralize COQL string-literal escaping; reused by both the legacy and batched query builders.

Follow-up (separate ticket): evaluate routing all workspaces through the safe batched path via `DEST_ZOHO_DELETION_BATCHING_SUPPORTED_WORKSPACE_IDS` and deprecating the legacy path. Optional defense-in-depth raised: capping deletions per event and a dropped-key metric.

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist

### DIFF
--- a/src/cdk/v2/destinations/zoho/utils.test.ts
+++ b/src/cdk/v2/destinations/zoho/utils.test.ts
@@ -104,6 +104,46 @@ describe('searchRecordIdV2', () => {
   };
   const mockQuery = "SELECT id FROM Leads WHERE Email = 'test@example.com'";
 
+  // Many-field identifier payload used to exercise the recursive AND grouping. Every key is
+  // mapped as an identifier below so the allowlist keeps them all (see identifierMappings override).
+  const multiRecordIdentifierFields = {
+    Name: 'rid1927ce14265c006ae11555ec6e1cdbbac',
+    Name1: 'Liam Bailey',
+    Has_Chargeback: true,
+    Last_Client_Purchase: '2021-06-15T00:00:00Z',
+    Purchase_Category: ['category1', 'category2'],
+    Lifetime_Client_Revenue: 1200,
+    Name2: 'Olivia Smith',
+    Has_Chargeback2: false,
+    Last_Client_Purchase2: '2022-01-10T00:00:00Z',
+    Purchase_Category2: ['category3', 'category4'],
+    Lifetime_Client_Revenue2: 800,
+    Name3: 'Noah Johnson',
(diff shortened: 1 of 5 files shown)


ItsSudip:
Superseded by #5267, which removes the legacy Zoho deletion path entirely and routes all deletions through the safe batched COQL path (instead of hardening the legacy path). Same ticket: INT-6478.

PR #5273, opened 2026-06-09
chore(klaviyo): migrate destination to TypeScript

shekhar-rudder:
## Summary
- Migrate the entire klaviyo destination (`config`, `util`, `batchUtil`, `transform`, `transformV2`, and the two co-located unit tests) from JavaScript to TypeScript.
- No runtime behavior changes — added types only, with non-null assertions (`!`) on JS function returns where the original code implicitly trusted non-null. Zero `any` annotations, zero `as` casts, zero `@ts-ignore`.
- Refine `memory-bank/23_js_to_ts_migration.md` with lessons from this migration: prefer `!` over if-throw validation, explicitly forbid `Object.assign` as a type-narrowing workaround, and document the one scoped exception (runtime mutation patterns like object-to-array property mutation) where a minimal local-helper is required.

## Test plan
- [x] `npm run build:ci` passes
- [x] `npm test -- --testPathPattern="klaviyo" --no-coverage` — 17/17 unit tests pass
- [x] `npm run test:ts -- component --destination=klaviyo` — 40/40 integration tests pass
- [x] `npx eslint src/v0/destinations/klaviyo/` clean
- [x] `npx prettier --check src/v0/destinations/klaviyo/` clean

### DIFF
--- a/src/v0/destinations/klaviyo/transform.ts
+++ b/src/v0/destinations/klaviyo/transform.ts
@@ -2,11 +2,7 @@
 /* eslint-disable no-underscore-dangle */
 /* eslint-disable  array-callback-return */
 import get from 'get-value';
-import {
-  ConfigurationError,
-  InstrumentationError,
-  TransformationError,
-} from '@rudderstack/integrations-lib';
+import { ConfigurationError, InstrumentationError } from '@rudderstack/integrations-lib';
 import { EventType, WhiteListedTraits, MappedToDestinationKey } from '../../../constants';
 import {
   CONFIG_CATEGORIES,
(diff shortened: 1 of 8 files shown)


src/v0/destinations/klaviyo/transform.ts
```diff
@@ -68,7 +68,7 @@ const identifyRequestHandler = async (
     traitsInfo = addSubscribeFlagToTraits(traitsInfo);
   }
 
-  let propertyPayload = constructPayload(message, MAPPING_CONFIG[category.name]);
+  let propertyPayload = constructPayload(message, MAPPING_CONFIG[category.name])!;
   // Extract other K-V property from traits about user custom properties
```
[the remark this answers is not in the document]
shekhar-rudder: This is a pure refactoring change. So I don't want to introduce any runtime behaviour changes

src/v0/destinations/klaviyo/transform.ts
```diff
@@ -68,7 +68,7 @@ const identifyRequestHandler = async (
     traitsInfo = addSubscribeFlagToTraits(traitsInfo);
   }
 
-  let propertyPayload = constructPayload(message, MAPPING_CONFIG[category.name]);
+  let propertyPayload = constructPayload(message, MAPPING_CONFIG[category.name])!;
```
maheshkutty, line 71: Why to add not null assertion here method will return null value if mapping is not available?

shekhar-rudder: Yes the method returns null if mappingJSON is invalid or empty

maheshkutty (approved):

PR #5288, opened 2026-06-12
feat(gaec): batch conversion adjustments with per-event partial failure handling

shekhar-rudder:
## Summary

- Implements the native batching framework for Google Ads Enhanced Conversions, grouping conversion adjustments by conversion name + customer into single API requests
- Adds a v1 network handler that maps partial failures per-event: when the API returns `partialFailureError` alongside a `results` array, each event is individually marked as success (200) or failure (400) based on its positional result, so successful events are not retried unnecessarily
- Adds `endpointPath` as a required field on the batching framework's `TransformedEvent` type, threaded through to the `batchedRequest` output for observability/stats tracking. Each destination sets a low-cardinality value (`/batch`, `/${action}`, `/lists/subscribe`, `/uploadConversionAdjustments`)

## Test plan

- [x] v0 handler unit tests pass (proxy + response handler unchanged)
- [x] v1 handler unit tests: table-driven tests covering success, partial failure with mixed results, all-failed batch, missing results array, undefined response body, non-2xx errors, statTags
- [x] Integration tests: v0 proxy (success + partial failure), v1 proxy (success, partial failure, invalid conversion, auth error), multi-event batch with mixed success/failure
- [x] Router tests: batching framework groups by conversion name, splits different conversions into separate batches
- [x] All 4 batching-framework destinations pass integration tests (GAEC: 30, PostHog: 17, iterable_audience: 20, custom_audience: 10)
- [x] TypeScript compiles clean, lint clean

### DIFF
--- a/src/v0/destinations/google_adwords_enhanced_conversions/networkHandler.js
+++ b/src/v0/destinations/google_adwords_enhanced_conversions/networkHandler.js
@@ -108,7 +108,12 @@ const gaecProxyRequest = async (request) => {
     googleAds,
   });
 
-  set(body.JSON, 'conversionAdjustments[0].conversionAction', `${conversionActionId}`);
+  // A request may carry multiple conversion adjustments when events are batched. They all
+  // share the same conversion name (grouping key), so the single resolved conversionActionId
+  // applies to every adjustment. For the non-batched path this is an array of one.
+  body.JSON.conversionAdjustments.forEach((_, index) => {
+    set(body.JSON, `conversionAdjustments[${index}].conversionAction`, `${conversionActionId}`);
+  });
 
   const response = await googleAds.addConversionAdjustMent(body.JSON);
 
--- a/src/v0/destinations/google_adwords_enhanced_conversions/networkHandler.js
+++ b/src/v0/destinations/google_adwords_enhanced_conversions/networkHandler.js
@@ -177,4 +177,4 @@ class networkHandler {
   }
 }
 
-module.exports = { networkHandler };
+module.exports = { networkHandler, gaecProxyRequest, gaecProcessAxiosResponse };
--- a/src/v0/destinations/google_adwords_enhanced_conversions/routerTransform.test.ts
+++ b/src/v0/destinations/google_adwords_enhanced_conversions/routerTransform.test.ts
@@ -1,0 +1,205 @@
+import { Integration } from './routerTransform';
+import { ChunkBatchStrategy } from '../../../services/destination/nativeBatching/batchDestination';
+import { processBatchedDestination } from '../../../services/destination/nativeBatching/processBatchedDestination';
+import type { Destination } from '../../../types/controlPlaneConfig';
+import type {
+  ProcessorTransformationOutput,
+  RouterTransformationRequestData,
+  RouterTransformationResponse,
+} from '../../../types/destinationTransformation';
+
+// ---------------------------------------------------------------------------
+// Test helpers
+// ---------------------------------------------------------------------------
+
+const destination: Destination = {
+  ID: 'gaec-dest-1',
+  Config: {
+    customerId: '1234567890',
+    subAccount: true,
+    loginCustomerId: '11',
+    listOfConversions: [{ conversions: 'Page View' }, { conversions: 'Product Added' }],
+    authStatus: 'active',
+  },
+  DestinationDefinition: {
+    ID: 'destDef-1',
+    Name: 'GOOGLE_ADWORDS_ENHANCED_CONVERSIONS',
+    DisplayName: 'Google Enhanced Conversions',
+    Config: {},
+  },
+  Name: 'google_adwords_enhanced_conversions',
+  Enabled: true,
+  WorkspaceID: 'ws-1',
+  Transformations: [],
+};
+
+type EventOverrides = { event?: string; type?: string; traits?: Record<string, unknown> };
+
+function makeInput(jobId: number, overrides?: EventOverrides): RouterTransformationRequestData {
+  const message = {
+    type: overrides?.type ?? 'track',
+    event: overrides?.event ?? 'Page View',
+    userId: '12345',
+    context: {
+      traits: overrides?.traits ?? { email: 'user@testmail.com' },
+    },
+    properties: {
+      gclid: 'gclid1234',
+      conversionDateTime: '2022-01-01 12:32:45-08:00',
+      order_id: 10000,
+      total: 1000,
+    },
+  };
+  const metadata = {
+    jobId,
+    userId: 'u1',
+    workspaceId: 'ws-1',
+    destinationId: 'gaec-dest-1',
+    destinationType: 'GOOGLE_ADWORDS_ENHANCED_CONVERSIONS',
+    secret: {
+      access_token: 'dummy-access-token',
+      refresh_token: 'dummy-refresh-token',
+      developer_token: 'dummy-developer-token',
+    },
+  };
+  return { message, metadata, destination } as unknown as RouterTransformationRequestData;
+}
+
+// The framework returns batchedRequest as a single output, an array, or undefined. For this
+// destination it is always a single output; narrow to it (and its JSON body) for assertions.
+const singleBatch = (resp: RouterTransformationResponse): ProcessorTransformationOutput => {
+  const { batchedRequest } = resp;
+  if (!batchedRequest || Array.isArray(batchedRequest)) {
+    throw new Error('expected a single batchedRequest');
+  }
+  return batchedRequest;
+};
+
+type EnhancedConversionsBody = { conversionAdjustments: unknown[]; partialFailure: boolean };
+
+const batchBody = (resp: RouterTransformationResponse): EnhancedConversionsBody =>
+  singleBatch(resp).body?.JSON as EnhancedConversionsBody;
+
+// ---------------------------------------------------------------------------
+// Tests
+// ---------------------------------------------------------------------------
+
+describe('GoogleAdwordsEnhancedConversions Integration', () => {
+  const integration = new Integration(destination);
+
+  describe('transformEvent', () => {
+    it('reshapes a single track event into a TransformedEvent carrying one adjustment', () => {
+      const result = integration.transformEvent(makeInput(1));
+
+      expect(result.endpoint).toBe('');
+      expect(result.method).toBe('POST');
+      expect(result.headers).toMatchObject({
+        Authorization: 'Bearer dummy-access-token',
+        'Content-Type': 'application/json',
+        'login-customer-id': '11',
+      });
+      expect(result.params).toMatchObject({
+        event: 'Page View',
+        customerId: '1234567890',
+        loginCustomerId: '11',
+        subAccount: true,
+      });
+      // body is the single conversion adjustment; the conversionAdjustments wrapper and
+      // partialFailure flag are re-added by wrapBody at batch time.
+      expect(result.body).toHaveProperty('adjustmentType', 'ENHANCEMENT');
+      expect(result.body).toHaveProperty('userIdentifiers');
+      expect(result.body).not.toHaveProperty('conversionAdjustments');
+    });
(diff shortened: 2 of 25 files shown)


src/v0/destinations/google_adwords_enhanced_conversions/networkHandler.js
maheshkutty, line 2: @shekhar-rudder We need to do a partial batching handling
https://github.com/rudderlabs/rudder-transformer/blob/develop/src/v1/destinations/google_adwords_offline_conversions/networkHandler.js#L335
https://developers.google.com/google-ads/api/reference/rpc/v24/ConversionAdjustmentUploadService/UploadConversionAdjustments?transport=rest#response-body

test/integrations/destinations/google_adwords_enhanced_conversions/router/batching-data.ts
```diff
@@ -0,0 +1,275 @@
...
+          ],
+        },
+      },
+    },
+    envOverrides,
+  },
+  {
+    name: 'google_adwords_enhanced_conversions',
+    description:
+      'Batching Framework: events with different conversion names are split into separate batches',
```
maheshkutty, line 173: Is an API request expect all the conversionAction should be of same type?

maheshkutty:
@shekhar-rudder We can add endpointPath for GAEC
https://github.com/rudderlabs/rudder-transformer/blob/develop/src/v0/destinations/google_adwords_remarketing_lists/util.ts#L115

shekhar-rudder:
> We can add endpointPath for GAEC

@maheshkutty Can you explain this part in detail

maheshkutty:
@shekhar-rudder adding endpointPath helps us know metrics around external APIs like the number of API calls, HTTP status, and latency.

src/v0/destinations/google_adwords_enhanced_conversions/networkHandler.js
shekhar-rudder: Addressed

test/integrations/destinations/google_adwords_enhanced_conversions/router/batching-data.ts
shekhar-rudder: conversion action (`event` field) is part of the query param, so if they are different, batching framework would automatically put them in separate chunks.

src/v0/destinations/google_adwords_enhanced_conversions/routerTransform.ts
```diff
@@ -0,0 +1,61 @@
...
+      params: result.params,
+    };
+  }
+
+  getBatchStrategy(): BatchStrategy<ConversionAdjustment> {
+    return new ChunkBatchStrategy<ConversionAdjustment>({
+      maxItems: MAX_CONVERSION_ADJUSTMENTS_PER_BATCH,
+      wrapBody: (bodies) => ({
+        conversionAdjustments: bodies,
+        partialFailure: true,
```
krishna2020, line 41: why customer_id is not being passed here ? https://developers.google.com/google-ads/api/reference/rpc/v24/UploadConversionAdjustmentsRequest

src/v0/destinations/google_adwords_enhanced_conversions/routerTransform.ts
```diff
@@ -0,0 +1,61 @@
...
+      // accessToken — all part of the framework's grouping key, so events only batch together
+      // when they target the same conversion action + customer, as the Google Ads API requires.
+      params: result.params,
+    };
+  }
+
+  getBatchStrategy(): BatchStrategy<ConversionAdjustment> {
+    return new ChunkBatchStrategy<ConversionAdjustment>({
+      maxItems: MAX_CONVERSION_ADJUSTMENTS_PER_BATCH,
+      wrapBody: (bodies) => ({
```
krishna2020, line 39: tangentially, I see `validate_only` option here. Can this be exposed to end user for scenarios where customers ask API responses ?

src/v0/destinations/google_adwords_enhanced_conversions/routerTransform.test.ts
```diff
@@ -0,0 +1,205 @@
...
+      expect(batch.batched).toBe(true);
+      expect(batch.statusCode).toBe(200);
+      const body = batchBody(batch);
+      expect(body.conversionAdjustments).toHaveLength(3);
+      expect(body.partialFailure).toBe(true);
+      expect(singleBatch(batch).params).toMatchObject({ event: 'Page View' });
+      expect(batch.metadata.map((m) => m.jobId)).toEqual([1, 2, 3]);
+    });
+
+    it('splits events with different conversion names into separate batches', async () => {
```
krishna2020, line 178: the high level request looks agnostic to event type, why to split events ? https://developers.google.com/google-ads/api/reference/rpc/v24/UploadConversionAdjustmentsRequest

src/v0/destinations/google_adwords_enhanced_conversions/routerTransform.test.ts
```diff
@@ -0,0 +1,205 @@
...
+      const inputs = [
+        makeInput(1, { event: 'Page View' }),
+        makeInput(2, { event: 'Product Added' }),
+        makeInput(3, { event: 'Page View' }),
+      ];
+      const results = await processBatchedDestination(inputs, Integration, {});
```
krishna2020, line 184: getBatchStrategy doesn't configure any groupKey - which part of code is splitting the events based on event name ?

src/v0/destinations/google_adwords_enhanced_conversions/networkHandler.js
krishna2020, line 106: ideally, this has to be part of transform.js ?

src/v0/destinations/google_adwords_enhanced_conversions/routerTransform.test.ts
krishna2020: If fetching conversionActionId moves as part of event transformation, can split can be avoided ?

src/v0/destinations/google_adwords_enhanced_conversions/networkHandler.js
shekhar-rudder: Yes. I'm not sure why this was added in network handler

src/v0/destinations/google_adwords_enhanced_conversions/routerTransform.test.ts
shekhar-rudder: groupKey is {endpoint, header, params}

src/v0/destinations/google_adwords_enhanced_conversions/routerTransform.test.ts
shekhar-rudder: Yes. `conversionActionId` is being sent as part of request body to Google so we don't really need it as part of group key. If we can refactor network handler, no split is required

src/v0/destinations/google_adwords_enhanced_conversions/routerTransform.ts
shekhar-rudder: Customer id is being sent as part of query params

src/v0/destinations/google_adwords_enhanced_conversions/routerTransform.ts
shekhar-rudder: > Can this be exposed to end user for scenarios where customers ask API responses ?

Based on my understanding, customers asking for API responses are seeing 200 in RudderStack. So adding `validate_only` won't help since it is meant only for invalid data

src/v1/destinations/google_adwords_enhanced_conversions/networkHandler.ts
```diff
@@ -0,0 +1,122 @@
...
+      },
+      response: responseWithIndividualEvents,
+    };
+  }
+
+  // Non-2xx status — complete failure for all events
+  const { response } = destinationResponse;
+  const errMessage = response?.error?.message || '';
+  throw new NetworkError(
+    `${errMessage}" during Google_adwords_enhanced_conversions response transformation`,
```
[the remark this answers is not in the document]
shekhar-rudder: Addressed

src/v0/destinations/iterable_audience/routerTransform.ts
```diff
@@ -103,14 +103,15 @@ class IterableAudienceIntegration extends BatchDestination<
...
-    const endpoint =
+    const { endpoint, endpointPath } =
       action === 'subscribe'
         ? getSubscribeEndpoint(this.accountConfig.dataCenter)
         : getUnsubscribeEndpoint(this.accountConfig.dataCenter);
 
     return {
       body: { action, subscriber },
       endpoint,
+      endpointPath,
```
maheshkutty, line 114: Why add endpointPath support for other destinations as part of this pr?
We can have a separate PR for that

shekhar-rudder: I made it a mandatory field in the batching framework, which caused TS errors for other batching framework destinations. I could have kept it optional, but I felt it was better to update those in this PR itself rather than leaving the parameter optional.

src/v1/destinations/google_adwords_enhanced_conversions/networkHandler.ts
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * Per-event response handler for batched Google Ads Enhanced Conversions requests.
+ *
+ * When the API returns a partialFailureError alongside a results array, each event
+ * is individually mapped to success or failure based on whether its positional entry
+ * in results is populated or empty.
+ *
+ * Ref: https://developers.google.com/google-ads/api/reference/rpc/v17/UploadConversionAdjustmentsResponse
+ */
+const gaecResponseHandler = (responseParams: ResponseHandlerParams): ResponseProxyObject => {
```
maheshkutty, line 40: Why the half logic in v0 and other in v1 proxy?

maheshkutty: I think we should copy paste the logic of v0 here, because it will increase code readibilty. All the proxy changes will in single file

shekhar-rudder: I didn't want to duplicate any code so went with reusing the v0 methods in v1.

maheshkutty: I prefer all proxy code in a single file rather than spread across different files

test/integrations/destinations/google_adwords_enhanced_conversions/router/batching-data.ts
maheshkutty: Means batching capability will not be fully utilized, and it is not deterministic as well; batching now depends on the number of unique events in a batch of events.

maheshkutty: I think we might need to update the grouping logic of frameworks, for we can skip it

maheshkutty (approved):

src/v1/destinations/google_adwords_enhanced_conversions/networkHandler.ts
```diff
@@ -0,0 +1,122 @@
...
+        return {
+          statusCode: isEventFailed ? 400 : 200,
+          metadata,
+          error: isEventFailed ? errorMessage : 'success',
+        };
+      },
+    );
+
+    return {
+      status: 400,
```
krishna2020, line 84: Is 400 root level status expected by rudder-server in case of partial failures ?

krishna2020 (approved):

src/v1/destinations/google_adwords_enhanced_conversions/networkHandler.ts
shekhar-rudder: We are following the same pattern in GAOC. My guess is that server doesn't read this field - it cares about the context level status code. And then it checks per job status code

PR #5291, opened 2026-06-15
fix: cleanup survicate cloud mode destination

achettyiitr:

- Survicate code-quality changes on top of the partner contributor changes.

- Resolves INT-6537

### DIFF
--- a/src/v0/destinations/survicate/transform.ts
+++ b/src/v0/destinations/survicate/transform.ts
@@ -8,24 +8,23 @@
  * - track: Track user events
  */
 
-import { ConfigurationError, InstrumentationError } from '@rudderstack/integrations-lib';
-import { defaultRequestConfig, simpleProcessRouterDest } from '../../util';
+import { formatZodError, InstrumentationError } from '@rudderstack/integrations-lib';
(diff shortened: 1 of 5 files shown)


src/v0/destinations/survicate/transform.ts
```diff
@@ -192,23 +181,15 @@ const processEvent = (event: SurvicateRouterRequest) => {
...
-  // Normalize + validate once here so each handler receives a validated message.
-  const msg = validateMessage(message);
-
   // Route based on message type. Unsupported types are already rejected by the
-  // schema's discriminated union, so `msg.type` is exhaustively narrowed here.
-  switch (msg.type) {
+  // schema's discriminated union, so `message.type` is exhaustively narrowed here.
+  switch (message.type) {
     case 'identify':
-      return processIdentifyEvent(msg, destination.Config);
```
krishna2020, line 207: Is it not better to pass destination.Config itself ?

krishna2020 (approved):

src/v0/destinations/survicate/transform.ts
achettyiitr: Since there is only one field in `destination.Config` "destinationKey", it's not needed. Passing entire thing is not needed.

PR #5303, opened 2026-06-19
feat(customerio): add v2 API support with native batching framework

maheshkutty:
🔒 Scanned for secrets using gitleaks 8.30.1

Implements Customer.io v2 batch API transformer using the native batching framework. Routes v2-enabled workspaces through CustomerIOIntegration (BatchDestination) targeting /api/v2/batch, while keeping the v1 path unchanged for all other workspaces.

Fixes in v2 util:
- Guard last_used against undefined timestamp (NaN sent to API)
- Replace all dots in trait keys, not just the first (regex /\./g)
- Use null-check for previousId to allow numeric 0 (Zod allows z.number())
- Zod refine now validates alias events require userId + previousId

[Resolves INT-6612](https://linear.app/rudderstack/issue/INT-6612/migrate-customerio-destination-to-v2-apis)

N/A

N/A

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist


src/v0/destinations/customerio/v2/config.ts
```diff
@@ -0,0 +1,44 @@
...
+const V2_HOST_EU = 'track-eu.customer.io';
+
+const host = (datacenter?: string): string => (datacenter === 'EU' ? V2_HOST_EU : V2_HOST);
+
+const getV2Endpoint = (datacenter?: string): string => `https://${host(datacenter)}/api/v2/batch`;
+const V2_BATCH_PATH = 'v2/batch';
+
+const MAX_OBJECT_SIZE_BYTES = 32 * 1024; // 32KB per object
+const MAX_BATCH_PAYLOAD = '500KB';
+
```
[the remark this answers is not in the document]
maheshkutty: Done

src/v0/destinations/customerio/v2/util.ts
```diff
@@ -0,0 +1,267 @@
...
+    return { id: String(userId) };
+  }
+  if (email) {
+    return { email: String(email) };
+  }
+  if (message.anonymousId) {
+    return { anonymous_id: String(message.anonymousId) };
+  }
+  throw new InstrumentationError('userId, email or anonymousId is required');
+};
```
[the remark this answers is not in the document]
maheshkutty: Only string value is allowed; updated the ZOD validation

src/v0/destinations/customerio/v2/types.ts
```diff
@@ -0,0 +1,76 @@
...
+        .refine(
+          (msg) => {
+            if (msg.type === 'alias') {
+              return !!msg.userId && msg.previousId != null;
+            }
+            if (msg.type === 'group') {
+              return true;
+            }
+            return !!msg.userId || !!msg.anonymousId || hasEmailTrait(msg);
+          },
```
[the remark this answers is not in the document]
maheshkutty: It is fixed; we will now accept string value only

src/v0/destinations/customerio/v2/util.ts
```diff
@@ -0,0 +1,267 @@
...
+  if (userId) {
+    return { id: String(userId) };
+  }
+  if (email) {
+    return { email: String(email) };
+  }
+  if (message.anonymousId) {
+    return { anonymous_id: String(message.anonymousId) };
+  }
+  throw new InstrumentationError('userId, email or anonymousId is required');
```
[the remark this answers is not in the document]
maheshkutty: It is fixed; we will now accept string value only

src/v0/destinations/customerio/transform.ts
```diff
@@ -129,7 +136,13 @@ function processSingleMessage(message, destination) {
 }
 
 function process(event) {
-  const { message, destination } = event;
+  const { message, destination, metadata } = event;
+  if (isBatchingFrameworkEnabled('CUSTOMERIO', metadata?.workspaceId ?? '')) {
+    // v2 output is a complete delivery request but not shaped like the v1
```
shekhar-rudder: This destination is router-only (`transformAt: router`), so `doProcessorTransformation` never calls `process()` in production — only `doRouterTransformation` runs, and that path goes through `routerTransform.ts` → `transformEvent` → `processV2`. This `isBatchingFrameworkEnabled` branch in `process()` therefore only services a code path that production never exercises.

src/v0/destinations/customerio/v2/util.ts
```diff
@@ -0,0 +1,264 @@
...
+export const buildRequestMeta = (
+  destination: CustomerIODestination,
+): {
+  endpoint: string;
+  endpointPath: string;
+  method: string;
+  headers: Record<string, unknown>;
+} => ({
+  endpoint: getV2Endpoint(destination.Config?.datacenter),
+  endpointPath: V2_BATCH_PATH,
```
shekhar-rudder, line 258: `destination.Config?.datacenter` uses optional chaining on `Config`, but `destination.Config.siteID` / `destination.Config.apiKey` just below access it directly. `Config` is server-provided and always present, so the `?.` on `Config` is unnecessary and inconsistent. Same pattern in `deviceActionFor` (`destination.Config?.deviceTokenEventName`). Suggest `destination.Config.datacenter` and `destination.Config.deviceTokenEventName`.

shekhar-rudder (commented):
A couple of points on the v2 migration — one router-only cleanup and one config-access consistency nit. The routing/gating, the four util fixes, and the test fixtures all check out.

src/v0/destinations/customerio/v2/transform.ts
```diff
@@ -0,0 +1,23 @@
...
+import { CustomerIODestination, CustomerIOV2Payload, CustomerIOV2ProcessorOutput } from './types';
+
+export const processV2 = (event: {
+  message: RudderMessage;
+  destination: CustomerIODestination;
+}): CustomerIOV2ProcessorOutput => {
+  const { message, destination } = event;
+  validateConfigFields(destination);
+  const envelope = buildEnvelope(message, destination);
+  const { endpoint, endpointPath, method, headers } = buildRequestMeta(destination);
```
shekhar-rudder: Regression for RETL / warehouse sources: the v1 path runs the `mappedToDestination` preprocessing before building the payload (`transform.ts` `processSingleMessage`):

```ts
const mappedToDestination = get(message, MappedToDestinationKey);
if (mappedToDestination) {
  addExternalIdToTraits(message);
  adduserIdFromExternalId(message);
}
```

`processV2` doesn't do this, so for `mappedToDestination: true` events the v2 path won't derive `userId` from `context.externalId` (so `personIdentifiers` falls through to email/anonymousId or throws `userId, email or anonymousId is required`) and won't fold `externalId` into traits. RETL is documented as supported for this destination (see README), and this breaks silently — the migration fixtures were generated from the existing v1 input, so there's no RETL case exercising it.

Please run the same `addExternalIdToTraits` / `adduserIdFromExternalId` preprocessing here before `buildEnvelope`, and add a `mappedToDestination` fixture to cover it.

shekhar-rudder (commented):
One more from a v1-vs-v2 parity pass: the RETL / mappedToDestination preprocessing isn't carried over to the v2 path.

src/v0/destinations/customerio/v2/transform.ts
maheshkutty: Good catch — this was a real silent regression for RETL/warehouse sources. Fixed in 541d124: `processV2` now runs `addExternalIdToTraits` / `adduserIdFromExternalId` before `buildEnvelope`, mirroring v1's `processSingleMessage`. Added a `mappedToDestination` processor fixture, and the two pre-existing RETL fixtures (v2 Test 43/44) now correctly fold the `externalId` into traits (they had captured the buggy output before).

src/v0/destinations/customerio/v2/util.ts
maheshkutty: Fixed in 541d124 — dropped the unnecessary `?.` on `Config` in both `buildRequestMeta` (`Config.datacenter`) and `deviceActionFor` (`Config.deviceTokenEventName`) for consistency with the direct `Config.siteID`/`Config.apiKey` access.

src/v0/destinations/customerio/transform.ts
maheshkutty: Correct — the destination is router-only, so this branch isn't exercised in production. I've kept it (it's the path the processor-mode component tests in `processor/dataV2.ts` use to validate `processV2` output directly) and updated the comment in 541d124 to explain that intent rather than the previous v1/v2-seam wording. Happy to remove it entirely if you'd prefer the v2 logic be tested only via the router fixtures.

shekhar-rudder: Yes lets remove it

maheshkutty: We need to remmove the processor test as well
I am not sure if we should do that, until we migrate existing pipeline to v2

shekhar-rudder (commented):
Follow-up on the RETL fix: the externalId preprocessing runs after schema validation in the router path, so externalId-only RETL events are still rejected upfront.

src/v0/destinations/customerio/v2/types.ts
```diff
@@ -0,0 +1,76 @@
...
+        .refine(
+          (msg) => {
+            if (msg.type === 'alias') {
+              return !!msg.userId && !!msg.previousId;
+            }
+            if (msg.type === 'group') {
+              return true;
+            }
+            return !!msg.userId || !!msg.anonymousId || hasEmailTrait(msg);
+          },
```
shekhar-rudder, line 74: RETL handling is still incomplete on the production (router) path due to validation ordering.

`processBatchedDestination` runs `validateInputs` -> `getInputSchema().safeParse()` (step 2) *before* `transformEvents` -> `processV2` (step 4), where `adduserIdFromExternalId` derives `userId` from `context.externalId`. This refine only accepts `userId || anonymousId || email`, with no awareness of `externalId` / `mappedToDestination`:

```ts
return !!msg.userId || !!msg.anonymousId || hasEmailTrait(msg);
```

So the canonical RETL identify — warehouse supplies the identifier via `context.externalId` (type `userId`), no top-level `userId`, no `anonymousId`, no email trait — is rejected here with `400 "userId, email or anonymousId is required"`, before the preprocessing that would have made it valid ever runs.

The new RETL fixture doesn't catch this: it's a `processor` test (that path skips `getInputSchema` entirely) and it also happens to carry an email trait. `router/dataV2.ts` has zero `mappedToDestination`/`externalId` fixtures, so the path that actually runs this schema in production has no RETL coverage.

Suggest making the refine RETL-aware (e.g. also pass when `context.externalId` / `context.mappedToDestination` is present) and adding a **router** fixture for an externalId-only RETL identify (no email) to lock it in.

src/v0/destinations/customerio/transform.ts
shekhar-rudder: I'm suggesting to remove processV2 from process method. No need to remove any processor tests

maheshkutty: These test will start failing
test/integrations/destinations/customerio/processor/dataV2.ts

src/v0/destinations/customerio/v2/types.ts
maheshkutty: Fixed — made the schema refine RETL-aware by adding an `isMappedToDestination` guard that checks `context.mappedToDestination`. RETL events that carry their identifier in `context.externalId` (no top-level `userId`/`anonymousId`/email) now pass schema validation; `adduserIdFromExternalId` then hydrates `userId` inside `processV2` as before.

Also added a router fixture that covers this exact path: an externalId-only RETL identify with no email, exercising the full `processBatchedDestination` → `validateInputs` → `transformEvents` → `processV2` flow.

src/v0/destinations/customerio/transform.ts
shekhar-rudder: Ok now I get the problem. We shouldn't have v2 processor tests, only v1 processor tests and router v1 and v2 tests.

src/v0/destinations/customerio/v2/types.ts
```diff
@@ -0,0 +1,85 @@
...
+  return !!traits?.email;
+};
+
+// RETL/warehouse sources supply the identifier via context.externalId and set
+// context.mappedToDestination. The identifier is derived by adduserIdFromExternalId
+// inside processV2, which runs *after* schema validation. Allow these events through
+// so preprocessing can hydrate userId before the payload is built.
+const isMappedToDestination = (msg: Record<string, unknown>): boolean => {
+  const ctx = msg.context as Record<string, unknown> | undefined;
+  return !!(ctx?.mappedToDestination);
```
shekhar-rudder: These `as Record<string, unknown>` casts (here and in `hasEmailTrait` just above) are a smell: they exist only because `context` and `traits` aren't declared in the inner `z.object({...})`, so the refine sees them as untyped passthrough fields and has to cast them back to read `context.mappedToDestination` / `traits.email`.

Cleaner to declare the fields you actually inspect in the schema so the refine reads them type-safely and the casts (and these helper wrappers) go away, e.g.:

```ts
context: z
  .object({
    mappedToDestination: z.unknown(),
    traits: z.object({ email: z.unknown() }).passthrough().nullish(),
  })
  .passthrough()
  .nullish(),
traits: z.object({ email: z.unknown() }).passthrough().nullish(),
```

Then `msg.context?.mappedToDestination` and `msg.traits?.email` are typed — no casts needed. (Keep the inspected leaves as `z.unknown()` rather than `z.string()` so you don't tighten validation and start rejecting events that pass today.)

shekhar-rudder (commented):
Small follow-up on the RETL-aware refine: the casts can be avoided by declaring the inspected fields in the schema.

src/v0/destinations/customerio/v2/types.ts
maheshkutty: Good call. Declared `traits` and `context` (with `mappedToDestination` and `context.traits`) directly in the inner `z.object({...})` as `z.unknown()` leaves (so no validation tightening), then read them type-safely in the refine — no `as Record<string, unknown>` casts needed. The `hasEmailTrait` and `isMappedToDestination` helpers are gone.

src/v0/destinations/customerio/transform.ts
maheshkutty: Done — removed `processV2` from `process()` entirely (along with the `isBatchingFrameworkEnabled` import and the `processV2` import). `process()` now only runs the v1 path.

On the test side, deleted `processor/dataV2.ts` and removed the `v2data` spread from `processor/data.ts`. The unique v2 behaviors from those tests (android add/delete_device, iPadOS → ios, screen, page fallbacks, group relationship actions, EU datacenter, alias → merge, RETL via externalId, error cases) have been migrated to `router/dataV2.ts` where they exercise the full production path.

src/v0/destinations/customerio/routerTransform.ts
```diff
@@ -0,0 +1,51 @@
...
+import { processV2 } from './v2/transform';
+import { getV2InputSchema, CustomerIOV2Payload, CustomerIODestinationConfig } from './v2/types';
+import { MAX_OBJECT_SIZE_BYTES, MAX_BATCH_PAYLOAD } from './v2/config';
+
+class CustomerIOIntegration extends BatchDestination<
+  CustomerIOV2Payload,
+  CustomerIODestinationConfig
+> {
+  transformEvent(input: RouterTransformationRequestData): TransformedEvent<CustomerIOV2Payload> {
+    const result = processV2({ message: input.message, destination: this.destination });
```
shekhar-rudder: `processV2` builds a full `defaultRequestConfig()` response (`body.JSON.batch = [envelope]`, empty `JSON_ARRAY/XML/FORM`, `statusCode`) only for `transformEvent` to immediately unwrap it (`const [body] = result.body.JSON!.batch`) and rebuild a `TransformedEvent`. Since `processV2` has no other caller, the request payload is effectively constructed three times (here, in `processV2`, and in `buildRequestMeta`).

Consider having `transformEvent` call `buildEnvelope(input.message, this.destination)` directly for the body and `buildRequestMeta(this.destination)` for the meta, returning the `TransformedEvent` from those. That drops `processV2`, the `CustomerIOV2ProcessorOutput` type, and the batch wrap/unwrap round-trip. And even with that direct call, the request should not be rebuilt in `transformEvent` — `buildRequestMeta` should remain the single source for endpoint/method/headers.

src/v0/destinations/customerio/transform.ts
```diff
@@ -194,8 +199,11 @@ const processRouterDest = (inputs, reqMetadata) => {
         });
       } else {
         // if not transformed
+        // v2-enabled events are routed through routerTransform.ts (BatchDestination
+        // framework) before processRouterDest is reached, so process() always
+        // returns the v1 shape here.
         const transformedPayload = {
-          message: process(event),
+          message: process(event) as CustomerIOBatchedRequest,
```
shekhar-rudder: Since v2 events no longer flow through `processRouterDest`, this change (the `CustomerIOBatchedRequest` import + `as CustomerIOBatchedRequest` cast + comment) looks like leftover from an earlier approach. Can this file be reverted to the original `message: process(event)`?

test/integrations/destinations/customerio/processor/data.ts
```diff
@@ -8457,3 +8457,5 @@ export const data = [
     },
   },
 ];
+
+export const data = [...v1data];
```
shekhar-rudder: Now that the processor v2 test was removed, this `v1data` split is a no-op (`export const data = [...v1data]`). Can it be reverted to the original `export const data = [...]`? (`router/data.ts` should stay as-is — it genuinely uses `v2data`.)

shekhar-rudder (commented):
Reviewed the v2 batching implementation — overall a clean, well-tested change that mirrors the established posthog BatchDestination pattern. A few cleanup suggestions around redundant request construction and now-unnecessary changes.

src/v0/destinations/customerio/routerTransform.ts
maheshkutty: Done in c6e9761e1. `transformEvent` now calls `buildEnvelope(message, this.destination)` for the body and spreads `buildRequestMeta(this.destination)` for endpoint/method/headers directly — no more wrap/unwrap round-trip. `buildRequestMeta` stays the single source for the request meta. Dropped `processV2` (deleted `v2/transform.ts`, no other caller) and the `CustomerIOV2ProcessorOutput` type; moved `validateConfigFields` + the RETL preprocessing inline into `transformEvent`.

src/v0/destinations/customerio/transform.ts
maheshkutty: Done in c6e9761e1. Reverted to `message: process(event)` and removed the leftover `CustomerIOBatchedRequest` import + cast + comment.

test/integrations/destinations/customerio/processor/data.ts
maheshkutty: Done in c6e9761e1. Reverted to `export const data = [...]` (dropped the `v1data` split). `router/data.ts` left as-is since it genuinely uses `v2data`.

shekhar-rudder (approved):

PR #5318, opened 2026-06-25
feat(versioning): carry integration major on Destination + dev-only test_destination

saikumarrs:
## What

The data-plane half of integration definition-versioning in the transformer — carry the **integration major** on the destination and dispatch on it, plus a dev-only destination that exercises it end-to-end. No production destination grows a real per-major branch here.

### INT-6495 — version on the `Destination` type + dispatch idiom
- `version?: number` on the `Destination` type (mirrors Go `DestinationT.Version`). Optional because it can be undefined — `0` / `undefined` / `1` all dispatch as v1 (`Number(undefined)` is `NaN`), so existing destinations behave identically.
- `destinationVersion?: number` received on the proxy/dataDelivery payload (`ProxyV0/V1Request` + zod schemas) — the proxy route carries no destination object, so the major arrives as a sibling field.
- Dispatch idiom documented in `CONTRIBUTING.md` and the `code-structure` Claude skill (in-file branch by default; sibling-module escalation; never route majors through `getDestHandler`).

### INT-6492 — dev-only `test_destination` (v1 only)
- `src/v0/destinations/test_destination/` dispatches on `Number(destination.version)` in `transform`/`processRouterDest` and on `destinationVersion` in `networkHandler`; `major >= 2` throws not-implemented (only v1 is built).
- Processor / router / dataDelivery integration fixtures (v1, no-version, v2-rejection) + a co-located unit test.
- Router capability registered in `features.ts`. Clearly labelled **"DEV-ONLY — NOT A REAL DESTINATION"**; uses the reserved `.invalid` endpoint. Control-plane exposure is gated to dev only (other repos).

## Out of scope
- No real per-major branch for any production destination (rollout milestone).
- `test_destination` v2 path — stubbed to throw; only v1 implemented.

## Testing
- `test_destination` unit 4/4 · component (processor/router/dataDelivery) 7/7
- `webhook` smoke 23/23 · `misc` (getFeatures) 2/2 — no regression
- Full pre-commit suite (`test:ts` + `test:js` + lint) green.

## References
- Linear: [INT-6495](https://linear.app/rudderstack/issue/INT-6495) (version on the `Destination` type + dispatch idiom) · [INT-6492](https://linear.app/rudderstack/issue/INT-6492) (dev-only `test_destination`)
- Paired (data plane sender): rudderlabs/rudder-server#7125 — carries `version` on `DestinationT`; sends `destination.version` (transform/router) + top-level `destinationVersion` (proxy).
- Stacked follow-up: #5319 — [INT-6576](https://linear.app/rudderstack/issue/INT-6576) `destVersion` metric label.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -954,14 +954,18 @@ A destination can ship multiple **integration majors** (e.g. v1 → v2 of its co
 > - **transformer architecture version** — the `src/v0` / `v1` / `v2` directory (`getDestHandler`'s `version` argument, the `version: 'v0'` field in component tests). Unrelated.
 > - **REST request-config version** — `version: '1'` inside `defaultRequestConfig()` / the proxy request shape. Unrelated.
 
-**The idiom — an in-file inline branch, by default.** At the top of your entry `process`, branch on `Number(event.destination.version)`. Keep the existing v1 logic in place and factor shared logic into `utils.ts`:
+**The idiom — an in-file inline branch, by default.** At the top of your entry `process`, branch on `getDestinationVersion(event.destination.version)`. Keep the existing v1 logic in place and factor shared logic into `utils.ts`:
 
 ```ts
+import { getDestinationVersion } from '../../../util/utils';
+
 const V2_MAJOR = 2;
 
 const process = (event) => {
-  // 0 / undefined / 1 all resolve to Number(...) < 2 and run the existing v1 path (defensive).
-  const major = Number(event.destination.version);
+  // getDestinationVersion normalizes the raw major (0 / undefined / non-numeric -> 1, fractions
+  // truncated), so 0 / undefined / 1 all run the existing v1 path (defensive) and metrics never
+  // see a bogus destVersion=0.
+  const major = getDestinationVersion(event.destination.version);
   if (major >= V2_MAJOR) {
     return processV2(event); // new major
   }
@@ -969,7 +973,7 @@ const process = (event) => {
 };
 ```
 
-- **Read `destination.version`, never `Config.apiVersion`.** It is a number and may be absent on routes that don't carry it — `Number(undefined)` is `NaN`, which falls to v1. So `0`, `undefined`, and `1` all behave identically (v1).
+- **Normalize with `getDestinationVersion(...)` (`src/util/utils.js`); read `destination.version`, never `Config.apiVersion`.** The raw major is a number that may be absent on routes that don't carry it. `getDestinationVersion` maps `0` / `undefined` / non-numeric to `1` and truncates fractions to a whole integer, so `0`, `undefined`, and `1` all behave identically (v1) and dashboards never see `destVersion=0`. Branching on the bare `Number(event.destination.version)` skips this normalization — always go through the helper.
 - **Shared logic lives in `utils.ts`**; v1 stays where it is. Until a destination ships a new major, every destination is v1 and runs exactly as today.
 - **Escalate to a sibling module only on large divergence.** If the v2 branch grows large, move it to a sibling module following the existing repo convention (`transformV2.ts` / `<dest>Transform-v2.ts`, as Klaviyo/HubSpot/TikTok/Snapchat do). **Prefer a sibling module over `./v1` / `./v2` subdirectories** — no destination uses subdirs today, so they fragment the layout — but they aren't strictly off-limits if a major is large enough to warrant its own tree. The entry `transform` stays the dispatcher either way.
 - **Branch `routerTransform`, `deleteUsers`, and the proxy/delivery handler only when a major actually changes them.** Most majors only touch `transform`.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -980,7 +980,7 @@ const process = (event) => {
 - **The proxy / dataDelivery route has no destination object** — the major arrives as a top-level **`destinationVersion`** on the proxy request payload. Read it there (e.g. in `networkHandler`'s `proxy`/`responseHandler`), not `destination.version`.
 - **Never route integration majors through `getDestHandler`.** That argument is the architecture version; routing majors there would explode the directory matrix.
 
-See `src/v0/destinations/test_destination` for a worked, dev-only example exercising this idiom across `process`, `processRouterDest`, and `networkHandler`.
+See `src/v0/destinations/test_destination` for a worked, dev-only example exercising this idiom across `process` and `networkHandler`. Its router runs through the native batching framework (`routerTransform.ts`), which reuses `process`, so the major dispatch lives in one place rather than a separate `processRouterDest`.
 
 ### 3. Test your destination integration
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -945,6 +945,39 @@ And here's an example of the mapping configuration files:
 ]
 ```
 
+#### Dispatching on the integration major (`destination.version`)
+
+A destination can ship multiple **integration majors** (e.g. v1 → v2 of its config + API). The control plane stamps the major on the destination and the data plane **carries** it to the transformer — there is no up-conversion (stored shape = delivered shape = processed shape). Your job is to read that major and branch on it **inside your own destination's code**.
+
+> **Terminology trap.** Three unrelated "versions" exist — do not conflate them:
+> - **integration major** — `destination.version` (a number; `destinationVersion` on the proxy payload). **This is the axis you branch on.**
+> - **transformer architecture version** — the `src/v0` / `v1` / `v2` directory (`getDestHandler`'s `version` argument, the `version: 'v0'` field in component tests). Unrelated.
+> - **REST request-config version** — `version: '1'` inside `defaultRequestConfig()` / the proxy request shape. Unrelated.
+
+**The idiom — an in-file inline branch, by default.** At the top of your entry `process`, branch on `Number(event.destination.version)`. Keep the existing v1 logic in place and factor shared logic into `utils.ts`:
+
+```ts
+const V2_MAJOR = 2;
+
+const process = (event) => {
+  // 0 / undefined / 1 all resolve to Number(...) < 2 and run the existing v1 path (defensive).
+  const major = Number(event.destination.version);
+  if (major >= V2_MAJOR) {
+    return processV2(event); // new major
+  }
+  return processV1(event); // existing v1 logic, unchanged
+};
+```
+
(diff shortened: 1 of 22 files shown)


saikumarrs:
Stacked follow-up (metric label, INT-6576): #5319 — review/merge this one first.

saikumarrs:
Pushed `cb10b1b95` — addressed code-review findings on the integration-major dispatch:

- **`getDestinationVersion` truncates to a whole integer** (`Math.trunc`) so a fractional major can't leak to metrics or mis-route; added float test cases.
- **`TEST_DESTINATION` gated out of the prod `/features` payload** (`NODE_ENV`) rather than relying solely on control-plane gating.
- **`networkHandler` composes on the generic handler** and overrides only `proxy` (drops the verbatim copy, so shared response-contract changes reach the fixture).
- Dropped the forked `TestDestinationResponseParams` type + unused import.
- Shared `V2_MAJOR` from `config.ts` across `transform`/`networkHandler`.
- Renamed `process` → `transformEvent` (exported as `process`) to stop shadowing the Node global.
- Deduped fixtures via `v1RequestShape` + `v2StatTags` helpers in `common.ts`.
- Docs (CONTRIBUTING + code-structure SKILL) now branch via `getDestinationVersion`, not bare `Number(...)`, so the normalization isn't bypassed.

Verified on Node 22.22.3 (`.nvmrc`): lint clean, `tsc --noEmit` 0 errors, full pre-commit suite green (component + 2334 unit tests).

src/v0/destinations/test_destination/transform.ts
```diff
@@ -0,0 +1,25 @@
...
+// Named transformEvent (not `process`) to avoid shadowing the Node.js global `process`; exported as `process`.
+const transformEvent = (event: TestDestinationProcessorRequest) => {
+  const major = getDestinationVersion(event.destination.version);
+  if (major >= V2_MAJOR) {
+    throw new ConfigurationError('test_destination v2 transformation is not yet implemented');
+  }
+  return processV1(event);
+};
+
+const processRouterDest = async (
```
shekhar-rudder: Lets use batching framework here even though batching is not required and it is a dummy destination

saikumarrs: Done in a1cbd30d0 — added `routerTransform.ts` (a `BatchDestination` subclass) that reuses the version-dispatching per-event `process` and wraps payloads under `batch`, registered `TEST_DESTINATION` in `batchedDestinationsMap` (dev-only), and added `routerTransform.test.ts` + updated the router component fixture to the batched shape. `TestDestinationV1Config` had to become a type alias (not interface) so it carries an implicit index signature the framework's `RouterTransformationRequestData` requires.

shekhar-rudder (commented):
Clean, well-tested change — the version dispatch, `getDestinationVersion` normalization, batching-framework integration, and network-handler composition all check out, and the test fixtures match the code. A few minor notes below: one dead-code cleanup in the dev fixture and two stale comments that describe the bare-`Number()` dispatch the PR's own docs warn against.

src/v0/destinations/test_destination/transform.ts
```diff
@@ -0,0 +1,25 @@
...
+  inputs: TestDestinationRouterRequest[],
```
shekhar-rudder: Since `TEST_DESTINATION` is in `batchedDestinationsMap` (GA), `isBatchingFrameworkEnabled` always returns `true`, so the router goes through `routerTransform.ts` and this `processRouterDest` (plus the `simpleProcessRouterDest` import) is never reached. Suggest removing it and inlining `transformEvent` — the module then only needs to export the per-event `process` that `routerTransform.ts` and the processor consume.

src/v0/destinations/test_destination/transform.test.ts
```diff
@@ -0,0 +1,44 @@
...
+  endpoint: 'https://eu.test-destination.invalid/v1/events',
+  headers: { 'Content-Type': 'application/json', 'X-Api-Key': 'v1-secret-key' },
+  params: {},
+  body: { JSON: message, JSON_ARRAY: {}, XML: {}, FORM: {} },
+  files: {},
+};
+
+describe('test_destination process — integration-major dispatch', () => {
+  // major 0 / undefined / 1 all resolve to Number(...) < 2 and run the v1 path
+  const v1Cases: { name: string; version: number | undefined }[] = [
```
shekhar-rudder, line 32: This comment describes `Number(...) < 2`, but dispatch now goes through `getDestinationVersion(...)`, not a bare `Number()` comparison — and the new CONTRIBUTING/skill docs explicitly warn against branching on bare `Number(event.destination.version)`. Suggest rewording to reference the `getDestinationVersion` normalization (0/undefined/non-numeric → 1) so the comment matches the code it documents.

test/integrations/destinations/test_destination/common.ts
```diff
@@ -0,0 +1,79 @@
...
+  DestinationDefinition: destinationDefinition,
+  Config: { restApiKey: 'v1-secret-key', dataCenter: 'eu' },
+  Enabled: true,
+  WorkspaceID: 'test-workspace-id',
+  Transformations: [],
+  version: 1,
+};
+
+// No version stamped — must dispatch identically to v1 (Number(undefined) => NaN < 2).
+export const destinationNoVersion: Destination = {
```
shekhar-rudder, line 29: `Number(undefined) => NaN < 2` describes the pre-normalization dispatch. Dispatch now normalizes via `getDestinationVersion` (undefined → major 1), so suggest rewording to match — e.g. "undefined normalizes to major 1".

src/v0/destinations/test_destination/transform.ts
saikumarrs: Done in 1103d1b9b — removed `processRouterDest` and the `simpleProcessRouterDest` import. `test_destination` is batching-GA so the legacy router path is unreachable; the module now exports only the per-event `process` (consumed by `routerTransform.ts` and the processor). Kept the `transformEvent`-aliased-to-`process` naming to avoid shadowing the Node `process` global.

src/v0/destinations/test_destination/transform.test.ts
saikumarrs: Done in 1103d1b9b — reworded to reference the `getDestinationVersion` normalization (0 / undefined / non-numeric → 1) instead of the old bare `Number(...) < 2`.

test/integrations/destinations/test_destination/common.ts
saikumarrs: Done in 1103d1b9b — reworded to "getDestinationVersion normalizes undefined to major 1".

src/constants/batchedDestinationsMap.ts
```diff
@@ -4,6 +4,7 @@ export const batchedDestinationsMap: Record<string, true> = {
   POSTHOG: true,
   CUSTOM_AUDIENCE: true,
   ITERABLE_AUDIENCE: true,
+  TEST_DESTINATION: true, // dev-only fixture (INT-6492); only ever reached in dev — gated out of prod at features.ts
 };
```
shekhar-rudder, line 8: The dev-only / INT-6492 explanation is repeated in several places (here, `features.ts`, and the file headers). One canonical spot should be enough — suggest trimming this to a short pointer and keeping the full explanation in a single location.

src/features.ts
```diff
@@ -103,6 +103,9 @@ const defaultFeaturesConfig: FeaturesConfig = {
     CUSTOM_AUDIENCE: true,
     ITERABLE_AUDIENCE: true,
     SURVICATE: true,
+    // dev-only test fixture for definition-versioning (INT-6492); kept out of the prod /features
+    // payload here (NODE_ENV==='production'), so exposure no longer rests solely on control-plane gating.
+    ...(process.env.NODE_ENV !== 'production' && { TEST_DESTINATION: true }),
   },
```
shekhar-rudder, line 109: I don't think there's real harm in dropping the `process.env.NODE_ENV !== 'production'` check here. rudder-server calls the transformer based on the workspace/control-plane config regardless of what's in the `/features` payload, so this gating doesn't actually prevent the destination from being processed — control-plane gating governs exposure either way. Could simplify to just `TEST_DESTINATION: true`

CONTRIBUTING.md
```diff
@@ -945,6 +945,43 @@ And here's an example of the mapping configuration files:
...
+- **Branch `routerTransform`, `deleteUsers`, and the proxy/delivery handler only when a major actually changes them.** Most majors only touch `transform`.
+- **The proxy / dataDelivery route has no destination object** — the major arrives as a top-level **`destinationVersion`** on the proxy request payload. Read it there (e.g. in `networkHandler`'s `proxy`/`responseHandler`), not `destination.version`.
+- **Never route integration majors through `getDestHandler`.** That argument is the architecture version; routing majors there would explode the directory matrix.
+
+See `src/v0/destinations/test_destination` for a worked, dev-only example exercising this idiom across `process`, `processRouterDest`, and `networkHandler`.
+
```
shekhar-rudder, line 984: Stale after the latest commit: `processRouterDest` was removed from `test_destination` (it's batching-GA, so the router runs through `routerTransform.ts`). This now exercises the idiom across `process` and `networkHandler` only — suggest updating the list to drop `processRouterDest`.

shekhar-rudder (commented):
A few follow-ups: consolidate the repeated dev-only note, the NODE_ENV gating in features.ts is likely unnecessary, and one CONTRIBUTING.md line went stale after dropping processRouterDest.

src/features.ts
saikumarrs: Done in e60c08c1f — dropped the gate, back to plain `TEST_DESTINATION: true`. Agree: the /features payload is informational and control-plane workspace config governs whether the destination is processed, so the NODE_ENV check added complexity without actually preventing exposure.

src/constants/batchedDestinationsMap.ts
saikumarrs: Done in e60c08c1f — config.ts is now the single canonical spot; this entry and the features.ts one are trimmed to a short `see .../config.ts` pointer.

CONTRIBUTING.md
saikumarrs: Done in e60c08c1f — dropped `processRouterDest` from the list; noted the router runs through the batching framework (`routerTransform.ts`), which reuses `process`. Also caught a related stale spot in the same sweep: the `v1RequestShape` comment in the router/processor common fixture still claimed the router used it — reworded since the router now asserts the batched-request shape.

shekhar-rudder (approved):

PR #5324, opened 2026-06-29
refactor(batch): add record event dispatch to BatchDestination framework

shekhar-rudder:
## Summary

- **`makeRouterInputSchema` is now a single-variant builder.** It takes one `message` schema plus an optional `destinationConfig` (→ `destination.Config`) and optional `connectionConfig` (→ `connection.config`). The previous multi-variant `variants: [...]` API — which built a `z.union` internally and validated `destinationConfig` outside the union — has been removed. All existing batching destinations (posthog, custom_audience, google_adwords_enhanced_conversions, iterable_audience) are migrated to the flat shape.
- **Union responsibility moved to `VDMV2ObjectDestination`.** Hybrid (record + event-stream) destinations declare two schemas — `recordSchema` and `eventStreamSchema`, each built via `makeRouterInputSchema` with the shared `destinationConfig` — and the framework unions them in `getInputSchema()`. Record events validate the connection config (requires `object`); event-stream events don't, so the record schema is never wrongly applied to them.
- **Adds `VDMV2ObjectDestination`, an object-based record-dispatch layer on top of `BatchDestination`:**
  - `transformObjectRecord(input)` returns a handler map of object type → action → handler for record events; `transformEventStream(input)` returns a message type → handler map for event-stream events.
  - The framework validates object types, actions, and message types automatically before invoking handlers, making the handler map the single source of truth (removes redundant validation from `buildRecordEvent`).
  - Constrains `TConnectionConfig` to `{ destination: { object: string } }` so object-type access is type-checked, moves `MappedToDestination` pre-processing into the framework, and uses `ConfigurationError` for framework-level validation.
- **Migrates CustomerIO** to extend `VDMV2ObjectDestination` with record + event-stream support. Transform-method parameters are typed inline with `z.infer<typeof recordInputSchema>` / `z.infer<typeof eventStreamInputSchema>` so the types are unmistakably schema-derived and cannot drift.
- Documents the onboarding flow and conventions in the `vdm-v2-object-destination` skill.

## Test plan

- [ ] Unit tests pass for `makeRouterInputSchema` (single-variant) and the union built by `VDMV2ObjectDestination`
- [ ] Unit tests pass for the BatchDestination / VDMV2ObjectDestination framework
- [ ] Unit tests pass for CustomerIO record + event-stream transforms and `buildRecordEvent`
- [ ] Integration tests pass for CustomerIO
- [ ] Lint clean

### DIFF
--- a/src/services/destination/nativeBatching/__tests__/batchDestination.test.ts
+++ b/src/services/destination/nativeBatching/__tests__/batchDestination.test.ts
@@ -1,4 +1,5 @@
 import { z } from 'zod';
+import { InstrumentationError } from '@rudderstack/integrations-lib';
 import {
   BatchDestination,
   TransformedEvent,
@@ -6,9 +7,10 @@ import {
   CustomBatchStrategy,
   parseSizeToBytes,
 } from '../batchDestination';
+import { VDMV2ObjectDestination, type RecordInput } from '../vdmV2ObjectDestination';
 import type { BatchStrategy } from '../batchDestination';
 import type { RouterTransformationRequestData } from '../../../../types/destinationTransformation';
-import type { Destination } from '../../../../types/controlPlaneConfig';
+import type { Destination, Connection } from '../../../../types/controlPlaneConfig';
 import type { Metadata, RudderMessage } from '../../../../types/rudderEvents';
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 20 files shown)


src/services/destination/nativeBatching/vdmV2ObjectDestination.ts
```diff
@@ -0,0 +1,132 @@
...
+import get from 'get-value';
+import { ConfigurationError } from '@rudderstack/integrations-lib';
+import type { RouterTransformationRequestData } from '../../../types/destinationTransformation';
+import { MappedToDestinationKey } from '../../../constants';
+import { addExternalIdToTraits, adduserIdFromExternalId } from '../../../v0/util';
+import { BatchDestination } from './batchDestination';
+import type { TransformedEvent } from './types';
+
+// Record message shape known to the framework after schema validation
+export type RecordMessage = {
```
maheshkutty: Can we use [RudderRecordV2](https://github.com/rudderlabs/rudder-transformer/blob/feat/INT-6653-batch-destination-record-dispatch/src/types/rudderEvents.ts#L60)

src/services/destination/nativeBatching/__tests__/batchDestination.test.ts
```diff
@@ -189,3 +279,84 @@ describe('strategy classes', () => {
     expect(result[0].jobIds).toEqual(new Set([1, 2]));
   });
 });
+
+describe('VDMV2ObjectDestination — record dispatch', () => {
```
maheshkutty: Any reason to keep this test as part of this file?
I see we have a test file for vdmV2ObjectDestination

shekhar-rudder: Updated

src/services/destination/nativeBatching/vdmV2ObjectDestination.ts
shekhar-rudder: Updated

maheshkutty (approved):
LGTM

PR #5326, opened 2026-06-30
feat(braze): add v1 network handler for per-event delivery

ItsSudip:

Migrates Braze's delivery network handler from **v0** to **v1**.

- **New `src/v1/destinations/braze/networkHandler.ts`** exporting `prepareProxy`, `proxy`, `processAxiosResponse`, `responseHandler`. Where v0 returns a single status for the entire proxy batch, the v1 `responseHandler` returns one `DeliveryJobState` per event in `DeliveryV1Response.response[]`, correlated by `metadata.jobId`.
- The handler is a **like-for-like port of v0's decision logic**:
  - non-2xx upstream → throw `TransformerProxyError` with one per-job entry at the upstream status code
  - 2xx + `message === 'success'` → per-job `statusCode: 200`
  - 2xx + `message === 'success'` + non-empty `errors[]` → increment the `braze_partial_failure` stat, still treated as delivered
  - 2xx + `message !== 'success'` + non-empty `errors[]` → throw (application-level error)
- **`BrazeResponseHandlerParams`** (in `src/v0/destinations/braze/types.ts`) gains `rudderJobMetadata: ProxyMetdata[]` so the handler can map each job to its result.
- Braze `proxy_v1` requests now route to this handler automatically via the network-handler factory's auto-discovery — no registry edit needed. The v0 handler is left untouched.

**Tests**
- New unit tests (`networkHandler.test.ts`) cover the four response shapes plus `metadata.jobId` correlation.
- `networkHandlerFactory.test.ts`: the existing "v1 request falls back to v0" test used Braze as its example (valid only while Braze had no v1 handler) — repointed to `marketo`, and added a case asserting Braze now resolves to its v1 handler.

Resolves INT-6633

A like-for-like migration to the v1 networkHandler. v0 reports a single outcome for a whole Braze proxy batch; v1 reports one outcome per event keyed by `metadata.jobId`, which is the prerequisite for per-event delivery response handling. Delivery semantics are unchanged.

No change to delivery behaviour. Braze `proxy_v1` requests previously fell back to the v0 handler (wrapped by the v0→v1 adapter in `nativeIntegration.ts`); they now execute the native v1 handler. Per-job verdicts are functionally identical — confirmed by the existing Braze `dataDelivery` component fixtures, which route through the new handler and pass unchanged. The `BrazeResponseHandlerParams` change is additive.

N/A

N/A

The factory auto-discovers v1 handlers, so adding the file activates the v1 path for Braze on deploy. Should a rollback be needed, the built-in v1→v0 fallback restores prior behaviour once the handler is removed.

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist

### DIFF
--- a/src/v1/destinations/braze/networkHandler.ts
+++ b/src/v1/destinations/braze/networkHandler.ts
@@ -1,0 +1,108 @@
+import { prepareProxyRequest, proxyRequest } from '../../../adapters/network';
+import { getDynamicErrorType, processAxiosResponse } from '../../../adapters/utils/networkUtils';
+import { TransformerProxyError } from '../../../v0/util/errorTypes';
+import { TAG_NAMES } from '../../../v0/util/tags';
+import { isHttpStatusSuccess } from '../../../v0/util/index';
+import stats from '../../../util/stats';
+import type { DeliveryJobState, DeliveryV1Response, ProxyMetdata } from '../../../types';
+import { BrazeResponseHandlerParams } from '../../../v0/destinations/braze/types';
+
+const DESTINATION = 'braze';
+
+/**
+ * Typed subset of the Braze REST API response body. The `message` field
+ * signals overall success ('success') or failure. The `errors` array is
+ * present on both partial failures (message='success') and full application-
+ * level errors (message!='success').
+ */
+type BrazeApiResponse = {
+  message?: string;
+  errors?: unknown[];
+};
+
+function isBrazeApiResponse(value: unknown): value is BrazeApiResponse {
+  return typeof value === 'object' && value !== null;
+}
+
+/**
+ * Maps every job in `rudderJobMetadata` to a `DeliveryJobState` using the
+ * same `error` string and the given `statusCode`. The `error` field is the
+ * JSON-serialised Braze response body so downstream consumers can inspect it.
+ * `JSON.stringify(undefined)` returns `undefined` at runtime despite the
+ * TypeScript return type, so the `?? ''` guard ensures we never emit
+ * `error: undefined`.
+ */
+const buildJobStates = (
+  response: unknown,
+  statusCode: number,
+  rudderJobMetadata: ProxyMetdata[],
+): DeliveryJobState[] =>
+  rudderJobMetadata.map((metadata) => ({
+    statusCode,
+    metadata,
+    error: JSON.stringify(response) ?? '',
+  }));
+
+const responseHandler = (params: BrazeResponseHandlerParams): DeliveryV1Response => {
+  const { destinationResponse, rudderJobMetadata } = params;
+  const { response, status } = destinationResponse;
+
+  // Guard 1: non-2xx HTTP status — destination rejected the request entirely
+  if (!isHttpStatusSuccess(status)) {
+    throw new TransformerProxyError(
+      `Request failed for ${DESTINATION} with status: ${status}`,
+      status,
+      { [TAG_NAMES.ERROR_TYPE]: getDynamicErrorType(status) },
+      destinationResponse,
+      '',
+      buildJobStates(response, status, rudderJobMetadata),
+    );
+  }
+
+  const responseBody: BrazeApiResponse | null = isBrazeApiResponse(response) ? response : null;
+  const errors = responseBody?.errors;
+  const hasErrors = Array.isArray(errors) && errors.length > 0;
+  const brazeMessage = responseBody?.message;
+
+  // Guard 2: partial failure — destination accepted the request but some items
+  // within the batch were invalid. Braze signals this with message='success'
+  // and a non-empty errors array. Emit a metric so ops can track frequency;
+  // the batch is still considered delivered (fall through to success).
+  if (brazeMessage === 'success' && hasErrors) {
+    stats.increment('braze_partial_failure');
+  }
+
+  // Guard 3: application-level error — destination returned 2xx but with an
+  // error message (message!='success') and errors, meaning the entire request
+  // was rejected at the application layer despite the transport succeeding.
+  if (brazeMessage !== 'success' && hasErrors) {
+    throw new TransformerProxyError(
+      `Request failed for ${DESTINATION} with status: ${status}`,
+      status,
+      { [TAG_NAMES.ERROR_TYPE]: getDynamicErrorType(status) },
+      destinationResponse,
+      '',
+      buildJobStates(response, status, rudderJobMetadata),
+    );
+  }
+
+  return {
+    status,
+    message: `Request for ${DESTINATION} Processed Successfully`,
+    response: buildJobStates(response, status, rudderJobMetadata),
+  };
+};
+
+function networkHandler(this: {
+  responseHandler: typeof responseHandler;
+  proxy: typeof proxyRequest;
+  prepareProxy: typeof prepareProxyRequest;
+  processAxiosResponse: typeof processAxiosResponse;
+}) {
+  this.responseHandler = responseHandler;
+  this.proxy = proxyRequest;
+  this.prepareProxy = prepareProxyRequest;
+  this.processAxiosResponse = processAxiosResponse;
+}
+
+export { networkHandler, responseHandler };
--- a/src/v1/destinations/braze/networkHandler.ts
+++ b/src/v1/destinations/braze/networkHandler.ts
@@ -10,21 +10,6 @@ import { BrazeResponseHandlerParams } from '../../../v0/destinations/braze/types
 const DESTINATION = 'braze';
 
 /**
- * Typed subset of the Braze REST API response body. The `message` field
- * signals overall success ('success') or failure. The `errors` array is
- * present on both partial failures (message='success') and full application-
- * level errors (message!='success').
- */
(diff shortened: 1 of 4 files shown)


src/v1/destinations/braze/networkHandler.ts
```diff
@@ -0,0 +1,108 @@
...
+ * level errors (message!='success').
+ */
+type BrazeApiResponse = {
+  message?: string;
+  errors?: unknown[];
+};
+
+function isBrazeApiResponse(value: unknown): value is BrazeApiResponse {
+  return typeof value === 'object' && value !== null;
+}
```
achettyiitr: BrazeApiResponse / isBrazeApiResponse are redundant. BrazeResponseHandlerParams already types response as { message?; errors? }, so the extra type and guard don't add any compile-time safety — they only guard against null/non-object at runtime. Reasonable as defensive code, just know it's not buying you type checking you didn't already have.

src/v1/destinations/braze/networkHandler.ts
```diff
@@ -0,0 +1,108 @@
...
+      destinationResponse,
+      '',
+      buildJobStates(response, status, rudderJobMetadata),
+    );
+  }
+
+  const responseBody: BrazeApiResponse | null = isBrazeApiResponse(response) ? response : null;
+  const errors = responseBody?.errors;
+  const hasErrors = Array.isArray(errors) && errors.length > 0;
+  const brazeMessage = responseBody?.message;
```
achettyiitr: Can't we do something like this as responseBody would be nil in case isBrazeApiResponse returns nil?


 ```
 if (isBrazeApiResponse(response)) {
    const errors = response.errors;
    const hasErrors = Array.isArray(errors) && errors.length > 0;
    const brazeMessage = response.message;

    if (brazeMessage === 'success' && hasErrors) {
      stats.increment('braze_partial_failure');
    }

    if (brazeMessage !== 'success' && hasErrors) {
      throw new TransformerProxyError(
        `Request failed for ${DESTINATION} with status: ${status}`,
        status,
        { [TAG_NAMES.ERROR_TYPE]: getDynamicErrorType(status) },
        destinationResponse,
        '',
        buildJobStates(response, status, rudderJobMetadata),
      );
    }
  }
```

File: src/v1/destinations/braze/networkHandler.ts [hunk 3]
```diff
@@ -0,0 +1,108 @@
...
+  // and a non-empty errors array. Emit a metric so ops can track frequency;
+  // the batch is still considered delivered (fall through to success).
+  if (brazeMessage === 'success' && hasErrors) {
+    stats.increment('braze_partial_failure');
+  }
+
+  // Guard 3: application-level error — destination returned 2xx but with an
+  // error message (message!='success') and errors, meaning the entire request
+  // was rejected at the application layer despite the transport succeeding.
+  if (brazeMessage !== 'success' && hasErrors) {
```
achettyiitr: Will this be a breaking change w.r.t the existing implenmentation?

src/v1/destinations/braze/networkHandler.ts
```diff
@@ -0,0 +1,108 @@
...
+const buildJobStates = (
+  response: unknown,
+  statusCode: number,
+  rudderJobMetadata: ProxyMetdata[],
+): DeliveryJobState[] =>
+  rudderJobMetadata.map((metadata) => ({
+    statusCode,
+    metadata,
+    error: JSON.stringify(response) ?? '',
+  }));
```
achettyiitr, line 29: error field carries the full JSON body on success (statusCode: 200). buildJobStates sets error: JSON.stringify(response) for delivered jobs. customerio's v1 handler uses error: 'success' instead. Is it expected? Even for other network handler's I see https://github.com/rudderlabs/rudder-transformer/blob/feature/int-6633-rudder-transformer-v0-to-v1-network-handler-migration-for/src/v1/destinations/am/networkHandler.ts#L40-L44

src/v1/destinations/braze/networkHandler.ts
```diff
@@ -0,0 +1,108 @@
...
+  const hasErrors = Array.isArray(errors) && errors.length > 0;
+  const brazeMessage = responseBody?.message;
+
+  // Guard 2: partial failure — destination accepted the request but some items
+  // within the batch were invalid. Braze signals this with message='success'
+  // and a non-empty errors array. Emit a metric so ops can track frequency;
+  // the batch is still considered delivered (fall through to success).
+  if (brazeMessage === 'success' && hasErrors) {
+    stats.increment('braze_partial_failure');
+  }
```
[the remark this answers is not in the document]
ItsSudip: Thanks for flagging — this isn't a regression introduced by this PR.

The existing **v0** handler behaves identically: on a 2xx with `message === 'success'` and a non-empty `errors[]`, it increments the `braze_partial_failure` stat and treats the whole batch as delivered — it does not map `errors[*].index` back onto individual jobs either (see the partial-errors branch in `src/v0/destinations/braze/networkHandler.ts`).

This PR is a like-for-like **v0 → v1 network handler migration**. It introduces the per-event `response[]` plumbing (one `DeliveryJobState` per `metadata.jobId`), but deliberately preserves v0's delivery verdicts — the acceptance criterion is explicitly "no behavioural change vs v0". The existing component fixture `braze_v1_scenario_2` asserts exactly this: all jobs return `statusCode: 200` on a partial-failure response.

Attributing indexed partial failures to individual jobs is a genuine gap, but it is pre-existing in v0 and is being handled as separate, follow-up work — not by marking those jobs failed, which would change their terminal state and pull in retry/ordering side-effects. Out of scope for this migration.

src/v1/destinations/braze/networkHandler.ts
ItsSudip: No — the output is unchanged.

The class did change (`NetworkError` → `TransformerProxyError`), and that's required: `TransformerProxyError` is what carries the per-job `response` array (`this.response = response`), which is the whole point of a v1 handler. It's the same class Iterable and the other v1 handlers use.

But the v1 failure response is assembled by `handlevV1DeliveriesFailureEvents`, which derives every field from inputs we pass identically in both versions — per-job `error` from `error.destinationResponse.response`, `statusCode` from `errObj.status`, and `statTags` from the same `{ errorType: getDynamicErrorType(status) }` we hand in. `authErrorCategory` is falsy in both (v0 passed nothing, v1 passes `''`), so it's omitted either way.

Empirically: the existing `dataDelivery` fixtures assert the full `statTags`, `message`, `status` and per-job entries for 400/401/500/503/504/null-response, and they all pass **unchanged** — no fixture edits in this PR. Those same fixtures previously ran through v0's `NetworkError`.

src/v1/destinations/braze/networkHandler.ts
ItsSudip: Intentional — this preserves Braze's current behaviour exactly.

The existing v0→v1 adapter in `nativeIntegration.ts` already sets `error: JSON.stringify(destinationResponse.response)` for every job, successes included, so that's what Braze emits in production today. The pre-existing component fixtures assert it directly — `braze_v1_scenario_1` expects `error` to be `JSON.stringify({ events_processed: 2, purchases_processed: 1, message: 'success' })` on each delivered job. Emitting `'success'` instead would mean rewriting those fixtures, i.e. a behavioural change, and this PR is a like-for-like migration ("no behavioural change vs v0").

You're right that there's an inconsistency across handlers — AM and CustomerIO collapse it to `'success'` while Braze (and the generic v0→v1 adapter) carry the destination response body. That divergence pre-dates this PR. Worth noting too that the response body is what gives downstream consumers the destination's actual reply for a delivered event, which is useful for debugging; dropping it is lossy. Happy to standardise this across handlers as a separate, deliberate change if we want one convention.

src/v1/destinations/braze/networkHandler.ts
ItsSudip: Good catch — agreed, removed both. `BrazeResponseHandlerParams` already types `response` as `{ message?; errors? }`, so `BrazeApiResponse` just restated it, and `isBrazeApiResponse` was guarding a case optional chaining already handles (`response?.message` is safe for `null`/`undefined`, and property access on a primitive yields `undefined` rather than throwing). Now reading `response?.errors` / `response?.message` directly — same behaviour, 16 fewer lines.

src/v1/destinations/braze/networkHandler.ts
ItsSudip: Took this a step further, following your other comment: rather than wrapping the checks in `if (isBrazeApiResponse(response)) { … }`, I dropped the guard *and* the `BrazeApiResponse` type entirely — `BrazeResponseHandlerParams` already types `response`, so there's nothing left to narrow. The block is now:

```ts
const errors = response?.errors;
const hasErrors = Array.isArray(errors) && errors.length > 0;
const brazeMessage = response?.message;
```

Outcomes match your version exactly: for a null/non-object response, `brazeMessage` is `undefined` and `hasErrors` is `false`, so both guards fall through to the success return.

achettyiitr (approved):

PR #5331, opened 2026-07-01
fix(customerio): event stream events rejected by record-only connection schema

shekhar-rudder:
## Summary
- The `getV2InputSchema()` Zod schema applied `CustomerIOConnectionConfigSchema` (which requires `object: 'person'|'event'`) to **all** events, not just record messages. Event stream messages (identify, track, page, etc.) arriving with a connection object were rejected because their connection config lacks the record-specific `object` field.
- Split the input schema into a `z.union` of two variants: record messages validate connection config against the record-specific schema; event stream messages skip connection validation entirely.
- Added table-driven tests covering event stream messages with no destination object (VDM v1), nil config, and no connection.

## Test plan
- [x] Unit test proves the bug: event stream message with `config: {}` connection was rejected before the fix
- [x] All 9 unit tests pass (`routerTransform.test.ts`)
- [x] All 62 Customer.io unit tests pass
- [x] All 121 Customer.io integration tests pass (processor + router + dataDelivery)

### DIFF
--- a/src/v0/destinations/customerio/routerTransform.test.ts
+++ b/src/v0/destinations/customerio/routerTransform.test.ts
@@ -108,6 +108,48 @@ describe('CustomerIOIntegration — record event routing', () => {
     );
   });
 
+  it.each([
+    {
+      name: 'no destination object (VDM v1)',
+      connection: {
+        sourceId: 'src-1',
+        destinationId: 'dest-1',
+        enabled: true,
+        config: {},
+      },
+    },
+    {
+      name: 'connection with nil config',
+      connection: {
+        sourceId: 'src-1',
+        destinationId: 'dest-1',
+        enabled: true,
+        config: null,
+      },
+    },
+    {
+      name: 'no connection at all',
+      connection: undefined,
+    },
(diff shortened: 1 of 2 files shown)


src/v0/destinations/customerio/routerTransform.test.ts
```diff
@@ -108,6 +108,48 @@ describe('CustomerIOIntegration — record event routing', () => {
...
+      connection: {
+        sourceId: 'src-1',
+        destinationId: 'dest-1',
+        enabled: true,
+        config: null,
+      },
+    },
+    {
+      name: 'no connection at all',
+      connection: undefined,
```
krishna2020, line 132: Is this scenario possible ? For eventstream connection, at least sourceId & destinationId will be present

shekhar-rudder: Yes this scenario is not possible. But since event stream code doesn't depend on connection object being present, we should treat this case as a valid one and not throw an error

krishna2020 (approved):

saikumarrs (approved):

maheshkutty (approved):

PR #5341, opened 2026-07-07
fix(customerio): fall back to track event for device events missing id/token in v2

maheshkutty:

Align v2 device-event handling with v0: an Application Installed / deviceTokenEventName event that is missing userId/email or the device token now degrades to a normal track event instead of throwing. The guard moves into deviceActionFor, which returns null so the envelope routes through buildTrack. delete_device still throws, matching v0.

Adds a combined router test covering both fallback paths (anonymous_id and id identifiers) in a single batched request.

[Resolves INT-6710](https://linear.app/rudderstack/issue/INT-6710/customerio-fix-device-event-payload-creation-for-v2-implementation)

N/A

N/A

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist

### DIFF
--- a/src/v0/destinations/customerio/v2/util.ts
+++ b/src/v0/destinations/customerio/v2/util.ts
@@ -210,6 +210,7 @@ export const buildDevice = (
 };
 
 const deviceActionFor = (
+  message,
   evName: string,
   destination: CustomerIODestination,
 ): 'add_device' | 'delete_device' | null => {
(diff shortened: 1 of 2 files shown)


src/v0/destinations/customerio/v2/util.ts
```diff
@@ -218,7 +219,19 @@ const deviceActionFor = (
...
   }
-  return evName === DEVICE_DELETE_EVENT_NAME ? 'delete_device' : 'add_device';
+  if (evName === DEVICE_DELETE_EVENT_NAME) {
+    return 'delete_device';
+  }
+  // add_device: mirror v0 behaviour — a device-register event with a missing
+  // userId/email or device token degrades to a normal track event rather than
+  // building a device payload.
+  const id =
+    getFieldValueFromMessage(message, 'userIdOnly') || getFieldValueFromMessage(message, 'email');
```
shekhar-rudder: The `id`/`token` derivation here duplicates the identical block in `buildDevice` (the `userIdOnly || email` lookup plus `context.device.token`). Consider extracting a small `getDeviceCredentials(message)` helper and using it in both places. Note that `buildDevice`'s guard is still needed for the `delete_device` path (which returns before this check), so this is a readability/DRY cleanup rather than dead-code removal.

shekhar-rudder (commented):
Reviewed the v2 device-event fallback change. The logic faithfully mirrors v0: `delete_device` still throws on missing id/token, while a register event with a missing userId/email or token now degrades to a track event. Tests cover both fallback triggers. One minor DRY note inline.

src/v0/destinations/customerio/v2/util.ts
maheshkutty: Done

shekhar-rudder (approved):

PR #5348, opened 2026-07-08
fix(security): sandbox tenant-config custom-mapping evaluation in isolated-vm [INT-6725]

shekhar-rudder:
## Summary

Fixes **INT-6725** (parent INT-6708): tenant-supplied destination "custom mappings" were evaluated in-process by `JsonTemplateEngine` via the `Function` constructor with **no sandbox**, allowing SSTI → RCE (`Function(...)`, `require('child_process').execSync(...)`) and **full environment disclosure** (`process.env||$.userId` dumps every env var — verified). Affected tenant-config paths: HTTP `headers`/`queryParams`/`propertiesMapping` and GA4 v2 `eventProperties`.

This routes **all** such evaluation through an `isolated-vm` sandbox where `process`/`require`/host globals are unreachable, so the attack payloads resolve to a harmless fallback (or fail closed) and leak nothing.

## Approach

- **Centralized chokepoint:** `applyCustomMappings` / `applyJSONStringTemplate` (`src/v0/util/index.js`) are now **async** and delegate to a sandboxed client. Every current and future caller inherits the sandbox.
- **isolated-vm containment:** reuses the `custom_audience` pattern — an esbuild IIFE bundle of the engine (`dist/mappingsSandbox.bundle.js`) loaded into a **dedicated, per-`workspaceId`** isolate pool (`custom_mappings_ivm`), separate from custom_audience and sized lower (16 MB, `maxSize` 20, 500 ms exec). Isolates are created lazily and self-evict via LRU + TTL. `IvmScriptRunner` was promoted to `src/util/ivm/scriptRunner.ts`.
- **Fail closed:** a falsy `workspaceId` throws — never a shared/default isolate (no cross-tenant state).
- **Feature flag:** `CUSTOM_MAPPINGS_SANDBOX_ENABLED` (default on; `=false` reverts to byte-equivalent in-process eval, read per-call) — instant rollback without redeploy.
- **process stub:** the bundle sets `globalThis.process = {}` so `process.env.X` resolves to `undefined` → graceful fallback (deterministic; the real guarantee is V8 isolation + structured-clone-only marshaling — an adversarial review confirmed no host-data escape).

## Security verification

- Unit + component tests assert the INT-6725 payloads (`process.env||$.userId`, `process.env.SECRET||$.userId`, `Function('return process.env')()`, `({}).constructor.constructor('return process.env')()`, `Function('return this')()`) leak **no** planted host secret; `require` is unreachable.
- Focused adversarial security review: **verdict SECURE** (no escape via Function/constructor-chain/eval/`this`; structured-clone boundary rejects returning host-ish objects).

## Env vars (documented in `docs/features.md`)

`CUSTOM_MAPPINGS_SANDBOX_ENABLED`, `CUSTOM_MAPPINGS_IVM_MEMORY_MB` (16), `CUSTOM_MAPPINGS_IVM_CACHE_MAX_SIZE` (20), `CUSTOM_MAPPINGS_IVM_CACHE_TTL_MS` (300000), `CUSTOM_MAPPINGS_IVM_INIT_TIMEOUT_MS` (5000), `CUSTOM_MAPPINGS_IVM_EXEC_TIMEOUT_MS` (500). Observability via existing `ivm_cache_*` metrics tagged `cache:custom_mappings_ivm`.

## Testing

`build:ci` PASS · sandbox/runner/GA4/custom_audience TS suites 108/108 · chokepoint + `index.test.js` 196/196 · HTTP component 17/17 (incl. INT-6725 regression) · GA4 v2 component 11/11.

## Notes / follow-ups

- The `rudder-integrations-config` `from`/`path` regex stays as defense-in-depth but is **no longer** the security boundary — no regex change needed.
- `applyJSONStringTemplate` currently has **no production caller** (reachable only via tests) — flag for follow-up: wire the intended caller or remove.
- Deferred minors: relocate `docs/features.md` per org convention; startup bundle-existence check.


src/util/customMappings/sandboxClient.test.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    // Vector 2: the classic ({}).constructor.constructor('return process.env')() breakout.
+    // process is undefined inside the isolate, so no host env is reachable and the planted
+    // secret never appears in the output.
+    const ctorOut = await sandboxedApplyCustomMappings(
+      { userId: 'u1' },
+      [{ from: "({}).constructor.constructor('return process.env')() || $.userId", to: 'g' }],
+      WS,
+    );
+    expect(JSON.stringify(ctorOut)).not.toContain(SECRET);
```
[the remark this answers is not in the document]
shekhar-rudder: Good catch — done in 8585ce12b. Vector 2 currently *resolves* (the engine swallows the inner `process is not defined` throw and falls back to `$.userId`), so the direct `await` doesn't reject today, but I wrapped it in the same try/catch capture as Vector 1 so the no-secret assertion always runs regardless of whether the vector resolves or rejects. Tests remain green.

package.json
```diff
@@ -27,8 +27,8 @@
     "start": "node dist/src/index.js",
     "start:dev": "nodemon",
     "build:start": "npm run build && npm run start",
-    "build:ci": "tsc -p tsconfig.json && npm run build:custom-audience-sandbox",
-    "build:custom-audience-sandbox": "esbuild src/v0/destinations/custom_audience/template/templateEngineSandbox.ts --bundle --format=iife --platform=neutral --main-fields=main,module --outfile=dist/templateEngineSandbox.bundle.js",
+    "build:ci": "tsc -p tsconfig.json && npm run build:sandboxes",
+    "build:sandboxes": "esbuild src/v0/destinations/custom_audience/template/templateEngineSandbox.ts --bundle --format=iife --platform=neutral --main-fields=main,module --outfile=dist/templateEngineSandbox.bundle.js && esbuild src/util/customMappings/mappingsSandbox.ts --bundle --format=iife --platform=neutral --main-fields=main,module --outfile=dist/mappingsSandbox.bundle.js",
```
saikumarrs: There must be a better way than specifying input/output files configuration of esbuild within the script. I'm hoping a configuration file.
The script is not readable and risks maintenance.

saikumarrs (dismissed):

package.json
shekhar-rudder: Good call — the real issue was the duplicated flag set across two chained `esbuild` commands, which every new sandbox would copy again.

Since esbuild has no native config-file concept (unlike webpack), I collapsed both entry points into a single invocation and switched `--outfile` → `--outdir` with `--entry-names=[name].bundle`. The shared flags are now stated once, and adding a sandbox is just appending its entry path.

Output filenames are unchanged (`dist/templateEngineSandbox.bundle.js`, `dist/mappingsSandbox.bundle.js`), so the `test`/`test:ts`/`build:ci` consumers and runtime loaders are unaffected — verified by running `npm run build:sandboxes`. Done in dbf8127bc.

saikumarrs (dismissed):

saikumarrs (approved):

PR #5349, opened 2026-07-09
chore: migrate GAEC destination from JavaScript to TypeScript

ItsSudip:

Migrates the `google_adwords_enhanced_conversions` (GAEC) v0 destination from JavaScript to TypeScript:

- **`config.js` → `config.ts`**, **`transform.js` → `transform.ts`**, **`networkHandler.js` → `networkHandler.ts`** (with `networkHandler.test.js` → `.ts`)
- **New `types.ts`** with the destination's typed shapes: `GaecConfig`, `GaecPayload`, `ConversionAdjustment`/`UserIdentifierEntry`/`AddressInfo`, `GaecDeliveryRequest`, and shared-envelope aliases `GaecDestination` / `GaecRouterRequest` (built on `Destination<T>` / `RouterTransformationRequestData` from `src/types`)
- **Type guards instead of `as` casts or non-null assertions** at every JS-interop boundary (`isGaecPayload`, `isGaecDestination`, `isSdkErrorResponse`, `isPartialFailureBody`, `isMappingElement`)
- **ES imports only** (no `require`), which surfaces the real types of `@rudderstack/integrations-lib`'s `GoogleAdsSDK` — the networkHandler adapts to them: minimal method-syntax boundary interface for the SDK instance, `SdkResponse.responseBody` made optional (client-error responses lack it), numeric `loginCustomerId` stringified for the SDK's `string` config field, and a thin `requestLog`/`responseLog` adapter bridging the SDK's optional log params to `logger.js`'s required destructures
- `processRouterDest` typed as `(inputs: GaecRouterRequest[], reqMetadata) => Promise<RouterTransformationResponse[]>`
- Cleanups: `ENHANCEMENT`/`RESTATEMENT` hoisted to constants, `flag` sentinel replaced with `isConfiguredConversion`, file-level `eslint-disable` narrowed to one targeted line
- 4 new negative unit tests: `getConversionActionId` null / client-error / application-error branches and a non-2xx `responseHandler` case

Resolves INT-6744

Bring GAEC up to the codebase's TypeScript conventions: typed config/payload/delivery shapes, guard-based narrowing at untyped boundaries (no `as`, no `!`, no `any`), and reuse of the shared router envelope types — so future changes to this destination land on typed code.

Happy-path output is byte-identical (all 30 component tests pass against unchanged expectations). Degenerate-input paths now fail cleanly instead of crashing:

- Malformed router event (non-object `metadata`/`message`/`destination`), a null `constructPayload` result, or a non-string `message.type` now throw `InstrumentationError` (400) where the JS crashed with a `TypeError` (500)
- `loginCustomerId` is set in `params`/headers only when `subAccount` is true — inert end-to-end, since the proxy path already gated on `params.subAccount` before using it

N/A

N/A — all changes are scoped to the destination's directory.

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist


src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
```diff
@@ -0,0 +1,233 @@
...
+): GaecDeliveryRequest => {
+  const { Config } = destination;
+  const { event } = message;
+  const { listOfConversions, adjustmentType } = Config;
+  const isConfiguredConversion = listOfConversions.some((i) => i.conversions === event);
+  if (!event || !isConfiguredConversion) {
+    throw new ConfigurationError(
+      `Conversion named "${String(event)}" was not specified in the RudderStack destination configuration`,
+    );
+  }
```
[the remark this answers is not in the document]
ItsSudip: Addressed

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
```diff
@@ -0,0 +1,234 @@
+import get from 'get-value';
```
shekhar-rudder, line 1: Can we update the PR so that file rename doesn't show up as delete + new file?

ItsSudip: GitHub shows a delete+add pair as a rename only when ~50%+ of the file's content is identical, computed on the final PR diff — it can't be forced via commit metadata. In the hs migration (#4960) the conversion was annotation-level, so every file stayed above the bar (`transform.js → ts` kept 65% of its content, `config` 96%). Here the rewrite goes deeper (typed signatures, type guards at the JS-interop boundaries, ES imports, restructured request assembly), so `transform.ts` retains only 24% of the old content and `networkHandler.ts` 39% — below the threshold, hence delete + new file. `config.ts` (60%) and `networkHandler.test.ts` (53%) cleared it and do show as renames.

To make this reviewable, the branch is now split into: (1) a pure `git mv` rename commit (100% similarity, zero content change), (2) the TS migration as in-place edits to the `.ts` files, (3) the review fix. Reviewing commit-by-commit shows proper line diffs instead of a new file.

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
```diff
@@ -0,0 +1,234 @@
...
+
+  if (requireHash === false) {
+    updatedMapping = updateMappingJson(updatedMapping);
+  }
+
+  // `constructPayload` returns null when nothing in the message is mappable; the guard also
+  // narrows its loosely inferred return to the typed payload.
+  const payload = constructPayload(message, updatedMapping);
+  if (!isGaecPayload(payload)) {
+    throw new InstrumentationError('Payload could not be constructed from the message');
```
shekhar-rudder: Per the repo's migration guide (`memory-bank/23_js_to_ts_migration.md`, #6/#8 + checklist #5), preserve the original's implicit non-null trust with `!` rather than adding an if-throw. The original did `const payload = constructPayload(...)` then `payload.partialFailure = true` — a `null` return would surface as a `TypeError` (500). This guard throws `InstrumentationError` (400) instead, changing both the error type and the HTTP status. Suggest `const payload = constructPayload(message, updatedMapping)!;` and dropping the `isGaecPayload` guard.

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
```diff
@@ -0,0 +1,234 @@
...
+
+const isGaecDestination = (value: unknown): value is { Config: GaecConfig } =>
+  isRecord(value) && isRecord(value.Config);
+
+// `event` arrives loosely typed from the Zod passthrough in routerTransform.ts; narrow its
+// fields with type guards before handing them to the strictly typed processEvent.
+const process = (event: Record<string, unknown>): GaecDeliveryRequest => {
+  const { metadata, message, destination } = event;
+  if (!isRecord(metadata) || !isRecord(message) || !isGaecDestination(destination)) {
+    throw new InstrumentationError(
```
shekhar-rudder: The original was a one-line delegate: `const process = (event) => processEvent(event.metadata, event.message, event.destination)`. Adding this shape validation converts a would-be `TypeError` (500, **retryable**) into `InstrumentationError` (400, **non-retryable**) for malformed events — a behavior change with real retry-semantics impact, outside migration scope (guide #1, #6). Prefer typing the delegate via the input types over a runtime guard.

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
```diff
@@ -0,0 +1,234 @@
...
+    Authorization: `Bearer ${accessToken}`,
+    'Content-Type': JSON_MIME_TYPE,
+  };
+  const params: GaecDeliveryRequest['params'] = {
+    // `event` was validated as a configured conversion name (a string) in processTrackEvent.
+    event: typeof event === 'string' ? event : '',
+    customerId: filteredCustomerId,
+    accessToken,
+    loginCustomerId: undefined,
+    subAccount,
```
shekhar-rudder, line 96: Two deviations from checklist #10 (migrated body should stay byte-near-identical): (a) `responseBuilder` was rewritten from in-place mutation of `deliveryRequest` into a spread-literal assembly; (b) `params.loginCustomerId` is now only populated when `subAccount` is true, whereas the original always set it via `removeHyphens(loginCustomerId)`. Even if inert downstream, both are changes beyond adding types.

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
```diff
@@ -0,0 +1,234 @@
...
+  const newMapping: unknown[] = [];
+  mapping.forEach((element) => {
+    if (
+      isMappingElement(element) &&
+      get(element, 'metadata.type') &&
+      element.metadata?.type?.includes('hashToSha256')
+    ) {
+      // eslint-disable-next-line no-param-reassign -- intentional in-place mutation: this function's purpose is to rewrite metadata.type
+      element.metadata = { ...element.metadata, type: 'toString' };
+    }
```
shekhar-rudder, line 57: Minor: the original mutated in place — `element.metadata.type = 'toString'`. This replaces the whole `metadata` object. Output is equivalent, but it changes copy/reference semantics — the kind of rewrite checklist #10 asks migrations to avoid. Could keep the in-place assignment with a targeted `eslint-disable-next-line` (same pattern already used a few lines below).

src/v0/destinations/google_adwords_enhanced_conversions/networkHandler.ts
```diff
@@ -0,0 +1,296 @@
...
+    {
+      httpClient: {
+        // `statsClient` was never exported by util/stats, so the legacy `require` destructure
+        // always wired `undefined` here; kept explicit to preserve the SDK httpClient shape.
+        statsClient: undefined,
+        // logger.js's `requestLog`/`responseLog` destructure their second argument and treat
+        // `metadata` as required, while the SDK's IHttpLogger declares both as optional;
+        // adapt so a no-data or no-metadata call can't crash the logger.
+        logger: {
+          ...logger,
```
shekhar-rudder: The original passed `logger` straight to the SDK's `httpClient`. This introduces a runtime adapter that wraps `requestLog`/`responseLog` with default-filling logic — new runtime behavior, not a type annotation (guide #1). Since `logger` is imported from a `.js` file it infers as `any` (guide #5) and is assignable to the SDK's logger param directly. Can the adapter be dropped in favor of passing `logger` as before?

src/v0/destinations/google_adwords_enhanced_conversions/networkHandler.ts
```diff
@@ -0,0 +1,296 @@
...
+    addConversionAdjustMent(payload: GaecPayload): Promise<SdkResponse>;
+  } = new GoogleAdsSDK.GoogleAds(
+    {
+      accessToken: params.accessToken,
+      customerId: params.customerId,
+      // in-flight payloads built by the legacy JS transformer may carry a numeric
+      // loginCustomerId; the SDK config field is typed string
+      loginCustomerId:
+        params.subAccount && params.loginCustomerId ? String(params.loginCustomerId) : '',
+      developerToken: getDeveloperToken(),
```
shekhar-rudder, line 145: Minor: original was `loginCustomerId: params.subAccount ? params.loginCustomerId : ''`. The `String(...)` is a reasonable type fix, but the added `&& params.loginCustomerId` changes behavior when `subAccount` is true and `loginCustomerId` is falsy (original forwarded the falsy value; this sends `''`).

shekhar-rudder (commented):
This is a clean, well-typed result, but it reads as a **refactor + behavior change** rather than the type-only migration the repo's guide (`memory-bank/23_js_to_ts_migration.md`) prescribes. Key principles it runs against: **#1 Migration ≠ Refactoring**, **#6 prefer `!` over if-throw validation** (non-null assertions are explicitly *allowed*), and **checklist #10 — the migrated body should be byte-near-identical** apart from type annotations. The PR description itself lists behavior changes (400 vs 500, guards replacing `!`) as features — but those are exactly what the guide asks a migration to avoid. Specific spots flagged inline; happy to be corrected if the guide's approach was intentionally superseded here.

FYI, verified as fine (not flagged): the `statsClient: undefined` claim is accurate (stats.js never exported it), and passing `undefined` as the 4th `simpleProcessRouterDest` arg matches the guide's "pass undefined explicitly" rule.

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
ItsSudip: Fixed — replaced the `isGaecPayload` if-throw with the guide's `!` pattern: `const payload: GaecPayload = constructPayload(message, updatedMapping)!;`. A `null` return now surfaces exactly like the original (TypeError → 500, retryable). `GaecPayload`'s fields were made optional so the untyped return assigns directly, and the follow-up accesses use `!` to keep the original's implicit non-null trust.

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
ItsSudip: Fixed — `process` is back to the one-line delegate, typed via the input types instead of runtime guards. One note: the router-input schema only validates `message` (the rest of the envelope is `.passthrough()`), so `metadata`/`destination` are statically optional on the param type and the delegate uses `!` — type-only, so malformed events keep the original TypeError → 500 (retryable) semantics.

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
ItsSudip: Fixed both — restored the original in-place mutation of `deliveryRequest` (the payload/params types now have optional fields so `defaultRequestConfig()`'s empty slots assign directly), and `params.loginCustomerId` is always set via `removeHyphens(loginCustomerId)` again, with the `login-customer-id` header still gated on `subAccount`. Statement order and the `loginCustomerId && !isString(...)` condition match the original.

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
ItsSudip: Fixed — restored the in-place `element.metadata.type = 'toString'` with the targeted eslint-disable, no object replacement.

src/v0/destinations/google_adwords_enhanced_conversions/networkHandler.ts
ItsSudip: Adapter dropped — `logger` is passed straight through as in the original. One correction to the premise, though: `logger` does *not* infer as `any` — TS infers the real signatures from logger.js's implementation, and `requestLog`'s required (destructured) second argument is incompatible with `IHttpLogger`'s optional one, so the direct pass didn't compile; that's why the adapter existed. Resolved with type-only JSDoc `@type` annotations on `requestLog`/`responseLog` in logger.js (zero runtime change — the guide's "modify the source .js only if absolutely necessary and safe" case). Runtime behavior is now byte-identical to the original, including the original's implicit assumption that the SDK always passes log data.

src/v0/destinations/google_adwords_enhanced_conversions/networkHandler.ts
ItsSudip: Fixed to `params.subAccount ? String(params.loginCustomerId) : ''` as suggested. Worth noting the residual delta: a byte-identical restoration isn't type-sound (the SDK config field is `string`; the original forwarded raw `string | number | undefined`), but the only divergent case — `subAccount: true` with a falsy `loginCustomerId` — is unreachable, since the transform throws `ConfigurationError` before ever producing such params.

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
```diff
@@ -0,0 +1,215 @@
...
+const processTrackEvent = (
+  metadata: Metadata,
+  message: GaecInputMessage,
+  destination: { Config: GaecConfig },
+): GaecDeliveryRequest => {
+  const { Config } = destination;
+  const { event } = message;
+  const { listOfConversions, adjustmentType } = Config;
+  const isConfiguredConversion =
+    Array.isArray(listOfConversions) && listOfConversions.some((i) => i.conversions === event);
```
shekhar-rudder, line 123: Per the JS→TS migration guide (`memory-bank/23_js_to_ts_migration.md`), this is a behavior change beyond a pure migration. Two things here go against the guide:

- The `Array.isArray(listOfConversions)` guard is a runtime check added to satisfy TS — the guide explicitly says *"Don't add runtime checks (like `Array.isArray`) just to satisfy TypeScript!"*
- It also changes runtime behavior: when `listOfConversions` is missing, the original crashed with a `TypeError` (500); this now throws `ConfigurationError` (400). Principle #1 ("Migration ≠ Refactoring") requires preserving the original error path.

If the intent is to keep the original behavior, drop the `Array.isArray` guard and keep `listOfConversions.some(...)` as-is.

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
```diff
@@ -0,0 +1,215 @@
...
+  if (!event || !isConfiguredConversion) {
```
shekhar-rudder: Operator change vs. the original `event === undefined || event === ''`. The guide's "Preserve Original JavaScript Behavior → Changing operators" section flags exactly this: `!event` additionally catches `0`, `false`, `NaN`. For a pure migration, keep the original condition (`event === undefined || event === ''`).

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
```diff
@@ -0,0 +1,215 @@
...
+    firstAdjustment.userIdentifiers = arr.filter((item) => !!item);
+  }
+
+  // `isFeatureEnabled` treats a flag value of 'ALL' as enabled regardless of workspaceId,
+  // so a missing/non-string workspaceId is passed as '' rather than short-circuiting to false.
+  const { workspaceId } = metadata;
+  if (
+    isFeatureEnabled(
+      'DEST_GAEC_ADJUSTMENT_TYPE_SUPPORTED_WORKSPACE_IDS',
+      typeof workspaceId === 'string' ? workspaceId : '',
```
shekhar-rudder: Behavior change vs. the original `metadata?.workspaceId`. This now passes `''` instead of `undefined` and introduces a conditional the original didn't have. It's practically inert (the `'ALL'`/unset-flag paths dominate), but per principle #1 a pure migration should pass the value through unchanged (`metadata.workspaceId`).

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
```diff
@@ -0,0 +1,215 @@
...
+      `Any of email, phone, firstName, lastName, city, street, countryCode, postalCode or streetAddress is required in traits.`,
+    );
+  }
+  const firstAdjustment: ConversionAdjustment = payload.conversionAdjustments![0];
+  firstAdjustment.adjustmentType = ADJUSTMENT_TYPE_ENHANCEMENT;
+  // Removing the null values from userIdentifier
+  // (truthiness guard rather than `!== undefined`: the vendored src/util/lodash-es-core.js
+  // declares a global `undefined`, which disables `!== undefined` narrowing project-wide)
+  const arr = firstAdjustment.userIdentifiers;
+  if (arr) {
```
shekhar-rudder: The `if (arr)` guard is an added runtime check — the original assigned `arr.filter(...)` unconditionally. Since the earlier `if (!...userIdentifiers) throw` already guarantees `userIdentifiers` is truthy here, this guard is only for TS narrowing, which the guide warns against ("Don't add runtime checks just to satisfy TypeScript"). Consider `!` instead to keep the assignment unconditional like the original.

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
```diff
@@ -0,0 +1,215 @@
...
+  constructPayload,
+  defaultRequestConfig,
+  removeHyphens,
+  simpleProcessRouterDest,
+  getAccessToken,
+} from '../../util';
+import { JSON_MIME_TYPE } from '../../util/constant';
+import { isFeatureEnabled } from '../../../util/featureFlags';
+
+const ADJUSTMENT_TYPE_ENHANCEMENT = 'ENHANCEMENT';
```
shekhar-rudder, line 27: Hoisting `ENHANCEMENT`/`RESTATEMENT` to constants (and replacing the `flag` sentinel with `isConfiguredConversion` above) are refactors. They're reasonable cleanups, but the migration guide's checklist #10 asks the `.ts` body to stay "byte-near-identical" to the original — pure-migration PRs should defer cleanups like this to a follow-up. Flagging for scope; fine to keep if the team accepts migration + light cleanup.

shekhar-rudder (commented):
Reviewed against `memory-bank/23_js_to_ts_migration.md`.

**Type-safety mechanics are exemplary** — no `as` casts (only the `NetworkHandler as networkHandler` import alias), no `any` outside catch, correct `!` usage on `constructPayload(...)`, no `Object.assign` narrowing, existing types imported rather than re-created, test file migrated and `.js` originals deleted.

**However, the guide mandates a *pure* migration** ("Migration ≠ Refactoring", checklist #10 "byte-near-identical"), and this PR bundles several refactors and behavior changes:

- **Behavior changes** (inline comments): the `Array.isArray` guard (TypeError→ConfigurationError), the `!event` operator change, the `workspaceId` `''`-vs-`undefined` change, and the added `if (arr)` guard.
- **Refactors beyond types**: hoisted `ENHANCEMENT`/`RESTATEMENT` constants + `flag`→`isConfiguredConversion`, the `isPartialFailureBody` helper, the class rename, and 4 new negative unit tests (adding coverage is outside migration scope).

Each deviation is deliberate and well-commented, so this is really a question of whether the team accepts "migration + light hardening" or wants a strict pure-migration PR. The behavior-changing items are the ones worth reconciling before merge.

src/v0/destinations/google_adwords_enhanced_conversions/networkHandler.ts
```diff
@@ -0,0 +1,262 @@
...
+const gaecResponseHandler = (responseParams: GaecResponseHandlerParams): GaecV0HandlerResult => {
+  const { destinationResponse } = responseParams;
+  const message = 'Request Processed Successfully';
+  const { status } = destinationResponse;
+  if (isHttpStatusSuccess(status)) {
+    // for google ads enhance conversions the partialFailureError returns with status 200
+    // a successful 200 may also have an empty/undefined body (no partial failures),
+    // so guard against reading partialFailureError off a non-object body
+    const responseBody = destinationResponse.response;
+    const partialFailureError = isPartialFailureBody(responseBody)
```
shekhar-rudder, line 207: The original read `const { partialFailureError } = destinationResponse?.response || {}`. Replacing it with the `isPartialFailureBody()` type-guard helper is a restructure beyond type annotations (checklist #10, "byte-near-identical"). Behavior matches, so this is a scope note rather than a correctness issue.

src/v0/destinations/google_adwords_enhanced_conversions/networkHandler.ts
```diff
@@ -0,0 +1,262 @@
...
+
+  constructor() {
+    this.proxy = gaecProxyRequest;
+    this.responseHandler = gaecResponseHandler;
+    this.processAxiosResponse = gaecProcessAxiosResponse;
+    this.prepareProxy = prepareProxyRequest;
+  }
+}
+
+export { NetworkHandler as networkHandler, gaecProxyRequest, gaecProcessAxiosResponse };
```
shekhar-rudder, line 262: Renaming the class `networkHandler` → `NetworkHandler` (and dropping the original `eslint-disable`) is a lint-motivated refactor, re-aliased on export to keep the factory lookup working. Against the guide's "byte-near-identical" rule for a pure migration, though harmless. Flagging for scope.

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
ItsSudip: Keeping this one deliberately — it's not migration drift but a separately-committed fix responding to CodeRabbit's earlier comment on this PR about exactly this TypeError: https://github.com/rudderlabs/rudder-transformer/pull/5349#discussion_r3550034282. It's intentionally isolated from the migration commits as `5b82eee77 fix(gaec): guard listOfConversions before conversion lookup`, per the same migration-vs-fix commit split we used for the rename restructure. On semantics: a missing `listOfConversions` is a dashboard-config problem, so `ConfigurationError` (400, permanent + actionable) matches our error-classification standard, whereas the original TypeError → 500 would retry a config error indefinitely. Happy to move it to a follow-up PR instead if you'd prefer the migration commits stay 100% pure.

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
ItsSudip: Keeping these, taking the "fine to keep" option — both came from an internal pre-push review pass applying our own review standards (hoist reused string sentinels, self-documenting names), and neither alters behavior. Point taken on scope: future migration PRs will carry cleanups like this in a separate commit or follow-up.

src/v0/destinations/google_adwords_enhanced_conversions/networkHandler.ts
ItsSudip: Keeping this one — the original `const { partialFailureError } = destinationResponse?.response || {}` can't compile in TS: `response` is `unknown`, and destructuring from `unknown` isn't allowed (`|| {}` doesn't narrow it), while `as`/`any` are off-limits under the same guide. The type guard is the minimal cast-free narrowing (guide #9's "runtime patterns TS cannot express" case), and the outcomes are identical for every input class — falsy, primitive, array, and object all yield exactly the `partialFailureError` the original destructure produced.

src/v0/destinations/google_adwords_enhanced_conversions/networkHandler.ts
ItsSudip: Keeping the rename — the export alias means zero consumer impact (the factory's `networkHandler.networkHandler` lookup and `v1/networkHandler.ts` imports are untouched, covered by the dataDelivery component tests), and reverting would reintroduce the lint violation the original suppressed with a disable comment. Noted for scope on future migrations, same as the constants thread.

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
ItsSudip: Fixed — restored the original condition: `event === undefined || event === '' || !isConfiguredConversion`.

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
ItsSudip: Fixed — direct pass-through restored (`metadata.workspaceId`); the conditional was a leftover from when `metadata` was typed `Record<string, unknown>` — with the shared `Metadata` type (required string `workspaceId`) it compiles directly.

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
ItsSudip: Fixed — guard removed; the assignment is unconditional again with a type-only `!` (`arr!.filter(...)`), since the throw above already guarantees `userIdentifiers` is present.

shekhar-rudder (approved):

PR #5351, opened 2026-07-13
fix: google enhanced conversion hashing

ItsSudip:

Fixes PII hashing in the Google Ads Enhanced Conversions (GAEC) destination to follow Google's documented normalization requirements, and validates that incoming data is consistent with the `requireHash` setting.

- **Normalization before hashing** (previously trim-only): email → trim + lowercase + gmail.com/googlemail.com username dot/`+suffix` stripping; phone → strip `space ( ) . -` separators and prepend `+` (E.164); first/last name and street address → trim + lowercase.
- **Hashing moved out of the mapping**: `hashToSha256` removed from `trackConfig.json`; the five PII fields now flow through a new `processUserIdentifiers` (GAEC `utils.ts`) which runs the shared `processAudienceRecord` pipeline (normalize → hashing-consistency check → validate → hash) — the same machinery GARL, fb_custom_audience, and tiktok_audience already use.
- **Hashing-consistency validation**: a pre-hashed value with hashing enabled, or a raw value with hashing disabled, now aborts with an `InstrumentationError` and emits the `audience_hashing_inconsistency` metric (previously: silent double-hashing / raw values sent to Google).
- **Data hygiene**: invalid identifiers (malformed email, non-numeric phone) are dropped with the `google_adwords_enhanced_conversions_invalid_field` metric; empty values are dropped instead of hashing `''`; identifier entries emptied by drops are pruned, and if nothing survives the existing "Any of email, phone, …" error fires. RESTATEMENT adjustments skip the pipeline (their identifiers are deleted before send).
- **Shared normalizers**: `normalizeEmail`/`normalizePhone` moved verbatim from GARL's `util.ts` into `src/v0/util/googleUtils/userDataNormalization.ts` (plus `normalizeName`); GARL imports them from there — zero behavior change (GARL unit + 81 component tests pass unmodified).

Resolves INT-6658

GAEC's trim-only hashing produced hashes Google can never match (e.g. `"John "` / `"USER@Gmail.com"`), and pre-hashed inputs were silently double-hashed — both surfaced as `hashed_email`/`hashed_first_name` failures. This PR brings GAEC to parity with GARL's Google-spec-compliant pipeline (verified against Google's Enhanced Conversions docs and the `OfflineUserAddressInfo` proto).

**Dependent change**: this PR is stacked on the GAEC JS→TS migration — base branch is `feature/int-6744-refactor-migrate-gaec-v0-destination-from-javascript-to` (#5349). Merge #5349 first.

Yes — intentional, per the team decision on INT-6658 (customers will be informed before release; no feature flag):

| Input (hashing enabled unless noted) | Before | After |
| --- | --- | --- |
| Mixed-case / unnormalized email, name, phone | hashed as-is → unmatchable at Google | normalized then hashed (matchable) |
| Pre-hashed (64-hex) value | silently double-hashed | `InstrumentationError` + metric |
| Raw value with `requireHash: false` | sent raw → rejected by Google | `InstrumentationError` + metric |
| Empty / whitespace-only value | `sha256('')` sent | dropped |
| Invalid email / non-numeric phone | hashed garbage | dropped + metric |
| No identifiers survive | n/a | existing "Any of email, phone, …" error |

N/A

`src/v0/util/googleUtils/userDataNormalization.ts` — Google Ads user-data normalizers (`normalizeEmail`, `normalizePhone`, `normalizeName`) shared by GARL and GAEC, relocated verbatim from GARL's `util.ts` into the existing Google-vendor-shared `googleUtils` module. Unit-tested in `userDataNormalization.test.ts`.

- Post-release monitoring: `audience_hashing_inconsistency{destType="google_adwords_enhanced_conversions"}` and `google_adwords_enhanced_conversions_invalid_field` show affected-event volume.
- Coverage: 35 GAEC component tests (5 new: normalization/gmail rules, pre-hashed → error, raw + `requireHash: false` → error, invalid-email drop, restatement skip), new unit suites for `utils.ts` and the normalizers; GARL suites pass unchanged as the code-motion regression check.

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist

### DIFF
--- a/src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
+++ b/src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
@@ -1,5 +1,4 @@
-import get from 'get-value';
-import { cloneDeep, isNumber } from 'lodash';
+import { isNumber } from 'lodash';
 import { InstrumentationError, ConfigurationError } from '@rudderstack/integrations-lib';
 import isString from 'lodash/isString';
 import type {
@@ -12,6 +11,7 @@ import type {
   ConversionAdjustment,
 } from './types';
 import { trackMapping } from './config';
+import { processUserIdentifiers } from './utils';
 import type { Metadata } from '../../../types';
 import type { RouterTransformationResponse } from '../../../types/destinationTransformation';
 import {
@@ -27,43 +27,13 @@ import { isFeatureEnabled } from '../../../util/featureFlags';
 const ADJUSTMENT_TYPE_ENHANCEMENT = 'ENHANCEMENT';
 const ADJUSTMENT_TYPE_RESTATEMENT = 'RESTATEMENT';
 
-/** Shape of a single element in the mapping JSON (trackConfig.json). */
-interface MappingElement {
-  metadata?: {
-    type?: string | string[];
-  };
-  [key: string]: unknown;
-}
-
-const isMappingElement = (value: unknown): value is MappingElement =>
-  typeof value === 'object' && value !== null;
-
-/**
- * This function is helping to update the mappingJson.
- * It is removing the metadata field with type "hashToSha256"
- * @param mapping -> it is the configMapping.json
- * @returns
- */
-const updateMappingJson = (mapping: unknown[]): unknown[] => {
-  const newMapping: unknown[] = [];
-  mapping.forEach((element) => {
-    if (
-      isMappingElement(element) &&
-      get(element, 'metadata.type') &&
-      element.metadata?.type?.includes('hashToSha256')
-    ) {
-      // eslint-disable-next-line no-param-reassign -- intentional in-place mutation: this function's purpose is to rewrite metadata.type
-      element.metadata.type = 'toString';
-    }
-    newMapping.push(element);
-  });
-  return newMapping;
-};
+const MISSING_IDENTIFIERS_ERROR =
+  'Any of email, phone, firstName, lastName, city, street, countryCode, postalCode or streetAddress is required in traits.';
 
 const responseBuilder = (
   metadata: Metadata,
   message: GaecInputMessage,
-  destination: { Config: GaecConfig },
+  destination: { Config: GaecConfig; [key: string]: unknown },
   payload: GaecPayload,
 ): GaecDeliveryRequest => {
   // typed at construction: the builder starts from empty slots and is mutated in place
@@ -114,11 +84,11 @@ const responseBuilder = (
 const processTrackEvent = (
   metadata: Metadata,
   message: GaecInputMessage,
-  destination: { Config: GaecConfig },
+  destination: { Config: GaecConfig; [key: string]: unknown },
 ): GaecDeliveryRequest => {
-  const { Config } = destination;
+  const { Config, ID } = destination;
   const { event } = message;
-  const { listOfConversions, adjustmentType } = Config;
+  const { listOfConversions, adjustmentType, requireHash } = Config;
   const isConfiguredConversion =
     Array.isArray(listOfConversions) && listOfConversions.some((i) => i.conversions === event);
   if (event === undefined || event === '' || !isConfiguredConversion) {
@@ -126,25 +96,18 @@ const processTrackEvent = (
       `Conversion named "${String(event)}" was not specified in the RudderStack destination configuration`,
     );
   }
-  const { requireHash } = Config;
-  let updatedMapping = cloneDeep(trackMapping);
 
-  if (requireHash === false) {
-    updatedMapping = updateMappingJson(updatedMapping);
-  }
-
-  // `!` mirrors the original JS's implicit non-null trust in `constructPayload`'s return
-  // (migration guide #6) — a null surfaces as a TypeError on the next access, exactly
-  // like the original JS.
-  const payload: GaecPayload = constructPayload(message, updatedMapping)!;
+  // `!` mirrors the original JS's implicit non-null trust in `constructPayload`'s return —
+  // a null surfaces as a TypeError on the next access, exactly like the original JS.
+  // hashToSha256 has been removed from trackConfig.json; the mapping now extracts raw values
+  // only (trim). processUserIdentifiers handles normalization, consistency-check, and hashing.
+  const payload: GaecPayload = constructPayload(message, trackMapping)!;
 
   payload.partialFailure = true;
   // `?.` on [0] is a genuine runtime guard: the array may be empty, and the intended
   // failure is this InstrumentationError, not a TypeError
   if (!payload.conversionAdjustments![0]?.userIdentifiers) {
-    throw new InstrumentationError(
-      `Any of email, phone, firstName, lastName, city, street, countryCode, postalCode or streetAddress is required in traits.`,
-    );
+    throw new InstrumentationError(MISSING_IDENTIFIERS_ERROR);
   }
   const firstAdjustment: ConversionAdjustment = payload.conversionAdjustments![0];
   firstAdjustment.adjustmentType = ADJUSTMENT_TYPE_ENHANCEMENT;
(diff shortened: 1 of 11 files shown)


src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
```diff
@@ -154,15 +117,36 @@ const processTrackEvent = (
...
+    processUserIdentifiers(payload, {
+      requireHash,
+      workspaceId: metadata.workspaceId,
+      destinationId,
+    });
+
+    // After pruning, re-check: processUserIdentifiers may have dropped all identifiers
+    // (e.g. all were invalid), leaving userIdentifiers empty.
+    if (!firstAdjustment.userIdentifiers || firstAdjustment.userIdentifiers.length === 0) {
+      throw new InstrumentationError(MISSING_IDENTIFIERS_ERROR);
```
shekhar-rudder, line 146: This re-check is confusing and worth refactoring. `processUserIdentifiers` returns `void` and prunes by mutating `firstAdjustment.userIdentifiers` in place; this check works only because the local `firstAdjustment` here is the *same object reference* that the util mutates. The correctness of "did everything get dropped?" is implicitly coupled to that mutation staying a property reassignment — if the util ever reassigned `conversionAdjustments[0]` instead, this alias would silently read the stale object.

Consider having `processUserIdentifiers` return the surviving identifiers (or a boolean / count) and letting the caller decide, so the "all dropped → error" logic doesn't depend on shared-reference mutation semantics.

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
```diff
@@ -27,43 +27,13 @@ import { isFeatureEnabled } from '../../../util/featureFlags';
...
-};
+const MISSING_IDENTIFIERS_ERROR =
+  'Any of email, phone, firstName, lastName, city, street, countryCode, postalCode or streetAddress is required in traits.';
 
 const responseBuilder = (
   metadata: Metadata,
   message: GaecInputMessage,
-  destination: { Config: GaecConfig },
+  destination: { Config: GaecConfig; [key: string]: unknown },
   payload: GaecPayload,
```
shekhar-rudder, line 37: Rather than loosening the inline param type with an index signature to read `ID`, use the destination type that already models this. `Destination<GaecConfig>` (aliased here as `GaecDestination` in `types.ts`) already declares `ID: string` alongside `Config`, so typing the param as `GaecDestination` gives you `ID` type-safely and drops the `[key: string]: unknown` bridge (in both `responseBuilder` and `processTrackEvent`).

Note: `ID` is a sibling of `Config` on the destination, not a field of `GaecConfig` itself, so the fix is the destination type, not `GaecConfig`.

shekhar-rudder (commented):
Two refactor suggestions on the new hashing wiring — both non-blocking. The change itself is correct and well-tested; these are about the coupling/typing around `processUserIdentifiers` and the destination param.

shekhar-rudder (commented):
Suggestion on making `processUserIdentifiers` more declarative.

src/v0/destinations/google_adwords_enhanced_conversions/utils.ts
```diff
@@ -0,0 +1,215 @@
...
+    }
+    if (typeof value === 'string' || typeof value === 'number') {
+      rawFields[key] = value;
+    } else {
+      // eslint-disable-next-line no-param-reassign -- intentional in-place delete of an unusable non-scalar field
+      delete entry[key];
+    }
+  };
+  collectHashableField(emailEntry, 'hashedEmail');
+  collectHashableField(phoneEntry, 'hashedPhoneNumber');
```
shekhar-rudder: `processUserIdentifiers` would be clearer if it were table-driven and returned its result instead of mutating the payload in place.

The field→container binding is currently restated three times — these `collectHashableField` calls, the `writeBack` calls below, and the `addressInfo` prune. Expressing it once as a data table and looping generically makes adding or removing a PII field a one-line change:

```ts
const targets = [
  { key: 'hashedEmail',         obj: findEntry('hashedEmail') },
  { key: 'hashedPhoneNumber',   obj: findEntry('hashedPhoneNumber') },
  { key: 'hashedFirstName',     obj: addressInfo },
  { key: 'hashedLastName',      obj: addressInfo },
  { key: 'hashedStreetAddress', obj: addressInfo },
].filter((t) => t.obj);
```

Alongside that, have the function build and return the surviving identifiers rather than mutating in place, so the "did everything get dropped?" check in `transform.ts` reads the return value instead of relying on the shared `firstAdjustment` reference. The non-hashable `city`/`state`/`countryCode`/`postalCode` address fields need to be carried through untouched. The nested `addressInfo` resolution and the drop-non-scalar-before-`processAudienceRecord` ordering can stay as they are.

src/v0/destinations/google_adwords_enhanced_conversions/utils.test.ts
```diff
@@ -0,0 +1,291 @@
...
+      expect(identifiers).toEqual([{ addressInfo: { city: 'London' } }]);
+    });
+
+    it('prunes the entire address entry when all hashable AND non-hashable address fields are absent', () => {
+      // Only firstName provided, it's invalid
+      const payload = makePayload({ hashedFirstName: '' });
+      processUserIdentifiers(payload, makeDestCtx(true));
+      const identifiers = payload.conversionAdjustments?.[0]?.userIdentifiers;
+      expect(identifiers).toEqual([]);
+    });
```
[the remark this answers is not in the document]
ItsSudip: Fixed — the two pruning tests now pass whitespace-only values (`'   '`), which survive `makePayload`'s truthy insertion, normalize to `''` inside the pipeline, and are dropped there. The tests now exercise the normalization-drop path with unchanged expectations.

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
ItsSudip: Addressed — `processUserIdentifiers` now returns the surviving identifier entries instead of mutating the payload. The caller throws on an empty result and assigns the survivors explicitly, so the "all dropped → error" logic no longer depends on shared-reference mutation semantics.

src/v0/destinations/google_adwords_enhanced_conversions/transform.ts
ItsSudip: Done — `responseBuilder` / `processTrackEvent` / `processEvent` now take `GaecDestination`, and `GaecProcessInput.destination` is typed the same, so the index-signature bridge is gone. One deliberate keep: the runtime `typeof ID === 'string'` guard stays (with a comment) — the envelope isn't runtime-validated (the Zod schema passes `destination` through unchecked), so the `''` fallback covers the type-vs-runtime gap for the metric label.

src/v0/destinations/google_adwords_enhanced_conversions/utils.ts
ItsSudip: Refactored as suggested — the field→container binding is declared once in a `hashableTargets` table, collection loops over it, and the surviving entries are rebuilt fresh in mapping order and returned. Non-hashable address fields carry through on a local copy; non-scalars simply aren't collected. Side benefit: the `writeBack` helper and all three `no-param-reassign` eslint-disables are gone, since nothing mutates the payload anymore.

src/v0/destinations/google_adwords_enhanced_conversions/utils.ts
```diff
@@ -0,0 +1,209 @@
...
+    if (hashedLastName) {
+      rebuiltAddressInfo.hashedLastName = hashedLastName;
+    }
+    const hashedStreetAddress = survivingValue('hashedStreetAddress');
+    if (hashedStreetAddress) {
+      rebuiltAddressInfo.hashedStreetAddress = hashedStreetAddress;
+    }
+    if (Object.keys(rebuiltAddressInfo).length > 0) {
+      survivors.push({ addressInfo: rebuiltAddressInfo });
+    }
```
shekhar-rudder: **Optional simplification — the address rebuild still restates each field 3×.**

The earlier table-driven refactor unified the *collection* half (`hashableTargets`), but the *rebuild* half is still `delete` + conditional-reassign per field. It's really "for each address key: replace with surviving value, else drop", which is a loop over a single key list — the same list can also drive collection so the field set is declared exactly once.

Declare the hashed address sub-fields once (module scope, next to `GAEC_FIELD_CONFIG`):

```ts
// The hashed address sub-fields, listed once — used both to collect raw values and to
// rebuild the surviving addressInfo, so adding/removing one is a single edit.
const ADDRESS_HASH_KEYS = ['hashedFirstName', 'hashedLastName', 'hashedStreetAddress'] as const;
```

Drive collection from it:

```ts
const hashableTargets = [
  { key: 'hashedEmail', container: emailEntry },
  { key: 'hashedPhoneNumber', container: phoneEntry },
  ...ADDRESS_HASH_KEYS.map((key) => ({ key, container: addressInfo })),
];
```

Then this whole block collapses to a loop:

```ts
if (addressInfo) {
  // Non-hashable fields (city, state, countryCode, postalCode) carry through untouched;
  // each hashed sub-field is replaced by its surviving value or dropped.
  const rebuiltAddressInfo: AddressInfo = { ...addressInfo };
  ADDRESS_HASH_KEYS.forEach((key) => {
    const value = survivingValue(key);
    if (value) {
      rebuiltAddressInfo[key] = value;
    } else {
      delete rebuiltAddressInfo[key];
    }
  });
  if (Object.keys(rebuiltAddressInfo).length > 0) {
    survivors.push({ addressInfo: rebuiltAddressInfo });
  }
} else if (addressEntry) {
  survivors.push(addressEntry);
}
```

Behaviorally identical (copy → drop raw → re-add processed), ~15 lines lighter, and adding/removing a hashed address field becomes a one-line list edit on both paths. Verified locally: lint clean, `tsc` clean, all 68 unit tests + the component suite pass.

Leaving email/phone explicit is fine — the `{ hashedEmail }` shorthand reads better than computed keys and those two are fixed by Google's API. Non-blocking.

shekhar-rudder (approved):

src/v0/destinations/google_adwords_enhanced_conversions/utils.ts
ItsSudip: Addressed

PR #5361, opened 2026-07-13
refactor(nativeBatching): discriminate the hybrid input union on message.type

maheshkutty:

Hybrid (record + event-stream) destinations validated their input with a plain z.union. Because `destination.Config` is declared in both variants, a single bad Config failed both branches, and Zod reported it as one opaque `invalid_union` issue with no usable path. `resolveIssues` existed to work backwards from that: it searched for a branch that failed only on destination/connection and assumed that was the branch the caller meant — a heuristic standing in for information the schema had already discarded.

Build the union with z.discriminatedUnion instead. Zod 3 only reads a discriminator at the top level of an option, so makeHybridInputSchema lifts the nested `message.type` to a synthetic key in a preprocess step, discriminates on it, and types the result as the clean variant union so the key never reaches destinations (validateInputs forwards the original input, not Zod's output).

Zod now selects exactly one branch and reports only its issues, with real paths (`destination.Config.apiKey`, `connection.config.destination.object`), so resolveIssues is deleted rather than simplified. Error classification is unchanged: isConfigIssue and the every(...) rule are untouched.

The customerio schemas are unchanged — their `.refine()`s sit on `message`, nested inside the envelope, so each variant is still a ZodObject.

[Resolves INT-6792](https://linear.app/rudderstack/issue/INT-6792/use-discriminator-union-for-input-validation)

N/A

N/A

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist


src/services/destination/nativeBatching/inputSchema.ts
```diff
@@ -62,3 +62,50 @@ export function makeRouterInputSchema(spec: {
...
+ *
+ * Validates the variants as a *discriminated* union rather than a plain `z.union`.
+ * A plain union collapses every branch's failure into one opaque `invalid_union` issue
+ * with no usable path — so a bad `destination.Config`, which both branches declare,
+ * fails both, and the caller could only guess which branch was meant. Discriminating on
+ * `message.type` makes Zod pick exactly one branch and report only its issues, with real
+ * paths (`destination.Config.apiKey`, `connection.config.destination.object`, …).
+ */
+export function makeHybridInputSchema<TRecord extends ZodTypeAny, TEventStream extends ZodTypeAny>(
+  recordSchema: TRecord,
```
shekhar-rudder: `makeHybridInputSchema` is only ever consumed by `VDMV2ObjectDestination.getInputSchema()` — unlike `makeRouterInputSchema`, which destination authors call directly. Since it's specific to the hybrid base class, consider moving it into `vdmV2ObjectDestination.ts` (or inlining it into `getInputSchema`) rather than placing it in the shared `inputSchema.ts` and re-exporting it publicly through `batchDestination.ts`. That keeps the framework's public surface limited to builders destinations actually use, and colocates the discriminator machinery with the only thing that depends on it.

shekhar-rudder (commented):
Reviewed the discriminated-union refactor — the logic is correct and well-tested (traced through the preprocess/discriminator flow, config-error classification, and the discriminator-leak edge case; nativeBatching units pass). One design note on where the new builder lives.

src/services/destination/nativeBatching/inputSchema.ts
maheshkutty: Good call — done.

`makeHybridInputSchema` is gone from `inputSchema.ts` and its re-export is dropped from `batchDestination.ts`, so the shared module is back to exporting only `makeRouterInputSchema` (the builder destinations actually call). The discriminator machinery now lives in `vdmV2ObjectDestination.ts` as a module-private `buildInputSchema`, colocated with its only consumer. Renamed off the `make*` prefix since it is no longer a public builder.

One knock-on: the union tests in `processBatchedDestination.test.ts` had imported the helper from the barrel. Rather than keep it exported just for them, `UnionIntegration` now extends `VDMV2ObjectDestination` and builds its variants with `makeRouterInputSchema` — matching the pattern in `vdmV2ObjectDestination.test.ts`. That is better coverage anyway: the tests exercise the schema the framework really builds instead of a hand-rolled copy that could drift from it.

Verified: typecheck / eslint / prettier clean; framework 46/46, customerio 56 unit + 123 component passing.

src/services/destination/nativeBatching/__tests__/processBatchedDestination.test.ts
```diff
@@ -519,41 +521,45 @@ describe('validateInputs', () => {
...
+    destinationConfig: unionDestConfig,
+    connectionConfig: z.object({ destination: z.object({ object: z.string() }) }).passthrough(),
+  });
+  const eventVariant = makeRouterInputSchema({
+    message: z.object({ type: z.enum(['track', 'identify']), userId: z.string() }).passthrough(),
+    destinationConfig: unionDestConfig,
+  });
+
+  class UnionIntegration extends VDMV2ObjectDestination<
+    TestBody,
```
shekhar-rudder: These hybrid discriminated-union tests exercise the schema that `VDMV2ObjectDestination` builds, so they belong in a `vdmV2ObjectDestination` test file rather than here in `processBatchedDestination`'s tests.

src/services/destination/nativeBatching/__tests__/processBatchedDestination.test.ts
```diff
@@ -619,5 +625,32 @@ describe('validateInputs', () => {
     );
     expect(errors).toHaveLength(1);
     expect(errors[0].statTags?.errorType).toBe('instrumentation');
+    // Non-record types route to the event-stream variant, whose `type` enum names the
+    // event types it accepts — rather than merging in the record variant's complaints.
+    expect(errors[0].error).toContain('message.type');
+    expect(errors[0].error).not.toContain('identifiers');
```
shekhar-rudder: Assert on the full error message (`toBe`/`toEqual`) rather than `toContain` here and in the other assertions — that way the test pins the exact error string surfaced to the user.

shekhar-rudder (commented):
Overall this is a clean, well-scoped refactor and the discriminated-union approach is sound — verified the unit and CustomerIO component suites pass. Two comments on the test changes.

src/services/destination/nativeBatching/__tests__/processBatchedDestination.test.ts
maheshkutty: Agreed — moved. The hybrid discriminated-union cases now live in `vdmV2ObjectDestination.test.ts` under a `VDMV2ObjectDestination input validation` describe, reusing the `TestObjectDestination` fixture already defined there rather than restating the variants. `processBatchedDestination.test.ts` keeps only its own generic `validateInputs` case, and the imports it had picked up for the union tests (`VDMV2ObjectDestination`, `makeRouterInputSchema`) are gone with them.

src/services/destination/nativeBatching/__tests__/processBatchedDestination.test.ts
maheshkutty: Done — every assertion now pins the full message with `toBe` instead of `toContain`:

- `destination.Config.apiKey: Required`
- `connection.config.destination.object: Required`
- `message.identifiers: Required` (record variant)
- `message.userId: Required` (event-stream variant)
- `message.type: Invalid enum value. Expected 'track' | 'identify', received 'group'; message.userId: Required`

Worth noting the last one: the unknown-type case actually surfaces *two* joined issues, which `toContain` was hiding — so this caught something real. I captured each string from the actual validator output rather than transcribing them by hand.

shekhar-rudder (approved):

PR #5362, opened 2026-07-13
chore: log unknown destinations resolved via handler-name lookup

shekhar-rudder:
## What

Emit a `warn` log when `getDestinationHandlerName` resolves a destination that isn't in the registry (and `REJECT_UNKNOWN_DESTINATIONS` is not enabled, so the request proceeds).

## Why

The destination-validation middleware already logs `Unknown destination encountered: <name>` on this same fallback path, but the handler-name lookup resolved unknown destinations silently. That left a gap: requests reaching the transformer through this path produced no signal. Mirroring the log across both entry points lets a single log-based alert catch every unknown destination regardless of which path handled it.

## How

- Added the matching `logger.warn(\`Unknown destination encountered: ...\`)` on the fallback branch of `getDestinationHandlerName`.
- Added a unit test asserting the warning fires (and the handler name still resolves) for an unknown destination.

Behavior is unchanged when `REJECT_UNKNOWN_DESTINATIONS=true` — that path still throws a `ConfigurationError`.

### DIFF
--- a/src/features.test.ts
+++ b/src/features.test.ts
@@ -7,6 +7,7 @@ import defaultFeaturesConfig, {
 } from './features';
 import { DestHandlerMap } from './constants/destinationCanonicalNames';
 import { getIntegrations } from './routes/utils';
+import logger from './logger';
 
 const getDestinationDirectories = () =>
   ['v0/destinations', 'v1/destinations', 'cdk/v2/destinations'].flatMap((destinationRoot) =>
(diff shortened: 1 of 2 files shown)


shekhar-rudder:
Closing: the added log is redundant. `DestinationValidationMiddleware` already emits `Unknown destination encountered: <name>` on every entry point that later reaches `getDestinationHandlerName` (processor, router, batch, deleteUsers), and no route reaches the handler lookup without passing through that middleware first. So this wasn't closing a silent gap — it logged the same unknown destination twice per request, inflating any log-based alert ~2x. The middleware already provides the single alert signal we wanted.

PR #5370, opened 2026-07-15
refactor(braze): group-preserving chunking + size caps + per-output metadata

ItsSudip:

Introduces a Braze batching refactor plus a rollout gate for the new per-job delivery-mapping output shape needed by INT-6634 (296 "Delivered with Warning" support).

**Always-on improvements** (regardless of flag):
1. **Cross-chunk straddle prevention.** Track chunking now walks source-job groups instead of individual items, so a single job's `attributes` + `events` (+ purchases) contributions always land in the same chunk. Prerequisite for correct metadata↔chunk ownership.
2. **Byte-size caps at chunking time.** Enforces Braze's documented limits — 100 KB per item, 4 MB per batch. Oversized single items and oversized single jobs are rejected up-front with `InstrumentationError` instead of building a batch Braze would reject wholesale.
3. **Type-safety cleanups.** `TaggedItem` is a discriminated union on `type`; type predicates (`isTrackBody` / `isSubscriptionBody` / `isMergeBody`) + a discriminated `JobClassification` push all narrowing into the type system — no `as` casts introduced. `BrazeSubscriptionBatchPayload.subscription_groups` typed as `BrazeSubscriptionGroup[]` (was `unknown[]`). `SubscriptionRun.entries` / `MergeRun.entries` renamed to `items` for consistency across `Run` variants.

**Gated changes** (`BRAZE_PER_JOB_DELIVERY_MAPPING_ENABLED=true`, default **OFF**):
- **One `BatchRequestOutput` per outgoing HTTP request.** Each `/users/track` / subscription-groups / alias-merge chunk becomes its own `BatchRequestOutput` with its own scoped metadata slice (instead of a single `MultiBatchRequestOutput` carrying all endpoints' chunks + a flat metadata list).
- **Insertion-order runs preserved.** Outputs are emitted in the input's insertion-order runs so per-user jobIds stay monotonically ascending across the emitted list.
- **`destInfo` positional map on every metadata entry.** Track metadata carries `destInfo.attributesIndices` / `eventsIndices` / `purchasesIndices` arrays; sub/merge metadata carries `destInfo: {}` (present-but-empty) so the v1 networkHandler sees a uniform shape across every chunk. Populated per the design doc's Option 1 contract — top-level index arrays, no per-destination wrapper.
- **Per-chunk metadata scoping.** A 5xx on any single chunk of a multi-chunk delivery no longer wrongly marks jobs from *unrelated* chunks as failed — latent behavior only fixed in the ON path.

When OFF (default), the router output shape is identical to pre-INT-6808 (single `MultiBatchRequestOutput` with `batchedRequest[]` and flat `metadata`) — verified by the unchanged integration snapshots in `router/data.ts` + `router/identityResolution.ts`.

Resolves [INT-6808](https://linear.app/rudderstack/issue/INT-6808)

Prep for INT-6634 (296 warning correlation). The network handler needs a way to map Braze's `errors[input_array, index]` back to originating `jobId`s; that requires (a) each job's contributions to be reachable within a single proxy response and (b) each metadata to carry positional info tying it back to the outgoing payload. Both requirements demand the batching-layer changes above. Gating the new output shape behind `BRAZE_PER_JOB_DELIVERY_MAPPING_ENABLED` lets INT-6634 land + soak in production without a hard cutover.

- **Router output shape unchanged by default** (flag OFF): single `MultiBatchRequestOutput` per invocation, flat `metadata`, no `destInfo` — identical to develop.
- **Byte-size caps added (both paths)**: oversized single items (> 100 KB) and oversized single jobs (> 75 items or > 4 MB) are now rejected up-front. Previously these silently created batches Braze would reject. Motivation: fail fast at the transformer.
- **`entries` → `items` rename on internal `Run` types**: not user-visible, internal naming consistency only.

N/A

N/A — all changes are Braze-scoped.

- Group-preserving chunking may close a `/users/track` chunk one item early to keep a job's contributions together. Chunk-count impact in practice: negligible (a typical job contributes ≤ 2 items).
- When the flag is ON, insertion-order runs may produce more discrete outputs when a batch interleaves multiple endpoint types (track ↔ subscription ↔ merge). Real batches are usually dominated by one type, so extra outputs are rare in practice.
- Byte-size measurement uses `Buffer.byteLength(JSON.stringify(item))` per item at chunk time — O(N) hot-path cost but bounded by the same items already being serialized shortly after.
- `isPerJobDeliveryMappingEnabled()` is read dynamically per invocation (not module-load), so a rollout config flip takes effect without a service restart.

@coderabbitai review

<hr>

### Developer checklist

- [x] **No breaking changes are being introduced.** — default (flag OFF) preserves pre-INT-6808 router output shape verified via unchanged integration snapshots.

- [x] All related docs linked with the PR? — Linear ticket + design doc referenced.

- [x] All changes manually tested? — braze unit + component tests all pass locally.

- [ ] Any documentation changes needed with this change? — none required for this PR; INT-6634 will add the 296 contract docs.

- [x] Is the PR limited to 10 file changes? — 5 files.

- [x] Is the PR limited to one linear task? — INT-6808 only.

- [x] Are relevant unit and component test-cases added in **new readability format**? — unit tests cover both OFF (default assertions) and ON (env-swap describe block); integration tests added for both paths.

### Reviewer checklist

### DIFF
--- a/src/v0/destinations/braze/util.ts
+++ b/src/v0/destinations/braze/util.ts
@@ -947,7 +947,7 @@ const processBatch = (transformedEvents: BrazeTransformedEvent[]) => {
  */
 // ===========================================================================
 // ON-path helpers & processBatchWithDeliveryMapping
-// Selected by transform.ts when BRAZE_PER_JOB_DELIVERY_MAPPING_ENABLED='true'.
+// Selected by transform.ts when BRAZE_PER_JOB_DELIVERY_MAPPING_WORKSPACE_IDS enables the workspace.
 // Everything above this section is bit-identical to develop.
 // ===========================================================================
 
@@ -1365,8 +1365,8 @@ const classifyJobRun = (json: unknown): JobClassification => {
 // byte-size caps per item (100 KB) and per batch (4 MB), and up-front
 // oversized-job rejection.
 //
-// Selected by `transform.ts` when `BRAZE_PER_JOB_DELIVERY_MAPPING_ENABLED`
-// is set to `'true'`.
+// Selected by `transform.ts` when `BRAZE_PER_JOB_DELIVERY_MAPPING_WORKSPACE_IDS`
+// enables the invocation's workspace.
 // ---------------------------------------------------------------------------
 const processBatchWithDeliveryMapping = (
   transformedEvents: BrazeTransformedEvent[],
--- a/src/v0/destinations/braze/util.ts
+++ b/src/v0/destinations/braze/util.ts
@@ -34,16 +34,16 @@ import {
 } from './config';
 import { JSON_MIME_TYPE, HTTP_STATUS_CODES } from '../../util/constant';
 import {
-  BrazeTrackRequestBody,
-  BrazeSubscriptionBatchPayload,
-  BrazeMergeBatchPayload,
-  BrazeBatchRequest,
   BrazeDestination,
   BrazeRouterRequest,
   BrazeBatchHeaders,
   BrazeTransformedEvent,
   BrazeBatchResponse,
+  BrazeBatchRequest,
   BrazeDestInfo,
+  BrazeTrackRequestBody,
+  BrazeSubscriptionBatchPayload,
+  BrazeMergeBatchPayload,
   BrazeSubscriptionGroup,
   BrazeAliasToIdentify,
   BrazeUserExportResponse,
@@ -56,6 +56,13 @@ import {
   BrazeMergeUpdate,
 } from './types';
 import type { Metadata } from '../../../types';
+
+type TrackChunk = {
+  attributes: BrazeUserAttributes[];
+  events: BrazeEvent[];
+  purchases: BrazePurchase[];
+  externalIds: Set<string>;
+};
 
 const formatGender = (gender: unknown) => {
   if (typeof gender !== 'string') {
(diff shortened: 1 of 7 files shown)


src/v0/destinations/braze/util.ts
```diff
@@ -537,247 +539,251 @@ const processDeduplication = (
...
-        subscription_groups: deduplicatedSubscriptionGroups,
-      });
-      responseArray.push({
-        ...response,
-        headers,
-      });
-    }
-  }
-}
+const computeItemByteSize = (item: unknown): number => Buffer.byteLength(JSON.stringify(item));
```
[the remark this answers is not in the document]
ItsSudip: Thanks — the observation is correct: `computeItemByteSize` sums the per-item `JSON.stringify` sizes and doesn't count the `{"…":[…]}` envelope, the inter-item commas, or the injected `partner` field. I checked whether that undercount can actually push a request past what Braze accepts, and it can't. Keeping the current accounting.

**Empirical check against live Braze.** I built a boundary payload where the transformer's accounting and the real body straddle 4 MB:

| Measurement | Bytes | vs 4 MB |
|---|---:|---:|
| `computeItemByteSize` sum (what the chunker checks) | 4,194,288 | 16 B **under** |
| Actual serialized body (what Braze receives) | 4,194,342 | 38 B **over** |
| Envelope overhead (wrapper + 41 commas) | 54 | — |

So the chunker treats this chunk as "within 4 MB" while the wire body is 38 B over. I sent it to the live `/users/track` endpoint and Braze accepted it — it doesn't hard-reject at the documented 4 MB boundary. The documented cap has margin, and the undercount lives entirely inside it.

**The undercount is a small, bounded, fixed term.** It's the envelope plus one comma per item — nothing that scales with payload. Item counts are hard-capped at `TRACK_BRAZE_MAX_REQ_COUNT = 75` per type, so even with all three arrays full the un-accounted overhead is only a few hundred bytes (54 B in the single-array test above, before the ~25 B `partner` field) — under ~0.01% of the 4 MB budget, and it can't compound.

**Per-item cap backstops it.** Every contribution is independently rejected above `TRACK_BRAZE_MAX_ITEM_BYTE_SIZE = 100 KB`, so no single item can blow the envelope either.

Making the accounting byte-exact would mean serializing the full envelope (keys + separators + partner) for every candidate group on the chunking hot path — real per-decision CPU cost to reclaim tens of bytes against a cap Braze already tolerates going over. Not a worthwhile trade, so leaving as-is.

src/v0/destinations/braze/util.ts
```diff
@@ -938,6 +945,530 @@ const processBatch = (transformedEvents: BrazeTransformedEvent[]) => {
...
+// oversized-job rejection.
+//
+// Selected by `transform.ts` when `BRAZE_PER_JOB_DELIVERY_MAPPING_WORKSPACE_IDS`
+// enables the invocation's workspace.
+// ---------------------------------------------------------------------------
+const processBatchWithDeliveryMapping = (
+  transformedEvents: BrazeTransformedEvent[],
+): BrazeBatchResponse[] => {
+  const { destination } = transformedEvents[0];
+  const workspaceId = transformedEvents[0].metadata?.[0]?.workspaceId || '';
```
achettyiitr:
```suggestion
  const { destination, metadata } = transformedEvents[0];
  const workspaceId = metadata?.[0]?.workspaceId || '';
```

src/v0/destinations/braze/util.ts
```diff
@@ -938,6 +945,530 @@ const processBatch = (transformedEvents: BrazeTransformedEvent[]) => {
...
+// three Braze payload shapes, so this classifier is total. If the contract
+// is ever violated we throw rather than silently drop the job.
+const classifyJobRun = (json: unknown): JobClassification => {
+  if (isTrackBody(json)) return { type: 'track', body: json };
+  if (isSubscriptionBody(json)) return { type: 'subscription', body: json };
+  if (isMergeBody(json)) return { type: 'merge', body: json };
+  throw new InstrumentationError(
+    'Braze processBatchWithDeliveryMapping: body is neither track, subscription, nor merge',
+  );
+};
```
achettyiitr, line 1352: Check if we can use something like this?

```
switch (transformedEvent.batchedRequest?.endpointPath) {
  case 'users/track':
    // body is BrazeTrackRequestBody
  case 'v2/subscription/status/set':
    // body is BrazeSubscriptionBatchPayload
  case 'users/merge':
    // body is BrazeMergeBatchPayload
  default:
    throw new InstrumentationError(...);
}
```

ItsSudip: Good call on `endpointPath` — it's the authoritative discriminator (the transform sets it via `getTrackEndPoint`/`getSubscriptionGroupEndPoint`/`getAliasMergeEndPoint`), so dispatching on it is more direct than sniffing the body shape.

One constraint before I switch it, though: TypeScript can't derive the body's type from the `endpointPath` string, so each `case` would need an `as` cast (`body as BrazeTrackRequestBody`, etc.). That's exactly what the current `classifyJobRun` avoids — `isTrackBody`/`isSubscriptionBody`/`isMergeBody` are `json is X` type guards, so the body is narrowed **cast-free**. Moving to the `switch` reintroduces `as` casts, which we've been deliberately keeping out of this PR.

It also means updating the ON-path unit fixtures, which currently don't set `endpointPath` (they rely on the body shape).

Three ways I can take it — happy to go whichever you prefer:
1. **Keep `classifyJobRun`** as-is (cast-free, classifies by body shape).
2. **`switch` on `endpointPath` with `as` casts** per case (closest to your snippet; removes the predicates + `classifyJobRun`).
3. **Hybrid** — `switch` on `endpointPath` (authoritative) but narrow via the existing guards inside each case and throw on a path/body mismatch. Stays cast-free and adds a consistency check, at the cost of a few extra lines.

I'd lean 3; 2 removes the most code. Which would you prefer?

achettyiitr (approved):

PR #5375, opened 2026-07-20
fix(customerio): stop rejecting non-object traits and context in v2 schema

maheshkutty:

The v2 event-stream schema validated `traits` and `context.traits` against an object shape, so events carrying either as a non-object — some sources send `context` as an array — were rejected with "Expected object, received array" before reaching the transform.

Drop the shape validation and pass both through as `z.unknown()`. The refine logic is unchanged; it now reads via `get-value`, which returns undefined on a non-object root instead of throwing, so email and mappedToDestination detection still work for the object case.

Resolves INT-XXX

N/A

N/A

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist

### DIFF
--- a/src/v0/destinations/customerio/v2/types.ts
+++ b/src/v0/destinations/customerio/v2/types.ts
@@ -1,4 +1,5 @@
 import { z } from 'zod';
+import get from 'get-value';
 import { makeRouterInputSchema } from '../../../../services/destination/nativeBatching/batchDestination';
 import { RECORD_IDENTIFIER_KEYS } from './config';
 import { CustomerIODestinationConfigSchema, CustomerIOConnectionConfigSchema } from '../types';
@@ -34,8 +35,6 @@ export type CustomerIOV2Payload = {
   [key: string]: unknown;
 };
 
-const emailTraitSchema = z.object({ email: z.unknown() }).passthrough().nullish();
-
 const recordMessageSchema = z
   .object({
     type: z.literal('record'),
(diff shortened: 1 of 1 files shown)


src/v0/destinations/customerio/v2/types.ts
```diff
@@ -63,18 +62,14 @@ const eventStreamMessageSchema = z
...
-      })
-      .passthrough()
-      .nullish(),
+    // traits and context are passed through untyped (not validated here): some
+    // sources send them as a non-object (e.g. an array). RETL/warehouse sources
+    // set context.mappedToDestination and supply the identifier via externalId;
+    // adduserIdFromExternalId (called in processV2, after this validation) hydrates
+    // userId — so those events must pass the refine even without a top-level
+    // userId/anonymousId/email.
+    traits: z.unknown(),
```
shekhar-rudder, line 71: **Possible regression: non-object `traits` (e.g. an array) now produces mangled attributes instead of being rejected.**

Relaxing `traits` to `z.unknown()` lets non-object `traits` reach the downstream identify path, which the old `emailTraitSchema` (`z.object({...}).nullish()`) previously rejected. In `v2/util.ts` (`buildTraitAttributes`, ~L52-54):

```js
const pathToTraits = message.traits ? 'traits' : 'context.traits';
const traits = getFieldValueFromMessage(message, 'traits') || {};
Object.keys(traits).filter(...).forEach((trait) => set(attributes, ..., get(message, `${pathToTraits}.${trait}`)));
```

On an identify event that still passes the refine (e.g. it has `userId`) but carries `traits` as an array/string:

- `traits: [{ email: "x" }]` → `Object.keys([...]) === ["0"]` → attributes become `{ "0": { email: "x" } }`
- `traits: ["a", "b", "c"]` → attributes become `{ "0": "a", "1": "b", "2": "c" }`
- `traits: "hello"` → attributes become `{ "0": "h", "1": "e", ... }`

No crash (the `|| {}` guards null/falsy, and `Object.keys` on a string/array yields index keys), but a malformed attribute payload is silently sent to Customer.io instead of the event being rejected. The PR's target case — `context` as an array — is handled gracefully (`get` returns `undefined` throughout); only the `traits` variant regresses.

PR #5380, opened 2026-07-20
fix(customerio): accept non-object traits and context in v2

maheshkutty:

The v2 event-stream schema validated `traits` and `context.traits` against an object shape, so events carrying either as a non-object — some sources send `context` as an array — were rejected with "Expected object, received array" before reaching the transform.

Relax both to `z.unknown()` so they pass through, and read them in the refine via `get-value`, which returns undefined on a non-object root instead of throwing. Email and mappedToDestination detection are unchanged.

Resolves INT-XXX

N/A

N/A

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist

### DIFF
--- a/src/v0/destinations/customerio/v2/types.ts
+++ b/src/v0/destinations/customerio/v2/types.ts
@@ -1,4 +1,5 @@
 import { z } from 'zod';
+import get from 'get-value';
 import { makeRouterInputSchema } from '../../../../services/destination/nativeBatching/batchDestination';
 import { RECORD_IDENTIFIER_KEYS } from './config';
 import { CustomerIODestinationConfigSchema, CustomerIOConnectionConfigSchema } from '../types';
@@ -34,8 +35,6 @@ export type CustomerIOV2Payload = {
   [key: string]: unknown;
 };
 
-const emailTraitSchema = z.object({ email: z.unknown() }).passthrough().nullish();
-
 const recordMessageSchema = z
   .object({
     type: z.literal('record'),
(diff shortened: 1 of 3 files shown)


shekhar-rudder:
**Code review**: no correctness bugs. Verified empirically that `get-value` returns `undefined` for non-object intermediates (so the `refine` change is equivalent to the old optional chaining) and that `set-value` never throws on non-object targets — identifiers still resolve via the `getFieldValueFromMessage` precedence fallbacks. The change is a clean, non-breaking fix for the batching flow (for non-object `traits` it replaces index-keyed junk with `{}`), and the `context.traits`-shadowing quirk is pre-existing and identical in the non-batching flow (`util.ts` `identifyResponseBuilder`), so out of scope here.

**Suggestion — add a regression test.** Please cover the exact input class this PR targets in the batching flow, so we don't regress:

- `traits` as an **array** (e.g. `['a','b']`) → asserts no index-keyed `{ "0": "a", "1": "b" }` junk in the emitted attributes.
- `traits` as a **primitive** (string/number) → asserts it contributes no free-form attributes.
- `context` as a **non-object** (e.g. an array) with a valid top-level identifier → asserts the message still passes schema validation (the `z.unknown()` relaxation) and builds a valid payload.
- a `mappedToDestination`/RETL event with non-object `traits` → asserts the identifier still resolves (via `context.externalId`) and nothing throws.

These belong alongside the existing v2 batching integration tests (`test/integrations/destinations/customerio/router/`).

shekhar-rudder (approved):

PR #5384, opened 2026-07-20
feat(braze): emit 296 per job on /users/track partial-failure responses

ItsSudip:

Extends the v1 Braze networkHandler with per-job "Delivered with Warning" (HTTP **296**) emission for `/users/track` partial-failure responses. Braze surfaces per-item failures inside a 2xx batch via `errors[input_array, index]`; this PR wires the correlation from those entries back to originating jobs via a per-metadata `destInfo` positional map, so each affected job emits `statusCode: 296` carrying the Braze `error.type` verbatim.

**Handler flow** (only the `2xx + message="success" + errors[]` branch changed; happy path, application-level errors, and non-2xx paths are untouched):

1. **Endpoint dispatch** — check `destinationRequest.endpointPath === 'users/track'`. Sub/merge responses continue to receive uniform per-job outcomes.
2. **Defensive fallback** — if any metadata is missing `destInfo` (payloads produced when the router-transform side isn't populating it) OR `destinationRequest` isn't supplied, skip correlation and fall back to uniform-200 rather than mis-attribute a warned index to a random job. **This makes the change safely shippable ahead of the router-transform side** — today's payloads have no `destInfo`, so the fallback keeps behavior bit-identical to `develop`.
3. **Correlation** — build three `Map<index, error.type>` maps from `errors[]` (grouped by `input_array`: events / attributes / purchases). For each metadata, walk its `destInfo.eventsIndices` / `.attributesIndices` / `.purchasesIndices` and collect every matching `error.type`.
4. **Emission** — jobs with any warned-index intersection get `statusCode: 296` and an `error` field that concatenates every matching Braze `error.type` (join `'; '`, encounter order events → attributes → purchases, no deduplication). Non-matching jobs get 200 with the full response body.
5. **Metric** — `braze_delivered_with_warning` (labelled `destination_id`) incremented per 296. Existing `braze_partial_failure` counter still fires on every partial-failure branch entry.

Resolves [INT-6634](https://linear.app/rudderstack/issue/INT-6634)

The 296 correlation activates automatically once the router-transform side (tracked separately in INT-6808 / PR #5370) starts populating `destInfo` on outgoing metadata. Until then, the defensive fallback keeps `develop` behavior intact — no ordering constraint between the two PRs on the transformer side.

- **Partial-failure branch behavior change (gated by presence of `destInfo`)**: previously the branch just incremented `braze_partial_failure` and returned 200 for every job. Now, for `/users/track` responses whose metadata carries `destInfo`, it emits 296 per affected job. Every other code path (happy 2xx, application-level 2xx errors, 4xx, 5xx) is unchanged.
- `BrazeResponseHandlerParams` gains an optional `destinationRequest?: ProxyV1Request` field. Optional so unit-test call sites don't have to construct one; production always supplies it (`nativeIntegration.ts` forwards `deliveryRequest` verbatim).
- New `BrazeErrorEntry` type in `types.ts` (`{ type, input_array, index }`) — narrows what was previously `errors?: unknown[]` on the response type. Matches the shape Braze's `/users/track` API documents.

N/A

N/A — all changes are Braze-scoped.

- No extra IO. Correlation is a per-batch, in-memory scan over `errors[]` + `rudderJobMetadata` — O(errors + jobs × avg-indices-per-job), same order as the existing loop.
- Concatenation with no deduplication keeps the per-hit count informative for downstream consumers (two warned purchases with the same error type surface twice — this is intentional).
- `destInfo` shape is read defensively (`Array.isArray` + `typeof number` guard); a malformed indices field on one metadata is silently ignored rather than throwing.

@coderabbitai review

<hr>

### Developer checklist

- [x] **No breaking changes are being introduced.** — without `destInfo` on incoming metadata (today's state), the defensive fallback preserves current behavior end-to-end.

- [x] All related docs linked with the PR? — Linear ticket referenced.

- [x] All changes manually tested? — braze v0 + v1 unit tests + component tests all pass locally.

- [ ] Any documentation changes needed with this change? — none required; the 296 semantics are documented on the server-side ticket that consumes them.

- [x] Is the PR limited to 10 file changes? — 3 files.

- [x] Is the PR limited to one linear task? — INT-6634 only.

- [x] Are relevant unit and component test-cases added in **new readability format**? — 17 responseHandler unit tests (up from 6): happy path, defensive fallback (no / some / wrong-endpoint / missing-request), single-input-array correlation, cross-input-array correlation, multi-index concatenation, no-dedup contract, malformed-destInfo tolerance, order/identity preservation.

### Reviewer checklist

### DIFF
--- a/src/v0/destinations/braze/types.ts
+++ b/src/v0/destinations/braze/types.ts
@@ -9,6 +9,7 @@ import {
   MultiBatchRequestOutput,
   ProcessorTransformationOutput,
   ProxyMetdata,
+  ProxyV1Request,
 } from '../../../types/destinationTransformation';
 
 // Braze User Alias Object
@@ -143,15 +144,35 @@ export interface BrazeSubscriptionGroup {
   phones?: string[];
 }
 
+// Single entry in Braze's per-item partial-failure array. Braze surfaces
+// per-item validation failures inside an otherwise-2xx `/users/track`
+// response as `errors[i] = { type, input_array, index }` — the network
+// handler correlates each entry back to an originating job by looking up
+// `input_array` + `index` against the per-metadata `destInfo` positional
+// map populated on the router-transform side.
+export interface BrazeErrorEntry {
+  type: string;
+  input_array: string;
+  index: number;
+}
+
 export interface BrazeResponseHandlerParams {
   destinationResponse: {
     response?: {
       message?: string;
-      errors?: unknown[];
+      errors?: BrazeErrorEntry[];
     };
     status: number;
   };
   rudderJobMetadata: ProxyMetdata[];
+  // The framework's `deliver` step (nativeIntegration.ts) always forwards the
+  // original ProxyV1Request as `destinationRequest`. The v1 networkHandler
+  // uses its `endpointPath` (e.g. `'users/track'`) to decide whether to run
+  // the 296 per-item correlation logic — Braze's `/users/track` is the only
+  // endpoint that returns per-entry `errors[]`. Optional so unit-test call
+  // sites that don't need endpoint-dispatch can omit it; when absent the
+  // handler skips correlation and falls back to uniform per-job outcomes.
+  destinationRequest?: ProxyV1Request;
 }
 
 export interface BrazeUser extends BrazeUserAttributes {
--- a/src/v0/destinations/braze/types.ts
+++ b/src/v0/destinations/braze/types.ts
@@ -150,7 +150,7 @@ export interface BrazeSubscriptionGroup {
 // handler correlates each entry back to an originating job by looking up
 // `input_array` + `index` against the per-metadata `destInfo` positional
 // map populated on the router-transform side.
-export interface BrazeErrorEntry {
+export interface BrazeError {
   type: string;
   input_array: string;
   index: number;
@@ -160,7 +160,7 @@ export interface BrazeResponseHandlerParams {
   destinationResponse: {
     response?: {
       message?: string;
-      errors?: BrazeErrorEntry[];
+      errors?: BrazeError[];
     };
     status: number;
   };
--- a/src/v1/destinations/braze/networkHandler.ts
+++ b/src/v1/destinations/braze/networkHandler.ts
@@ -4,10 +4,53 @@ import { TransformerProxyError } from '../../../v0/util/errorTypes';
 import { TAG_NAMES } from '../../../v0/util/tags';
 import { isHttpStatusSuccess } from '../../../v0/util/index';
 import stats from '../../../util/stats';
-import type { DeliveryJobState, DeliveryV1Response, ProxyMetdata } from '../../../types';
-import { BrazeResponseHandlerParams } from '../../../v0/destinations/braze/types';
+import type {
+  DeliveryJobState,
+  DeliveryV1Response,
+  ProxyMetdata,
+  ProxyV1Request,
+} from '../../../types';
+import { BrazeErrorEntry, BrazeResponseHandlerParams } from '../../../v0/destinations/braze/types';
 
 const DESTINATION = 'braze';
+
+// HTTP 296 — "Delivered with Warning". Emitted per originating job when Braze
+// returns a 2xx with `message: "success"` + a non-empty `errors[]` that
+// identifies specific per-item failures inside a /users/track call.
+// Contract: rudder-server treats 296 as a delivered-but-warned outcome for
+// alerting; the batch itself remains a success at the transport level.
+const HTTP_STATUS_DELIVERED_WITH_WARNING = 296;
+
+// Braze's `endpointPath` value for the /users/track endpoint — the only
+// endpoint whose response carries per-entry `errors[]` correlatable back to
+// originating jobs. Sub/merge responses return a single top-level message
+// that applies uniformly to the whole call, so they skip 296 correlation
+// entirely and fall through to the uniform per-job outcome path.
+const TRACK_ENDPOINT_PATH = 'users/track';
+
+// Braze's `errors[i].input_array` values on /users/track responses,
+// corresponding to the three sub-arrays in the request body. Each maps to
+// the `destInfo` field name populated by the router transform.
+const TRACK_INPUT_ARRAYS = ['events', 'attributes', 'purchases'] as const;
+type TrackInputArray = (typeof TRACK_INPUT_ARRAYS)[number];
+
+const isTrackInputArray = (value: string): value is TrackInputArray =>
+  (TRACK_INPUT_ARRAYS as readonly string[]).includes(value);
+
+// `destInfo` field name for each track input_array. Kept as a plain lookup
+// so both the correlation logic and any future validation stay consistent.
+const DEST_INFO_KEY: Record<TrackInputArray, string> = {
+  events: 'eventsIndices',
+  attributes: 'attributesIndices',
+  purchases: 'purchasesIndices',
+};
+
+// True when the delivery request was aimed at /users/track. Uses the
+// framework-populated `endpointPath` on the ProxyV1Request. Absence of the
+// field (or a request the framework didn't attach) is treated as non-track:
+// the handler falls back to uniform-per-job outcomes rather than mis-attributing.
+const isTrackEndpoint = (destinationRequest: ProxyV1Request | undefined): boolean =>
+  destinationRequest?.endpointPath === TRACK_ENDPOINT_PATH;
 
 /**
  * Maps every job in `rudderJobMetadata` to a `DeliveryJobState` using the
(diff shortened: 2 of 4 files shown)


src/v0/destinations/braze/types.ts
```diff
@@ -143,15 +144,35 @@ export interface BrazeSubscriptionGroup {
   phones?: string[];
 }
 
+// Single entry in Braze's per-item partial-failure array. Braze surfaces
+// per-item validation failures inside an otherwise-2xx `/users/track`
+// response as `errors[i] = { type, input_array, index }` — the network
+// handler correlates each entry back to an originating job by looking up
+// `input_array` + `index` against the per-metadata `destInfo` positional
+// map populated on the router-transform side.
+export interface BrazeErrorEntry {
```
achettyiitr: What's the entry here?
```suggestion
export interface BrazeError {
```

src/v1/destinations/braze/networkHandler.ts
```diff
@@ -4,11 +4,54 @@ import { TransformerProxyError } from '../../../v0/util/errorTypes';
...
+import { BrazeErrorEntry, BrazeResponseHandlerParams } from '../../../v0/destinations/braze/types';
 
 const DESTINATION = 'braze';
 
+// HTTP 296 — "Delivered with Warning". Emitted per originating job when Braze
+// returns a 2xx with `message: "success"` + a non-empty `errors[]` that
+// identifies specific per-item failures inside a /users/track call.
+// Contract: rudder-server treats 296 as a delivered-but-warned outcome for
+// alerting; the batch itself remains a success at the transport level.
+const HTTP_STATUS_DELIVERED_WITH_WARNING = 296;
```
achettyiitr: Can this be moved to a global utility, as it needs to be used by others as well, instead of replicating it?

src/v1/destinations/braze/networkHandler.ts
```diff
@@ -70,6 +199,39 @@ const responseHandler = (params: BrazeResponseHandlerParams): DeliveryV1Response
...
+      const warned = buildWarnedIndexMaps(errors);
+      return {
+        status,
+        message: `Request for ${DESTINATION} Processed Successfully`,
+        response: buildTrackPartialFailureStates(
+          response,
+          rudderJobMetadata,
+          warned,
+          destinationId,
+        ),
```
achettyiitr: Move the metric side-effect out of buildTrackPartialFailureStates. This also drops the destinationId parameter from the builder.

```
// builder stays pure — no stats, no destinationId param
const buildTrackPartialFailureStates = (
  response: unknown,
  rudderJobMetadata: ProxyMetdata[],
  warned: Record<TrackInputArray, WarnedIndexMap>,
): DeliveryJobState[] => { /* ...no stats... */ };

// caller decides how to count
const states = buildTrackPartialFailureStates(response, rudderJobMetadata, warned);
const warnedCount = states.filter((s) => s.statusCode === HTTP_STATUS_DELIVERED_WITH_WARNING).length;
if (warnedCount > 0) {
  stats.increment('braze_delivered_with_warning', { destination_id: destinationId, count: warnedCount });
}
```

src/v1/destinations/braze/networkHandler.ts
```diff
@@ -4,11 +4,54 @@ import { TransformerProxyError } from '../../../v0/util/errorTypes';
...
+const TRACK_ENDPOINT_PATH = 'users/track';
+
+// Braze's `errors[i].input_array` values on /users/track responses,
+// corresponding to the three sub-arrays in the request body. Each maps to
+// the `destInfo` field name populated by the router transform.
+const TRACK_INPUT_ARRAYS = ['events', 'attributes', 'purchases'] as const;
+type TrackInputArray = (typeof TRACK_INPUT_ARRAYS)[number];
+
+const isTrackInputArray = (value: string): value is TrackInputArray =>
+  (TRACK_INPUT_ARRAYS as readonly string[]).includes(value);
```
achettyiitr: Can't we use set here?

const TRACK_INPUT_ARRAY_SET = new Set<string>(TRACK_INPUT_ARRAYS);
const isTrackInputArray = (v: string): v is TrackInputArray => TRACK_INPUT_ARRAY_SET.has(v);

src/v1/destinations/braze/networkHandler.ts
```diff
@@ -48,15 +185,7 @@ const responseHandler = (params: BrazeResponseHandlerParams): DeliveryV1Response
...
-  // Guard 2: partial failure — destination accepted the request but some items
-  // within the batch were invalid. Braze signals this with message='success'
-  // and a non-empty errors array. Emit a metric so ops can track frequency;
-  // the batch is still considered delivered (fall through to success).
-  if (brazeMessage === 'success' && hasErrors) {
-    stats.increment('braze_partial_failure');
-  }
-
-  // Guard 3: application-level error — destination returned 2xx but with an
+  // Guard 2: application-level error — destination returned 2xx but with an
```
achettyiitr: Is it possible that `brazeMessage` is a failure but `hasErrors` is false? Currently, in that case, we will fall back to the success scenario. It can be either with the parsing logic w.r.t. `errors`?

achettyiitr: Shall we do something like this?

```
if (brazeMessage !== 'success') {
    throw new TransformerProxyError(
      `Request failed for ${DESTINATION} with status: ${status}`,
      status,
      { [TAG_NAMES.ERROR_TYPE]: getDynamicErrorType(status) },
      destinationResponse,
      '',
      buildJobStates(response, status, rudderJobMetadata),
    );
  }
```

src/v1/destinations/braze/networkHandler.ts
```diff
@@ -70,6 +199,39 @@ const responseHandler = (params: BrazeResponseHandlerParams): DeliveryV1Response
...
+    const destinationId = rudderJobMetadata[0]?.destinationId ?? '';
+
+    // Only /users/track responses carry per-item errors correlatable back to
+    // originating jobs. Subscription-groups and alias-merge responses surface
+    // a single top-level message that applies uniformly to the whole call;
+    // for those, every job gets 200. We dispatch on the outgoing endpoint
+    // (via destinationRequest.endpointPath) rather than inspecting
+    // error.input_array values so that if Braze ever adds a new sub-array on
+    // /users/track, correlation still runs.
+    if (isTrackEndpoint(destinationRequest) && allMetadataHaveDestInfo(rudderJobMetadata)) {
```
achettyiitr: Why does all metadata need to have destInfo? We can still work without that right with partial metadata having the information?

src/v1/destinations/braze/networkHandler.ts
```diff
@@ -70,6 +199,39 @@ const responseHandler = (params: BrazeResponseHandlerParams): DeliveryV1Response
...
+      const warned = buildWarnedIndexMaps(errors);
+      return {
+        status,
+        message: `Request for ${DESTINATION} Processed Successfully`,
```
achettyiitr: Can we create a constant for this? Even for the failure message?

src/v1/destinations/braze/networkHandler.ts
achettyiitr, line 1: Can't we also send workspaceId for both metrics?

src/v1/destinations/braze/networkHandler.ts
```diff
@@ -28,8 +71,102 @@ const buildJobStates = (
...
+const readIndicesFor = (
+  metadata: ProxyMetdata,
+  inputArray: TrackInputArray,
+): number[] | undefined => {
+  const info = metadata.destInfo;
+  if (!info) return undefined;
+  const raw = info[DEST_INFO_KEY[inputArray]];
+  if (!Array.isArray(raw)) return undefined;
+  return raw.every((n) => typeof n === 'number') ? (raw as number[]) : undefined;
+};
```
achettyiitr, line 85: Minor, but the .every(...) ? (raw as number[]) : undefined ternary plus the as number[] cast can be flattened:
```suggestion
const readIndicesFor = (metadata: ProxyMetdata, inputArray: TrackInputArray): number[] | undefined => {
  const raw = metadata.destInfo?.[DEST_INFO_KEY[inputArray]];
  if (!Array.isArray(raw) || !raw.every((n) => typeof n === 'number')) return undefined;
  return raw as number[];
};
```

src/v1/destinations/braze/networkHandler.ts
```diff
@@ -28,8 +71,102 @@ const buildJobStates = (
...
+  metadata: ProxyMetdata,
+  warned: Record<TrackInputArray, WarnedIndexMap>,
+): string[] => {
+  const hits: string[] = [];
+  for (const inputArray of TRACK_INPUT_ARRAYS) {
+    const indices = readIndicesFor(metadata, inputArray);
+    if (indices) {
+      for (const idx of indices) {
+        const errorType = warned[inputArray].get(idx);
+        if (typeof errorType === 'string') hits.push(errorType);
```
achettyiitr: WarnedIndexMap is already a map<number, string>, can't we just check for undefined?
```suggestion
        const errorType = warned[inputArray].get(idx);
          if (errorType) hits.push(errorType);
```

src/v1/destinations/braze/networkHandler.ts
```diff
@@ -28,8 +71,102 @@ const buildJobStates = (
...
+      }
+    }
+  }
+  return hits;
+};
+
+// Separator for concatenating multiple warned error.type strings on a single
+// 296 job — chosen for readability when the field is logged verbatim.
+const ERROR_TYPE_JOIN = '; ';
```
achettyiitr, line 130: Can this be moved to some common utility?

ItsSudip: Used only once here today and it's Braze-specific (chosen for the concatenated `error.type` field on 296 outcomes). Prefer to keep it local for now — happy to hoist it to a shared util the moment a second caller shows up.

src/v0/destinations/braze/types.ts
ItsSudip: Done — renamed to `BrazeError`.

src/v1/destinations/braze/networkHandler.ts
ItsSudip: Done — switched to a `Set` and dropped the `as readonly string[]` widening cast that the old `Array.includes` required.

src/v1/destinations/braze/networkHandler.ts
ItsSudip: Done — collapsed to the optional-chain form and extracted an `isNumberArray` type predicate so the `as number[]` cast could go away entirely (TS `Array.every` doesn't narrow the array's element type; a predicate does).

src/v1/destinations/braze/networkHandler.ts
ItsSudip: Done — `WarnedIndexMap` already types the value as `string`, so the truthy check is enough.

src/v1/destinations/braze/networkHandler.ts
ItsSudip: Done — added `DELIVERED_WITH_WARNING: 296` to `HTTP_STATUS_CODES` in `src/v0/util/constant.js` next to the existing rudder-server-recognised 2xx codes (`FILTER_EVENTS`, `SUPPRESS_EVENTS`). The local constant here is removed and the single usage now reads from the shared enum.

src/v1/destinations/braze/networkHandler.ts
ItsSudip: Done — extracted `SUCCESS_MESSAGE` (constant) and `failureMessage(status)` (builder). All four call sites use them.

src/v1/destinations/braze/networkHandler.ts
ItsSudip: Done — Guard 2 now throws whenever `brazeMessage !== 'success'`, so `{ message: 'failure' }` with an empty/absent `errors` no longer slips through as a silent 200. Added two regression tests for the widened branch.

src/v1/destinations/braze/networkHandler.ts
ItsSudip: Good point — dropped the all-or-nothing `allMetadataHaveDestInfo` gate. `readIndicesFor` returns `undefined` when destInfo is missing and `collectMatchingErrorTypes` produces no hits, so per-job correlation already handles the partial case: jobs with destInfo → 296/200 as attributed; jobs without → 200 (nothing to correlate). This is strictly more informative than the previous "silently downgrade everyone to 200 if anyone is missing destInfo". Rewrote the corresponding unit test to codify the new mixed-batch semantics.

src/v1/destinations/braze/networkHandler.ts
ItsSudip: Done — builder is pure now (no `stats`, no `destinationId`/`workspaceId` params). Caller counts warned states and emits a single aggregated metric.

One small deviation from the snippet: switched from `stats.increment('braze_delivered_with_warning', { …, count: warnedCount })` to `stats.counter('braze_delivered_with_warning', warnedCount, { … })`. `stats.increment` (`src/util/stats.js`) always adds 1 per call regardless of `tags`, so passing `count` inside the tags object would (a) not actually add N and (b) turn the count into a high-cardinality label. `stats.counter(name, delta, tags)` takes the delta as an integer parameter, which matches the intent.

src/v1/destinations/braze/networkHandler.ts
ItsSudip: Done — both `braze_partial_failure` and `braze_delivered_with_warning` now carry `{ destination_id, workspace_id }` labels.

src/v1/destinations/braze/networkHandler.ts
achettyiitr: My bad, count didn't need to go as part of tags. We need to use counter.

achettyiitr (approved):

PR #5385, opened 2026-07-21
fix(ivm): cap concurrent evaluations per isolate and queue the excess

shekhar-rudder:

Caps how many custom-mapping evaluations run concurrently on a single `isolated-vm` isolate and queues the excess.

`IvmScriptRunner` gains one optional knob — `maxConcurrentExecutions`. It defaults to unlimited, so `custom_audience` and every other consumer keep today's behaviour; only the `custom_mappings_ivm` pool opts in.

Resolves INT-6832

An isolate runs JS on one thread against one fixed heap. Extra concurrency on it buys no throughput — it only keeps more structured-clone argument copies and evaluation working sets alive on that heap at the same time. When the total breaches `memoryLimit`, V8 disposes the **whole isolate**, so every in-flight evaluation fails at once with a retryable 500, not just the one that tipped it. That amplification is what surfaced as the intermittent 500 spikes on Lovable's HTTP destination.

Nothing bounded that fan-in before this. `execute()` had promise-coalescing on isolate *creation* only; once an isolate was warm, unlimited calls could re-enter it. Callers actively fan out — the HTTP destination maps headers, query params and payload as three separate evaluations per event (`procWorkflow.yaml:39,50,59`), and `ga4_v2` `Promise.all`s over its mapping list (`customMappingsHandler.js:110`).

Bounding admission converts "the isolate dies and takes every in-flight event with it" into "the excess waits". Queued callers hold their args on the Node heap, where those args already were — the queue itself costs the isolate nothing.

This is follow-up #1 from the INT-6832 thread. Follow-up #2 (compile-caching the mapping template) is #5372; the two are independent and can merge in either order.

Only the `custom_mappings_ivm` pool changes behaviour, and only past the cap. **Nothing is ever shed** — the queue is unbounded by construction, so an overload costs latency rather than dropped events. `acquireSlot()` cannot reject; it only ever admits or queues.

One new tunable (falling back to unlimited — i.e. today's behaviour — if absent or unparseable, rather than wedging the sandbox):

| Env var | Default | Rationale |
|---|---|---|
| `CUSTOM_MAPPINGS_IVM_MAX_CONCURRENT_EXECUTIONS` | 100 | Sized from prod — see below. |

The cap alone is what protects the isolate heap, and a single-threaded isolate drains its queue at its own throughput either way, so an explicit queue bound would only have bought a strictly worse failure mode (a dropped event instead of a slower one).

#### Where the cap of 100 comes from

Sized against `enterprise-eu` / `rudder-eu-central-1b`, the busiest namespace for this pool by a wide margin (~380x `eu-central-1a`):

| Observation (7d) | Value |
|---|---|
| Peak single-pod eval rate | ~13.8k/s |
| Warm isolates per pod | 18 (cache max 20) |
| Observed heap per isolate | ~5.6 MB baseline |
| `ivm_platform_error`, `custom_mappings_ivm` | 879, **all in one hour, all one workspace** |

That error profile is the whole argument for this PR: 879 failures inside a single hour and zero across the rest of the week is not a steady error rate, it is one isolate dying and taking every in-flight evaluation with it.

An isolate runs on one thread, so sustaining ~13.8k evals/s implies a sub-100µs evaluation, which by Little's Law puts steady-state in-flight depth near 1. Concurrency only builds when arrivals transiently outrun the isolate — exactly the OOM regime. So 100 sits ~100x above anything normal traffic reaches (it will not throttle the hot workspace) and ~16x below the ~1600-concurrency OOM onset measured in #5372. Memory agrees: ~5.9 MB of eval working set at concurrency 100, on top of the ~5.6 MB baseline, against the 32 MB prod limit.

#### Measured under load

Verified locally against the real `http` destination path (single process, one workspace, 6s per phase). The cap was set to **8** rather than 100 for the run: by the Little's Law argument above, steady-state depth sits near 1, so a laptop cannot drive an isolate to depth 100 — lowering the cap exercises the identical code path at reachable load. 180,674 requests / 542,022 evaluations, **all HTTP 200**.

| offered concurrency | evals/s | queued/s | mean wait | in-flight depth |
|---|---|---|---|---|
| 1 | 6,900 | 0 | — | 0.39 |
| 4 | 14,803 | 0 | — | 0.84 |
| 8 | 16,518 | 0 | — | 0.94 |
| 16 | 17,003 | 0 | — | 0.97 |
| 32 | 17,551 | 13,553 | 0.39ms | **7.84** |
| 64 | 17,509 | 17,651 | 1.11ms | **20.43** |

Three things the run confirms:

- **Throughput is flat from concurrency 8 onward** (~17.5k evals/s) while offered load grows 8x. Extra concurrency bought zero throughput and became queue depth — the premise of the PR, measured. The implied ~57µs service time is consistent with the sub-100µs figure inferred from prod.
- **The gate is silent until it is genuinely saturated.** `ivm_execution_queued` stayed at exactly 0 through four phases of rising load. Past the cap, 312,295 evaluations queued and `ivm_execution_queue_wait_count` equalled that number exactly — every queued call was admitted, nothing lost or double-counted, and `ivm_platform_error` stayed at 0 across all 542k evaluations.
- **In-flight depth crosses the cap where queueing starts.** Depth above is Little's Law (throughput x residence), derived independently of the gate's own counters, and it crosses 8 at precisely the phase where the counter switches on.

N/A — no new packages; the gate is ~60 lines of plain promise plumbing.

`IvmScriptRunner` (shared by `custom_mappings` and `custom_audience`) gains the option above, opt-in per consumer.

`prometheus.js` gains a static registration for `ivm_execution_queue_wait` with explicit buckets — see the metrics bullet below for why.

- **Slot handoff, not free-and-wake.** `release()` hands its slot directly to the next waiter and leaves `slotsInUse` untouched. Decrementing *and* waking a waiter would let the accounting drift and admit past the cap. There is a regression test for exactly that — it fails with peak 3 against a cap of 2 under the buggy variant. The counter is named for slots *reserved*, not work in progress: a slot in transit between a departing caller and a waiter stays counted across the microtask gap in which neither is executing.
- **Waiters are FIFO.** `pop()` would be marginally cheaper, but it starves the oldest waiter under sustained saturation, and nothing bounds queue time here (`execTimeoutMs` covers evaluation only, and the queue is unbounded). The O(n) `shift()` is a pointer memmove on a queue that Little's Law keeps near-empty in steady state — orders of magnitude below the ~100µs evaluation it gates.
- **Gates are per isolate** (keyed like the isolate cache, i.e. per workspace) and dropped once fully idle, so the map stays bounded by the live isolate set rather than every workspace ever seen. A queued caller resolves with its own release closure and its own `cacheKey`, so it can never bind to another tenant's isolate.
- **New metrics:** `ivm_execution_queued` (counter) and `ivm_execution_queue_wait` (timer), both carrying the full `(workspaceId, functionName, cache)` label set that `ivm_platform_error` already uses, so all three join on the same series. The gate is per isolate — i.e. per workspace — and these two are the only overload signal there is, so "which tenant is at its cap" has to be answerable from them. `ivm_execution_queued` is incremented **at enqueue**, so a backlog shows up the moment it forms; the wait timer only samples on admission and so lags a burst by however long the queue takes to drain. `ivm_execution_queued` is the one to alert on. Cardinality is bounded by the workspaces that actually reach the cap — rare by design, since the cap sits far above normal traffic — not by call volume; `functionName` is a hardcoded expression literal per call site (three exist repo-wide, only one on a gated pool), so it adds no fan-out.
- **`ivm_execution_queue_wait` is declared statically with explicit buckets.** Left to `prometheus.js`'s dynamic path it inherits prom-client's default buckets, whose first boundary is 5ms — and a gated isolate drains far faster than that. In the load run above, **99.45% of 312k observations (310,578) landed in that single first bucket**, with the remaining 0.55% in the next and every higher bucket empty; `histogram_quantile()` on that series interpolates inside one bucket and cannot separate a healthy 0.4ms wait from a degraded 4ms one. The declared buckets start at 0.5ms and run to 1s; re-measured under the same load the distribution spans five buckets (p50 ~3.8ms, p99 ~19ms). Declaring it also pins the label set rather than inferring it from whichever call site observes first.
- **Metric emission ordering.** `ivm_execution_queued` is emitted before the waiter promise is constructed, and the wait timer after `resolve()`, so a throwing metric client can never strand a waiter in the queue holding a slot that was handed to it and never released.
- **Out of scope, deliberately:** the code default for `CUSTOM_MAPPINGS_IVM_MEMORY_MB` is still 16 while prod is overridden to 32 via helm — separate concern, prod already carries the bump. `custom_audience` stays ungated; enabling it there is a one-line config change if wanted.

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist

### DIFF
--- a/src/util/prometheus.js
+++ b/src/util/prometheus.js
@@ -981,6 +981,18 @@ class Prometheus {
         labelNames: ['cache'],
       },
       {
+        name: 'ivm_execution_queue_wait',
+        help: 'Time an evaluation waited for a free isolate concurrency slot (seconds)',
+        type: 'histogram',
+        labelNames: ['functionName', 'workspaceId', 'cache'],
+        // Declared explicitly because the default bucket set starts at 5ms, and a gated
+        // isolate drains far faster than that: measured locally, 99.45% of 312k waits landed
+        // in that single first bucket, which makes histogram_quantile() interpolate inside it
+        // and report nothing useful. These start at 0.5ms so a healthy sub-millisecond wait is
+        // distinguishable from a degraded one, and run to 1s to keep a real backlog on-scale.
+        buckets: [0.0005, 0.001, 0.0025, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1],
+      },
+      {
         name: 'fetchV2_call_duration',
         help: 'fetchV2_call_duration',
         type: 'histogram',
--- a/src/util/prometheus.js
+++ b/src/util/prometheus.js
@@ -989,8 +989,10 @@ class Prometheus {
         // isolate drains far faster than that: measured locally, 99.45% of 312k waits landed
         // in that single first bucket, which makes histogram_quantile() interpolate inside it
         // and report nothing useful. These start at 0.5ms so a healthy sub-millisecond wait is
-        // distinguishable from a degraded one, and run to 1s to keep a real backlog on-scale.
-        buckets: [0.0005, 0.001, 0.0025, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1],
+        // distinguishable from a degraded one, and extend to 10s so a real backlog (queue depth
+        // growing behind timed-out executions) stays on-scale instead of pinning p99 at the top
+        // finite bucket.
+        buckets: [0.0005, 0.001, 0.0025, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10],
       },
       {
         name: 'fetchV2_call_duration',
--- a/src/util/customMappings/sandboxClient.ts
+++ b/src/util/customMappings/sandboxClient.ts
@@ -28,6 +28,14 @@ const getRunner = (): IvmScriptRunner => {
       cacheName: 'custom_mappings_ivm',
       maxSize: Number.parseInt(process.env.CUSTOM_MAPPINGS_IVM_CACHE_MAX_SIZE || '20', 10),
       ttlMs: Number.parseInt(process.env.CUSTOM_MAPPINGS_IVM_CACHE_TTL_MS || '300000', 10),
+      // A workspace's evals all share one fixed-size isolate heap. Left ungated, a burst piles
+      // every in-flight eval onto that heap until V8 disposes the isolate and fails all of them
+      // at once — the amplification behind INT-6832. Excess evals queue instead; the queue is
+      // unbounded, so a burst costs latency rather than dropped events.
+      maxConcurrentExecutions: Number.parseInt(
+        process.env.CUSTOM_MAPPINGS_IVM_MAX_CONCURRENT_EXECUTIONS || '100',
+        10,
+      ),
     });
   }
   return runner;
--- a/src/util/ivm/scriptRunner.gate.test.ts
+++ b/src/util/ivm/scriptRunner.gate.test.ts
@@ -1,0 +1,345 @@
+/**
+ * Unit tests for IvmScriptRunner's per-isolate execution gate.
+ *
+ * An isolate has one fixed-size heap shared by every concurrent evaluation on it, so an
+ * ungated burst can breach `memoryLimit` and make V8 dispose the whole isolate — failing
+ * every in-flight call at once. These tests pin the gate that caps that concurrency.
+ *
+ * `isolated-vm` is mocked with an evalClosure that never settles on its own, so a test can
+ * hold executions open and observe exactly how many the gate lets run at a time.
+ *
+ * Kept separate from `scriptRunner.test.ts` (the platform-error counter suite) because the two
+ * need incompatible module mocks: that file stubs `fs` and `DisposableCache`, whereas these
+ * tests exercise the real cache so gate teardown can be observed against real eviction.
+ */
+
+interface MockPendingExecution {
+  code: string;
+  finish: (value: unknown) => void;
+  fail: (err: unknown) => void;
+}
+
+// Executions that have entered the isolate and not yet settled. Tests drain this to
+// free slots. Reset in beforeEach.
+let mockPendingExecutions: MockPendingExecution[] = [];
+let mockRunningNow = 0;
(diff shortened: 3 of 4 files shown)


krishna2020 (dismissed):

saikumarrs (dismissed):

src/util/prometheus.js
```diff
@@ -980,6 +980,18 @@ class Prometheus {
...
+        name: 'ivm_execution_queue_wait',
+        help: 'Time an evaluation waited for a free isolate concurrency slot (seconds)',
+        type: 'histogram',
+        labelNames: ['functionName', 'workspaceId', 'cache'],
+        // Declared explicitly because the default bucket set starts at 5ms, and a gated
+        // isolate drains far faster than that: measured locally, 99.45% of 312k waits landed
+        // in that single first bucket, which makes histogram_quantile() interpolate inside it
+        // and report nothing useful. These start at 0.5ms so a healthy sub-millisecond wait is
+        // distinguishable from a degraded one, and run to 1s to keep a real backlog on-scale.
+        buckets: [0.0005, 0.001, 0.0025, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1],
```
krishna2020: The low end is well argued, but the top boundary is below the range this metric exists to measure. `histogram_quantile` returns the highest *finite* boundary when the quantile lands in `+Inf`, so p99 pins at exactly `1` — a 1.1s wait and a 60s wait render identically, and a `p99 > 5s` alert can never fire.

Wait ≈ depth × service time, which your own table confirms (depth 20.43 × ~57µs ≈ 1.17ms, vs 1.11ms measured). Healthy, crossing 1s needs depth ~17.5k. But the gate exists for the degraded case: at 5ms service time it takes depth 200, and at the 500ms `execTimeoutMs` ceiling, 2 waiters.

Suggest appending `2.5, 5, 10`. Three more series per label set, and these labels only materialise for saturating workspaces anyway.

shekhar-rudder: Addressed

krishna2020 (approved):

PR #5387, opened 2026-07-23
chore: live integration test framework

achettyiitr:

### Real-destination live integration testing (HubSpot pilot)

Drives events through the full **transform → deliver** pipeline against **real destination accounts** and asserts on the genuine response — closing the contract-verification gap the mocked suite can't cover (mocks answer from static fixtures, so a destination changing its endpoint, schema, auth, or encoding ships green today).

The mocked `component.test.ts` suite is untouched and stays the merge gate. This is additive.

### What's included

- **Sibling runner** — `component.live.test.ts` + `jest.config.live.js` boot the real Koa app **without** `MockAxiosAdapter`, then per step: `POST /routerTransform` → map each output to a proxy request → `POST /v1/destinations/<dest>/proxy` → assert the delivery verdict. Run via `npm run test:live` / `test:live:coverage`.
- **Harness core** (`test/integrations/live/`):
  - `registry.ts` — enrollment by discovery (`destinations/*/live.ts` with `enabled: true`).
  - `secretResolver.ts` — reads `LIVE_SECRET_<DEST>` (JSON); throws if missing/invalid.
  - `runContext.ts` — per-run `runId` plus memoised `identity`/`email`/`now`/`register`.
  - `routerOutputToProxyRequest.ts` — ports the transform→deliver chaining rudder-server does in production.
  - `types.ts` — `LiveSpec` / `LiveScenario` / `LiveStep`.
- **Step model** — a scenario is an ordered list of steps (no lifecycle hooks): `pipeline` (seed → transform → deliver → assert), `action` (direct destination-API side effect, e.g. setup/cleanup), and `verify` (read-back assertion). Plus `retries` (for eventually-consistent routing) and `metadataOverride` (e.g. `dontBatch`).
- **HubSpot pilot** — `destinations/hs/live.ts`: 14 scenarios across event-stream and reverse-ETL, newApi and v1, covering create/update, `dontBatch`, `hsContactId`, non-unique-lookup search routing, and associations. Uses a private-app token; teardown deletes created contacts via the CRM API.
- **CI** — `.github/workflows/live-integration-tests.yml`: runs on `pull_request` and `push` to `main`/`develop`, on a self-hosted in-cluster runner, with a per-destination matrix. Credentials come from the dev Vault via GitHub-OIDC (JWT auth) — no long-lived secret in GitHub — reading `LIVE_SECRET_HS` from `engineering_shared/.../e2e_test/rudder-transformer`. `LOG_LEVEL=100` suppresses body logging.
- Credentials are required (the resolver throws if absent). See `test/integrations/live/README.md`.

- Resolves INT-6774

### DIFF
--- a/.claude/skills/live-integration-test/SKILL.md
+++ b/.claude/skills/live-integration-test/SKILL.md
@@ -1,0 +1,260 @@
+---
+name: live-integration-test
+description: Generate a destination's live (real-destination) integration test spec `test/integrations/destinations/<dest>/live.ts` by deriving its scenarios and event seeds from the destination's existing component tests (router/processor/dataDelivery `data.ts`). Use when enrolling a destination into the live suite, or when asked to "add live tests" / "generate live test cases" for a destination.
+argument-hint: <destination-folder-name>
+---
+
+# Live Integration Test — Enroll a Destination
+
+**Objective:** Create `test/integrations/destinations/<dest>/live.ts` — a `LiveSpec` whose
+`scenarios` drive events through the real `transform → deliver` pipeline against a **real
+destination account**, deriving the event seeds, config, and code-path coverage from the
+destination's existing (mocked) component tests. The mocked `component.test.ts` suite stays the
+merge gate; this suite is additive and asserts on the genuine delivery verdict.
+
+## Inputs
+
+- **Destination folder name**: `$ARGUMENTS[0]` — the folder under `test/integrations/destinations/`
+  and `src/.../destinations/`, e.g. `<dest>`.
+
+## Sources of truth (read these first)
+
+1. **Component test cases** — the raw material for scenarios:
+   - `test/integrations/destinations/<dest>/router/data.ts` and, if present,
+     `test/integrations/destinations/<dest>/processor/data.ts` (plus their imports: `config.ts`,
+     `upsertData.ts`, `errorValidationData.ts`, …). Most destinations have both a `processor/` and a
+     `router/` suite; some (e.g. router-only ones) have just `router/`. Each case is
+     `{ id, description, feature, input.request.body.input[].message, ...destination.Config..., output }`.
+     The **`message`** is the raw event and **`destination.Config`** is the config — these become the
+     scenario's `seed(ctx)` and `resolveConfig`/`configOverride`.
+   - `test/integrations/destinations/<dest>/dataDelivery/*.ts` — proxy/delivery cases (optional; not
+     every destination has them). Show what a delivered vs failed verdict looks like.
+2. **`test/integrations/destinations/<dest>/network.ts`** — the mocked API responses. This reveals
+   **which real endpoints the transform hits** and the branch each case exercises (create vs update,
+   search hit/miss, batch vs single, membership add/remove), which tells you what live
+   **setup/verify/cleanup** must do.
+3. **The transformer** — `src/v0|v1/destinations/<dest>/` or `src/cdk/v2/destinations/<dest>/`: the
+   config fields, auth mechanism, and the destination API endpoints (for setup/verify/cleanup steps).
+4. **The live harness contract** — `test/integrations/live/types.ts` (`LiveSpec`, `LiveScenario`,
+   `LiveStep`, `RunContext`), the shared `pollUntil` read-back helper
+   (`test/integrations/live/poll.ts`), and `test/integrations/live/README.md`.
+5. **The reference implementation** — read the HubSpot spec end-to-end as the worked example of
+   every pattern below. For non-trivial specs it is a **module folder**:
+   `destinations/hs/live.ts` is a one-line re-export of `destinations/hs/live/spec.ts`, with helpers
+   split across `live/api.ts` (real-API calls + auth), `live/setup.ts` (action bodies),
+   `live/verify.ts` (verify steps), and `live/profiles.ts` (trait profiles). It is a CRM/object
+   destination; adapt the shapes to your destination's category (see below).
+
+## The mapping: component case → live scenario
+
+| Component test | Live spec |
+| --- | --- |
+| mocked network via `network.ts` / `MockAxiosAdapter` | **no mock** — the runner boots the real app and hits real APIs |
+| a `data.ts` case = one input `message` + `destination.Config` + expected `output` | a **scenario** with a pipeline step whose `seed(ctx)` rebuilds that `message` |
+| hardcoded PII/ids in the message (email, userId, externalId, recordId, list/audience id) | `ctx.email()`, `ctx.identity(entity)`, `ctx.runId`, `ctx.now(offset?)`, `ctx.liveSecret.resourceIds`, or an id registered by a setup step — never a literal |
+| config variants across cases (api version, list/object type, mapping) | a base `resolveConfig` + per-scenario `configOverride(base)` |
+| a case that asserts a specific transformed request shape | a scenario asserting the event is **delivered** (verdict 2xx / 207); the shape is the transform's job, already covered by the mocked suite |
+
+**One scenario per distinct behavior, not per component case.** Collapse the many mocked cases into
+the *code paths* that matter live — these differ by destination category:
+
+- **CRM / object destinations**: create vs update; event-stream vs RETL (`mappedToDestination`);
+  lookup-by-id vs search-by-field; associations; per object type; API-version variants.
+- **Audience destinations**: add vs remove membership; list/segment creation; raw vs hashed identifiers.
+- **Conversion / event destinations**: the distinct event types and property mappings; enhanced /
+  offline conversion variants.
+- **Cross-cutting** (any category): batched vs `dontBatch`; config-driven variants.
+
+Give each a stable `id` and a one-line `description`.
+
+**Skip cases that can't be meaningfully delivered live** — pure validation/error cases (bad config,
+missing required field → 400, "more than one match" aborts, unsupported event type). Those are
+contract checks the mocked suite already owns; the live suite verifies real *delivery*, so don't
+port them.
+
+## Stateful cases: setup + verify as steps, teardown declared on the scenario
+
+If a case's code path depends on pre-existing destination state, express **setup** and **read-back**
+as ordered steps, and **declare teardown on the scenario** rather than adding a trailing cleanup step:
+
+- **update / lookup-by-id / search-by-field / remove-from-audience** → an `action` step
+  (`{ stepType: 'action', name: 'setup', run }`) creates the prerequisite state via the real
+  destination API and `ctx.register({ type, id })`s any ids; the
+  pipeline step then references `ctx.resources` for that id, or seeds the same `ctx.email()`/identity
+  the setup used.
+- **verify** → a `verify` step that reads the object/membership/association back from the
+  destination API and asserts its **fields** with jest `expect(...)` (it resolves on success and a
+  failed matcher fails the step — it returns `void`, no boolean), not just its existence: a create
+  scenario checks the object carries every property it was created with; an update scenario checks
+  the properties it changed; an association scenario checks the link actually exists between the
+  two records. Share one property/trait profile between the pipeline `seed` and the verify — a
+  `(ctx) => ({ ... })` factory referenced by both, so the seeded values and the assertion can't
+  drift (see `esContactCreateTraits` + `verifyContactProperties`, and `verifyAssociationExists`, in
+  `hs/live/verify.ts`). Poll eventually-consistent read-backs with the shared `pollUntil` helper
+  (`live/poll.ts`; `soft: true` returns the last value so the closing `expect(...)` prints a real
+  diff). Read-back is also the
+  real check for **batch** writes: a batch endpoint can return `207` (which counts as delivered)
+  even when an item fails, so the delivery verdict alone doesn't prove the write landed.
+- **cleanup** → **declare it on the scenario.** Set `cleanup: (ctx) => delete…(ctx)` on the
+  `LiveScenario`; the runner arms it at scenario start and drains it after the scenario's steps
+  finish — LIFO and best-effort (a failing cleanup is logged, not thrown), running **even if a
+  step failed**. For resources a setup action discovers dynamically, `ctx.addCleanup(fn)` from its
+  `run` is the imperative escape hatch. Declare `cleanup` explicitly on each scenario (don't
+  club it onto many at once). Don't put teardown on a `verify` step, and no trailing cleanup
+  action step.
+
+Use a dedicated `https.Agent({ keepAlive: false })` for these helper calls so read-back/cleanup
+sockets don't linger as open handles when the suite finishes.
+
+Identities from `ctx.identity` / `ctx.email` are always unique per scenario run
+(`ci-<runId>-…`), so destination 409s on duplicates are avoided without a per-spec strategy.
+
+For eventually-consistent destinations (search-indexed create/update routing), make the
+precondition deterministic in the **setup action** (poll until the record is stably searchable).
+When the transform itself re-searches that index at delivery time, setup polling alone can still
+race — a just-created record is occasionally missed and 409s — so also set `retries` on the
+pipeline step to re-run seed → transform → deliver with backoff. Only use `retries` where a failed
+attempt persists nothing (e.g. a 409-on-create), so repeating is safe; otherwise a re-delivery
+runs against the previous attempt's state.
+
+## Credentials — `LIVE_SECRET_<DEST>` (`LiveSecret` shape)
+
+`SecretResolver` reads one env var, `LIVE_SECRET_<DEST>`, a JSON blob:
+`{ authType, config, secret?, resourceIds?, oauthRefresh?, readback? }`. `config` merges into
+`destination.Config`; `secret` into `metadata.secret`; `resourceIds` supplies account-scoped ids
+(listId, pixelId, measurementId, …); `readback` holds credentials for `verify` steps.
+`resolveConfig(s)` maps `s.config` (plus fixed non-secret defaults taken from the component
+`destination.Config`) into the real `destination.Config`.
+
+**OAuth destinations** (`authType: 'oauth'`): supply `oauthRefresh` (the long-lived refresh token +
+`accountDefinition` + any `providerFields` like Salesforce `instance_url`). Note the harness README
+lists the `rudder-auth` container that services OAuth refresh as **deferred (not in the pilot)** — so
+until that layer lands, an OAuth destination's spec typically goes in with `enabled: false` (parked)
+and is validated for shape only. Prefer piloting `apiKey`/`basic` destinations first.
+
+## Steps
+
+1. **Read** the sources above for `<dest>`; enumerate the distinct behaviors its component cases
+   cover (dedupe error-only ones).
+2. **Choose** `authType`.
+3. **Write `resolveConfig`** from the component `destination.Config` — keep the non-secret fields as
+   fixed defaults, spread `s.config` last for the real credentials.
+4. **For each behavior**, add a scenario: a pipeline step whose `seed(ctx)` reproduces the component
+   `message` with ctx-based identities; a `configOverride` for config-dependent variants; and
+   `setup`/`verify`/`cleanup` steps for stateful ones. For create/update behaviors, factor the
+   asserted fields into a `(ctx) => ({ ... })` profile shared by the seed and a field-level verify
+   so the two stay in lockstep.
+5. **Enroll**: `enabled: true` (registry auto-discovers `destinations/*/live.ts`; no registration).
+   Use `enabled: false` to park it (e.g. OAuth destinations, until rudder-auth lands).
+6. **Verify**:
+   `npx tsc --noEmit -p tsconfig.test.json`
+   then, with real credentials:
+   `LIVE_SECRET_<DEST>='{...}' LOG_LEVEL=silent npm run test:live -- --destination=<dest>`.
+   (Run with `LOG_LEVEL=silent` until secret masking in `network.js` logs is verified — see the
+   README security note.)
+
+## Skeleton — `test/integrations/destinations/<dest>/live.ts`
+
+For a couple of scenarios a single `live.ts` is fine; beyond that, split it into a `live/` module
+folder (`spec.ts` + `api.ts`/`setup.ts`/`verify.ts`/`profiles.ts`) and make `live.ts` a one-line
+re-export — see `destinations/hs/live/`. Steps are plain objects with a required `stepType`
+discriminant (`'pipeline' | 'action' | 'verify'`); there are no `action()`/`verify()` wrapper
+helpers. The `seed` below is an event-stream `identify` (CRM shape) purely to illustrate structure
+— **use the event shape your destination's component `message`s actually take** (e.g. a `track`
+event with `properties` for a conversion, or an audience membership event), swapping literals for
+`ctx` values.
+
+```ts
+import axios from 'axios';
+import { Agent } from 'https';
+import { LiveSpec, RunContext } from '../../live/types';
+import { pollUntil } from '../../live/poll';
+
+// keepAlive:false so read-back/cleanup sockets don't linger as open handles.
+const agent = new Agent({ keepAlive: false });
+
+// Real-API helpers for stateful scenarios (create/read-back/delete). Derive endpoints + auth
+// from src/.../destinations/<dest> and network.ts. Example shapes only:
+// const createEntity = async (ctx: RunContext): Promise<void> => { ... ctx.register(...) };
+// const fetchEntity = async (ctx: RunContext, keys: string[]) => { ... };   // real GET/search
+// const cleanupEntity = async (ctx: RunContext): Promise<void> => { ... };
+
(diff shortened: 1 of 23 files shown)


.github/workflows/live-integration-tests.yml
```diff
@@ -0,0 +1,73 @@
...
+  pull_request:
+    types: ['opened', 'reopened', 'synchronize']
+  push:
+    branches:
+      - main
+      - develop
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.sha }}
+  cancel-in-progress: true
```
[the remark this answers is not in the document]
achettyiitr: For external Pr's we don't run the tests and is already blocked.

.github/workflows/live-integration-tests.yml
```diff
@@ -0,0 +1,97 @@
...
+#
+# Vault (KV v2): each destination's full LiveSecret JSON is stored as a field on the shared path
+#   engineering_shared/data/integrations_team/e2e_test/rudder-transformer
+# named after the env var the resolver expects — e.g. field `LIVE_SECRET_HS`.
+
+name: Live Integration Tests
+
+on:
+  pull_request:
+    types: ['opened', 'reopened', 'synchronize']
```
saikumarrs, line 18: The `pull_request` trigger is unfiltered — every PR, including feature-to-feature ones, spends a live run against the real HubSpot account. Restrict it to the protected base branches.

```suggestion
  pull_request:
    branches:
      - main
      - develop
    types: ['opened', 'reopened', 'synchronize']
```

Next step: confirm whether `release/*` and `hotfix-release/*` bases need it too — the Slack step below already treats those refs as special.

jest.config.live.js
```diff
@@ -0,0 +1,27 @@
...
+  bail: 0,
+  collectCoverage: false,
+  coverageDirectory: 'reports/live-coverage',
+  coverageReporters: ['text', 'text-summary', 'lcov', 'html'],
+  collectCoverageFrom: [
+    'src/**/*.{ts,js}',
+    'test/integrations/live/**/*.ts',
+    'test/integrations/destinations/**/live.ts',
+    '!src/**/*.d.ts',
+  ],
```
saikumarrs: `collectCoverageFrom` includes the harness itself. The base config deliberately excludes it — [`jest.config.js:30`](https://github.com/rudderlabs/rudder-transformer/blob/dfece994d55b20917a4072af0647acb5ddace4c9/jest.config.js#L30) sets `coveragePathIgnorePatterns: [..., 'test']`. Covering `test/integrations/live/**` measures nothing: the suite executes that code by definition, so it reports near-100% and dilutes the `src/` number this run is actually about.

```suggestion
  collectCoverageFrom: ['src/**/*.{ts,js}', '!src/**/*.d.ts'],
```

Next step: drop the two `test/**` globs.

test/integrations/component.live.test.ts
```diff
@@ -0,0 +1,109 @@
...
+import { LiveSecret, EnrolledDestination } from './live/types';
+
+describe('Live Integration Test Suite', () => {
+  // npm run test:live
+  // npm run test:live:coverage
+  // npm run test:live -- --destination=<dest>
+  // npm run test:live:coverage -- --destination=<dest>
+  const command = new Command()
+    .allowUnknownOption()
+    .allowExcessArguments()
```
saikumarrs, line 23: `.allowUnknownOption()` + `.allowExcessArguments()` make a typo silent — `--destinaton=hs` parses clean, `opts.destination` stays `undefined`, and the run quietly covers every enrolled destination instead of failing. They can't just be dropped, since jest's own argv flows through this parser.

Next step: keep the permissive parse and add the visibility back after it.

```ts
const opts = command.opts();
if (command.args.length) {
  console.warn(`[live] ignoring unrecognised args: ${command.args.join(' ')}`);
}
```

Pair it with a one-line log of the destination list resolved on line 48, so a typo surfaces as "running all N destinations" rather than passing unnoticed.

.github/workflows/live-integration-tests.yml
```diff
@@ -0,0 +1,97 @@
...
+      - main
+      - develop
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  id-token: write # required: lets the job mint the GitHub OIDC JWT
+  contents: read # for actions/checkout
```
saikumarrs, line 29: **nit:** keep the top-level block to the least privilege every job needs and let jobs opt into more — only the `live` job mints the OIDC JWT. `ut-tests.yml` and `verify.yml` already scope per job this way.

```suggestion
permissions:
  contents: read # for actions/checkout
```

Next step: add `permissions: { id-token: write, contents: read }` under `jobs.live`.

saikumarrs (commented):
Framework looks solid — the `pipeline`/`action`/`verify` step model with scenario-level `cleanup` is the right shape, and porting the transform-to-deliver chaining into `routerProxyRequests.ts` keeps the runner honest about what rudder-server actually does.

Nothing here blocks. Three themes:

**CI cost and scope** — the workflow runs the full matrix on every PR against a real HubSpot account, on a 30-minute self-hosted runner. Gating on changed destinations and protected base branches is the bulk of the comments below.

**Gaps worth marking, not necessarily fixing** — GZIP payloads, the processor-transform path, and the weak `toBeGreaterThan(0)` output assertions. TODOs with tickets are fine for the first two.

**The skill doc is the one I'd most like extended before this lands** — it generates `live.ts` and stops, but onboarding a destination also needs a Vault field, a workflow matrix entry, and a local credential blob whose shape the developer currently can't guess.

Details inline.

.github/workflows/live-integration-tests.yml
```diff
@@ -0,0 +1,97 @@
...
+    name: live (${{ matrix.destination }})
+    # required: puts the job in infra1-us → PrivateLink path to dev Vault
+    runs-on: [self-hosted, Linux, X64]
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - destination: hs
+            secretName: LIVE_SECRET_HS
```
saikumarrs: `secretName` is derivable — [`secretResolver.ts`](https://github.com/rudderlabs/rudder-transformer/blob/dfece994d55b20917a4072af0647acb5ddace4c9/test/integrations/live/secretResolver.ts) already contracts on `LIVE_SECRET_<DEST>`, and the header comment documents the Vault field by that same name. Carrying it in the matrix gives a second place to get it wrong on every new destination.

The snag is GHA has no `upper()`. Two ways out:

1. Derive it in a step before the Vault import, and collapse the matrix to `destination: [hs]`:
```yaml
      - id: secret
        run: echo "name=LIVE_SECRET_$(echo '${{ matrix.destination }}' | tr 'a-z' 'A-Z')" >> "$GITHUB_OUTPUT"
```
then reference `${{ steps.secret.outputs.name }}` in the Vault step.
2. Leave it, and add a comment tying the field to the resolver's convention so the coupling is at least documented.

Leaning toward (1) — it keeps enrollment to one line per destination.

.github/workflows/live-integration-tests.yml
```diff
@@ -0,0 +1,97 @@
...
+  live:
+    name: live (${{ matrix.destination }})
+    # required: puts the job in infra1-us → PrivateLink path to dev Vault
+    runs-on: [self-hosted, Linux, X64]
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - destination: hs
```
saikumarrs: Every PR runs the full live matrix regardless of what it touched — a `ga4` change spends a real HubSpot run and 30 minutes of a self-hosted runner. Gate the matrix on the PR's changed paths.

The primitive is already in the repo: [`verify.yml:46`](https://github.com/rudderlabs/rudder-transformer/blob/dfece994d55b20917a4072af0647acb5ddace4c9/.github/workflows/verify.yml#L46) uses `Ana06/get-changed-files` with `pull-requests: read`.

Next step: add a `detect` job that maps changed paths under `src/**/destinations/<dest>/` and `test/integrations/destinations/<dest>/` to a destination list, emits it as a JSON output, and feed it in:

```yaml
    needs: detect
    strategy:
      matrix:
        destination: ${{ fromJSON(needs.detect.outputs.destinations) }}
```

Keep the full matrix on `push` to `main`/`develop` — the regression sweep is worth it there.

.github/workflows/live-integration-tests.yml
```diff
@@ -0,0 +1,97 @@
...
+
+      - name: Import destination secret from dev Vault (OIDC/JWT)
+        id: vault
+        uses: hashicorp/vault-action@892a26828f195e65540a40b4768ae4571f51ebfc # v4.0.0
+        with:
+          url: https://vault.dev.rudderlabs.com
+          method: jwt
+          path: github-actions # = vault_jwt_auth_backend.path
+          role: rudder-transformer-ci # = vault_jwt_auth_backend_role.role_name
+          jwtGithubAudience: https://vault.dev.rudderlabs.com # MUST match bound_audiences, else auth fails
```
saikumarrs: `url`, `path`, `role` and `jwtGithubAudience` are environment-scoped infra values, not workflow logic — a Vault move or a role rename becomes a code change in every repo that authenticates this way.

```suggestion
          url: ${{ vars.VAULT_ADDR }}
          method: jwt
          path: ${{ vars.VAULT_JWT_AUTH_PATH }} # = vault_jwt_auth_backend.path
          role: ${{ vars.VAULT_JWT_ROLE }} # = vault_jwt_auth_backend_role.role_name
          jwtGithubAudience: ${{ vars.VAULT_ADDR }} # MUST match bound_audiences, else auth fails
```

Reusing `VAULT_ADDR` for both keeps the `url == jwtGithubAudience` invariant impossible to break by editing one of them.

Next step: create the three repo-level Actions vars, then swap the literals.

test/integrations/live/runPipelineStep.ts
```diff
@@ -0,0 +1,111 @@
...
+  const routerBody = buildRouterTransformBody(destination, message, config, jobId, {
+    secret: ctx.liveSecret.secret,
+    metadataOverride: step.metadataOverride,
+  });
+
+  const routerResponse = await http.post('/routerTransform', routerBody);
+  expect(routerResponse.status).toEqual(HttpStatusCode.Ok);
+
+  const routerOutputs = parseSuccessfulRouterOutputs(routerResponse.body);
+  expect(routerOutputs.length).toBeGreaterThan(0);
```
saikumarrs: `toBeGreaterThan(0)` accepts any non-zero count, so a batching regression that collapses two events into one output — or fans one into three — still passes. The step knows what it seeded; the expected count should come from the spec rather than be waved through.

Next step: add an optional `expectedOutputs?: number` to `PipelineStep` ([`types.ts:46`](https://github.com/rudderlabs/rudder-transformer/blob/dfece994d55b20917a4072af0647acb5ddace4c9/test/integrations/live/types.ts#L46)) and assert it when set:

```ts
if (step.expectedOutputs !== undefined) {
  expect(routerOutputs).toHaveLength(step.expectedOutputs);
} else {
  expect(routerOutputs.length).toBeGreaterThan(0);
}
```

Same argument applies to `proxyRequests.length` two lines down — that's where a `dontBatch` regression would slip through silently.

test/integrations/live/runPipelineStep.ts
```diff
@@ -0,0 +1,111 @@
...
+        const redactedJobStates = jobStates.map((js) => ({
+          ...js,
+          metadata: { ...js.metadata, secret: '[redacted]' },
+        }));
+        return (
+          `[live:${destination}:${scenarioId}:${step.name}] not delivered — ` +
+          `proxy HTTP ${deliveryResponse.status}, verdict.status ${deliveryOutput.status}, ` +
+          `message: ${deliveryOutput.message}\n` +
+          `job states: ${JSON.stringify(redactedJobStates, null, 2)}`
+        );
```
saikumarrs: Returning a pre-formatted string forces every consumer through string parsing. The retry loop above can't tell a 409-on-create (safe to repeat) from a hard auth failure (isn't), so it backs off and retries either way — and the final `throw new Error(failure)` is the only place that actually needs prose.

Next step: return the structured failure and format at the throw site.

```ts
type DeliveryFailure = {
  proxyStatus: number;
  verdictStatus: number;
  message: string;
  jobStates: unknown[];
};
```

`attemptDelivery` returns `DeliveryFailure | undefined`; `runPipelineStep` renders the `[live:<dest>:<scenario>:<step>]` message it throws. Redaction stays exactly where it is.

test/integrations/live/routerProxyRequests.ts
```diff
@@ -0,0 +1,155 @@
...
+    endpointPath: batchedRequest.endpointPath,
+    userId: batchedRequest.userId || '',
+    headers: batchedRequest.headers || {},
+    params: batchedRequest.params || {},
+    body: {
+      JSON: batchedRequest.body?.JSON || {},
+      JSON_ARRAY: batchedRequest.body?.JSON_ARRAY || {},
+      XML: batchedRequest.body?.XML || {},
+      FORM: batchedRequest.body?.FORM || {},
+      GZIP: {},
```
saikumarrs: `GZIP` isn't in `LiveProcessorOutputSchema` above, so zod strips it, and this hardcodes `{}` — a destination that emits a gzip body would deliver an empty one and could still come back 2xx. That's a false pass, not a failure. Fine to defer, but it needs a marker so the next reader doesn't take the empty object for a deliberate no-op.

```suggestion
      // TODO INT-NNNN: GZIP payloads aren't mapped — the schema above strips them, so
      // destinations that emit gzip bodies can't be covered live yet.
      GZIP: {},
```

Next step: file the ticket and drop the number in.

test/integrations/live/runPipelineStep.ts
```diff
@@ -0,0 +1,111 @@
...
+  destination,
+  scenarioId,
+  step,
+  ctx,
+  config,
+  http,
+}: RunPipelineStepParams): Promise<string | undefined> => {
+  const jobId = randomInt(1, 2_147_483_647);
+  const message = step.seed(ctx);
+  const routerBody = buildRouterTransformBody(destination, message, config, jobId, {
```
saikumarrs, line 35: The harness only drives `/routerTransform` -> `/proxy`. rudder-server also runs the processor transform (`/v0/destinations/<dest>`) ahead of delivery, and that response shape differs — so nothing here exercises that chaining, and nothing in the code or README says so.

Next step: pick one —
1. Add a `transformEndpoint?: 'router' | 'processor'` knob on `PipelineStep` and branch here.
2. Mark the gap: a `// TODO INT-NNNN` on this call plus a scope line in `test/integrations/live/README.md`.

Leaning toward (2) — (1) is more than this PR should carry.

test/integrations/live/routerProxyRequests.ts
```diff
@@ -0,0 +1,155 @@
...
+  return batched.map((br) => buildProxyRequest(br, item.metadata, item.destination));
+};
+
+export type BuildRouterTransformBodyOptions = {
+  secret?: Record<string, string>;
+  metadataOverride?: Record<string, unknown>;
+};
+
+// Builds a /routerTransform request body for a single live event (one input job).
+export const buildRouterTransformBody = (
```
saikumarrs: **nit:** this builds a transform *request*; everything above maps a transform *response* to proxy requests. The header comment on line 1 already has to describe both halves — that's the tell.

Next step: move `buildRouterTransformBody` and `BuildRouterTransformBodyOptions` to `live/routerTransformRequest.ts` and leave this file to the proxy mapping. `readString` is the only helper both need — lift `isRecord`/`readString`/`readNumber`/`readBoolean` into a small `live/coerce.ts`.

test/integrations/live/registry.ts
```diff
@@ -0,0 +1,54 @@
...
+const DESTINATIONS_DIR = join(__dirname, '..', 'destinations');
+
+// Dynamically require a spec module (discovered at runtime, not statically imported). A spec may be
+// exported as `live` or as the default export. Returns undefined if missing or not enabled.
+function loadSpec(destination: string): LiveSpec | undefined {
+  const specPathTs = join(DESTINATIONS_DIR, destination, 'live.ts');
+  const specPathJs = join(DESTINATIONS_DIR, destination, 'live.js');
+  if (!existsSync(specPathTs) && !existsSync(specPathJs)) {
+    return undefined;
+  }
```
saikumarrs: Drop the `.js` branch — every live spec is TypeScript, the skill and README document only `live.ts`, and ts-jest is what loads them. Worse, `require(.../'live')` resolves `.js` ahead of `.ts`, so a stale compiled `live.js` would silently shadow the spec someone is editing.

```suggestion
  const specPath = join(DESTINATIONS_DIR, destination, 'live.ts');
  if (!existsSync(specPath)) {
    return undefined;
  }
```

Next step: also fix the header comment on line 3 (`live.ts|js`) in the same pass.

test/integrations/destinations/hs/live/spec.ts
```diff
@@ -0,0 +1,353 @@
...
+          stepType: 'pipeline',
+          seed: (ctx) => ({
+            ...baseTimestamps(ctx, 'es-contacts-create'),
+            type: 'identify',
+            context: esLibrary,
+            traits: { email: ctx.email(), ...esContactCreateTraits(ctx) },
+            integrations: { All: true },
+          }),
+        },
+        verifyContactProperties(esContactCreateTraits),
```
saikumarrs: Verify should be declarable on the scenario the way `cleanup` is, with the framework owning the polling. Today every destination hand-rolls it: [`hs/live/verify.ts`](https://github.com/rudderlabs/rudder-transformer/blob/dfece994d55b20917a4072af0647acb5ddace4c9/test/integrations/destinations/hs/live/verify.ts) repeats the same `{ attempts: 4, delayMs: (n) => 1000 * 2 ** n, soft: true }` block in both helpers, and the next destination copies it a third time.

Next step: add an optional `verify` alongside `cleanup` on `LiveScenario` ([`types.ts:75`](https://github.com/rudderlabs/rudder-transformer/blob/dfece994d55b20917a4072af0647acb5ddace4c9/test/integrations/live/types.ts#L75)):

```ts
verify?: {
  check: (ctx: RunContext) => Promise<void>;
  attempts?: number;   // default 4
  delayMs?: (attempt: number) => number;  // default 1000 * 2 ** attempt
};
```

The runner wraps `check` in `pollUntil` and rethrows the last matcher error on exhaustion — same diff quality `soft: true` gives today, and destination code shrinks to the assertion. Keep `VerifyStep` for the ordering cases (verify-after-setup, mid-scenario); this is about the common trailing one.

.claude/skills/live-integration-test/SKILL.md
```diff
@@ -0,0 +1,260 @@
...
+   `message` with ctx-based identities; a `configOverride` for config-dependent variants; and
+   `setup`/`verify`/`cleanup` steps for stateful ones. For create/update behaviors, factor the
+   asserted fields into a `(ctx) => ({ ... })` profile shared by the seed and a field-level verify
+   so the two stay in lockstep.
+5. **Enroll**: `enabled: true` (registry auto-discovers `destinations/*/live.ts`; no registration).
+   Use `enabled: false` to park it (e.g. OAuth destinations, until rudder-auth lands).
+6. **Verify**:
+   `npx tsc --noEmit -p tsconfig.test.json`
+   then, with real credentials:
+   `LIVE_SECRET_<DEST>='{...}' LOG_LEVEL=silent npm run test:live -- --destination=<dest>`.
```
saikumarrs, line 166: This step hands the developer `LIVE_SECRET_<DEST>='{...}'` and never shows what goes in the braces — which is the one thing they can't derive. More broadly, the skill stops at generating `live.ts`, but enrolling a destination is a multi-system change. Four additions would close it:

1. **Emit a filled credential template.** By the time `resolveConfig` is written the skill knows the destination's config fields, so it should write a `.env`-ready skeleton with placeholders for the developer to fill:
```
LIVE_SECRET_HS={"authType":"apiKey","config":{"accessToken":"<private-app-token>"},"readback":{"accessToken":"<private-app-token>"}}
```
Without it, local validation is guesswork.

2. **Gate on OAuth in the Steps flow.** The Credentials section says `rudder-auth` is deferred, but step 2 ("Choose `authType`") doesn't branch on it. Make it a stop: if the destination is OAuth, park the spec with `enabled: false` and say so, rather than generating something that can't run.

3. **Cover the workflow change.** Enrollment needs a matrix entry in `.github/workflows/live-integration-tests.yml`. The skill never names that file.

4. **Print an onboarding checklist at the end** for the work that isn't in the repo:
   - Vault: add the `LIVE_SECRET_<DEST>` field under `engineering_shared/data/integrations_team/e2e_test/rudder-transformer`
   - GitHub: any Actions vars/secrets the destination needs
   - the sandbox account: who owns it, what the run creates, what cleanup is expected to remove

Next step: fold (1)-(3) into `## Steps` and add (4) as a closing `## After the spec — onboarding checklist` section.

.github/workflows/live-integration-tests.yml
achettyiitr: > Every PR runs the full live matrix regardless of what it touched — a ga4 change spends a real HubSpot run and 30 minutes of a self-hosted runner. Gate the matrix on the PR's changed paths.

Is it the billing that we are concerned about? Each of these integration tests takes a few minutes to complete. Based on [this](https://rudderstack.grafana.net/d/ae4da686-e16d-472e-9c50-75168236dfe8/actions-runner-controller?orgId=1&from=now-30m&to=now&timezone=browser&var-datasource=c7f0cc2b-c87e-407c-85b2-33d8d0b14eb8&var-runnerdeployment=$__all), we can re-evaluate on it later.

.github/workflows/live-integration-tests.yml
achettyiitr: No we don’t need to add `release`/, `hotfix-release` to the pull_request.branches filter. That filter is the base (target) branch.

.github/workflows/live-integration-tests.yml
saikumarrs: Yes, cost is part of the equation; maybe negligible at this point.
When we make this a required check, an unrelated failing test should not block a PR.

.github/workflows/live-integration-tests.yml
saikumarrs: What about main and develop branches? My main suggestion was around those branches.

.github/workflows/live-integration-tests.yml
achettyiitr: Currently we are planning to keep it non required only.

.github/workflows/live-integration-tests.yml
achettyiitr: Yeah, added only main and develop branches.

test/integrations/live/routerProxyRequests.ts
achettyiitr: Added in LiveProcessorOutputSchema.

test/integrations/component.live.test.ts
achettyiitr:
```
➜  rudder-transformer git:(feat.live-integration-tests) ✗ npm run test:live:coverage -- --destination=hs

> rudder-transformer@1.142.0 test:live:coverage
> npm run test:live -- --coverage --destination=hs


> rudder-transformer@1.142.0 test:live
> NODE_OPTIONS='--no-node-snapshot' jest -c jest.config.live.js --coverage --destination=hs

  console.warn
    [live] ignoring unrecognised args: -c jest.config.live.js --coverage

      27 |   if (command.args.length) {
      28 |     // eslint-disable-next-line no-console
    > 29 |     console.warn(`[live] ignoring unrecognised args: ${command.args.join(' ')}`);
         |             ^
      30 |   }
      31 |
      32 |   let server: Server;

      at test/integrations/component.live.test.ts:29:13
      at Object.<anonymous> (test/integrations/component.live.test.ts:16:1)

  console.log
    [live] resolved 1 destination(s): hs
```

Even with valid syntax it's throwing it.

achettyiitr: Currently we are logging `[live] resolved 1 destination(s): hs`, which should be sufficient.

.github/workflows/live-integration-tests.yml
```diff
@@ -57,18 +60,26 @@ jobs:
...
         run: npm ci && npm run prepare
 
+      - name: Resolve secret name
+        id: secret
+        run: echo "name=LIVE_SECRET_$(echo '${{ matrix.destination }}' | tr 'a-z' 'A-Z')" >> "$GITHUB_OUTPUT"
+
       - name: Import destination secret from dev Vault (OIDC/JWT)
         id: vault
         uses: hashicorp/vault-action@892a26828f195e65540a40b4768ae4571f51ebfc # v4.0.0
+        env:
```
saikumarrs: I earlier meant to add these as "Variables" in GitHub Actions settings as it'll be easy to change them without have to update the code.

achettyiitr: My bad. I thought you mean env variable. Made changes now.

saikumarrs (dismissed):

saikumarrs (approved):

PR #5374, opened 2026-07-19
fix(deps): bump jsonata from 2.0.6 to 2.2.1 (security) [Snyk]

achettyiitr:
snyk-top-banner

#### Snyk changed the following file(s):

- `package.json`
- `package-lock.json`
- Resolves INT-6841.

#### Vulnerabilities that will be fixed with an upgrade:

|  | Issue | Score | 
:-------------------------:|:-------------------------|:-------------------------
medium severity | Prototype Pollution <br/>[SNYK-JS-JSONATA-17339139](https://snyk.io/vuln/SNYK-JS-JSONATA-17339139) | &nbsp;&nbsp;**666**&nbsp;&nbsp; 

#### Breaking Change Risk

Merge Risk: Medium

> **Notice:** This assessment is enhanced by AI.

---

> [!IMPORTANT]
>
> - Check the changes in this PR to ensure they won't cause issues with your project.
> - Max score is 1000. Note that the real score may have changed since the PR was raised.
> - This PR was automatically created by Snyk using the credentials of a real user.

---

**Note:** _You are seeing this because you or someone else with access to this repository has authorized Snyk to open fix PRs._

For more information: 
🧐 [View latest project report](https://app.snyk.io/org/rudder-integrations/project/ac49de83-1b33-4830-9edc-231312430c7a?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr)
📜 [Customise PR templates](https://docs.snyk.io/scan-using-snyk/pull-requests/snyk-fix-pull-or-merge-requests/customize-pr-templates?utm_source=github&utm_content=fix-pr-template)
🛠 [Adjust project settings](https://app.snyk.io/org/rudder-integrations/project/ac49de83-1b33-4830-9edc-231312430c7a?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;fix-pr/settings)
📚 [Read about Snyk's upgrade logic](https://docs.snyk.io/scan-with-snyk/snyk-open-source/manage-vulnerabilities/upgrade-package-versions-to-fix-vulnerabilities?utm_source=github&utm_content=fix-pr-template)

---

**Learn how to fix vulnerabilities with free interactive lessons:**

🦉 [Prototype Pollution](https://learn.snyk.io/lesson/prototype-pollution/?source&#x3D;registry-oss-fix-pr&amp;loc&#x3D;fix-pr)

[...]

PR #5388, opened 2026-07-23
chore: validate full PR commit range and forward $1 to commit-msg hook

maheshkutty:
🔒 Scanned for secrets using gitleaks 8.30.1

- CI: lint --last on push and the full base..head range on pull_request events (per commitlint's CI guide); previously only linted HEAD~1..HEAD
- husky: forward ${1} (the commit-message file path) to the commit-msg script so the actual message is validated

Resolves INT-XXX

N/A

N/A

N/A

N/A

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist

### DIFF
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -1,6 +1,6 @@
 name: Commitlint
 
-on: [push]
+on: [push, pull_request]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.sha }}
@@ -41,6 +41,10 @@ jobs:
           npm --version
           npx commitlint --version
 
-      - name: Lint commits
-        run: |
-          npx --no-install commitlint --from=HEAD~1 --to=HEAD --verbose
+      - name: Validate current commit (last commit) with commitlint
+        if: github.event_name == 'push'
+        run: npx --no-install commitlint --last --verbose
+
+      - name: Validate PR commits with commitlint
+        if: github.event_name == 'pull_request'
+        run: npx --no-install commitlint --from ${{ github.event.pull_request.base.sha }} --to ${{ github.event.pull_request.head.sha }} --verbose
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -12,6 +12,10 @@ permissions:
 jobs:
   commitlint:
     runs-on: ubuntu-latest
+    # Release branches are assembled by automation from commits already linted on
+    # their way into develop. Re-linting the accumulated range adds no coverage and
+    # fails on historical bot commits whose footers exceed footer-max-line-length.
+    if: ${{ !startsWith(github.head_ref, 'release/') && !startsWith(github.head_ref, 'hotfix-release/') }}
     permissions:
       contents: read # to checkout repository code with full history (actions/checkout)
     steps:
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -1,9 +1,9 @@
 name: Commitlint
 
-on: [push, pull_request]
+on: [pull_request]
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref || github.sha }}
+  group: ${{ github.workflow }}-${{ github.head_ref }}
   cancel-in-progress: true
 
 permissions:
@@ -41,10 +41,5 @@ jobs:
           npm --version
           npx commitlint --version
 
-      - name: Validate current commit (last commit) with commitlint
-        if: github.event_name == 'push'
-        run: npx --no-install commitlint --last --verbose
-
       - name: Validate PR commits with commitlint
-        if: github.event_name == 'pull_request'
         run: npx --no-install commitlint --from ${{ github.event.pull_request.base.sha }} --to ${{ github.event.pull_request.head.sha }} --verbose
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,2 +1,2 @@
 #!/bin/sh
-npm run commit-msg ${1}
+npm run commit-msg -- "$1"
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,2 +1,2 @@
 #!/bin/sh
-npm run commit-msg
+npm run commit-msg ${1}


.github/workflows/commitlint.yml
```diff
@@ -1,6 +1,6 @@
 name: Commitlint
 
-on: [push]
+on: [push, pull_request]
```
shekhar-rudder: Since `develop`/`main` are protected and disallow direct pushes, the `push` trigger's only unique coverage (direct commits to protected branches) can never happen here — every change arrives via a PR. That leaves `push` doing shallow, redundant work: it only lints `--last` (a single commit), and pushing to a PR branch now fires **both** `push` and `pull_request`, whose concurrency groups differ (`github.sha` vs `github.head_ref`), so neither cancels the other and commitlint runs twice per push.

Suggest making this `pull_request`-only:
```yaml
on: [pull_request]
```
The `push` step below can then be dropped too. (The local `.husky/commit-msg` hook already gives contributors feedback at commit time before a PR exists.)

.husky/commit-msg
```diff
@@ -1,2 +1,2 @@
 #!/bin/sh
-npm run commit-msg
+npm run commit-msg ${1}
```
shekhar-rudder: `npm run commit-msg ${1}` doesn't actually forward `$1`. npm only passes positional args to a script when they follow a `--` separator; without it the argument is dropped, so this runs plain `commitlint --edit` (per the `commit-msg` script in `package.json`) and falls back to its default `./.git/COMMIT_EDITMSG` lookup. In the normal case that path coincides with `$1` so validation still happens by accident — but in exactly the case where forwarding matters (git worktrees, where `.git` is a file and the default lookup fails while `$1` holds the correct path) it still won't work, so the change is effectively a no-op.

Fix (also quote for paths with spaces):
```sh
npm run commit-msg -- "$1"
```

maheshkutty: Applied — thanks. One correction on the reasoning though: npm has forwarded bare positional args without `--` since npm 7, so `npm run commit-msg ${1}` did pass `$1` through. Verified on npm 10.9.2:

```
npm run show /path/to/MSG      → ARGV: ["/path/to/MSG"]
npm run show -- "/path/to/MSG" → ARGV: ["/path/to/MSG"]
```

Identical, so it wasn't a no-op falling back to the default lookup.

The real bug is the missing quotes: unquoted `${1}` word-splits on paths containing spaces. Your worktree point is valid and confirmed — in a worktree `.git` is a regular file (`gitdir: …`), not a directory, so the default `./.git/COMMIT_EDITMSG` lookup fails and the forwarded path is what makes it work.

Keeping `--` for explicitness (guards against a path starting with `-` being parsed as an npm flag). Fix is in.

.github/workflows/commitlint.yml
maheshkutty: Agreed, applied in 9b7df54 — `on: [pull_request]`, dropped the `--last` step and its `if:` guard, and simplified the concurrency group to `${{ github.workflow }}-${{ github.head_ref }}` (the `|| github.sha` fallback only existed to cover the empty `head_ref` on `push` events).

Verified the premises before applying:

- **Protection is in place** — org ruleset `master-branch-protection` applies a `pull_request` rule to `develop`/`main` requiring 1 approval, plus `deletion` and `non_fast_forward`. Direct pushes really can't happen.
- **The duplicate-run analysis is right** — `head_ref` is empty on `push`, so the two runs grouped under `Commitlint-<sha>` and `Commitlint-<branch>`. Different groups, neither cancelling the other.

Worth noting the change *strengthens* the skipped-hook case rather than weakening it, which was my initial concern: the `push` path only ran `commitlint --last` (one commit), so `--no-verify` on 5 commits let 4 through. The `pull_request` path validates the whole `base.sha..head.sha` range and catches all of them. `synchronize` still fires on every push to an open PR, so per-push coverage is unchanged.

One gap I'll flag rather than block on: `push` had unique coverage for merge commits landing on `develop`/`main` (e.g. the `chore(release): pull main into develop …` merges), whose subjects aren't in the PR's `base..head` range. Not worth keeping `push` for — those are bot-generated with consistent subjects, and a post-hoc failure on a protected branch isn't actionable anyway. If we want that covered, a PR-title lint on `pull_request_target` is the right tool; happy to add it separately.

shekhar-rudder:
`--from ${{ github.event.pull_request.base.sha }}` in `.github/workflows/commitlint.yml` makes every release PR fail.

For `release/vX → main` PRs, `base.sha` *is* the merge-base (verified for #5426 and #5413), so the linted range is the whole accumulated release — 47 and 40 commits respectively — not the PR author's own work.

I ran commitlint 19 with this repo's `commitlint.config.js` over both ranges. Two commits fail in each:

- `a9beb357` — `fix(deps): bump jsonata from 2.0.6 to 2.2.1 (security) [Snyk] (#5374)`
- `0d52ebc1` — `chore(deps): bump ryancyq/github-signed-commit (#5353)`

Both on `footer-max-line-length`: config-conventional's default is 100, and `commitlint.config.js` overrides only `body-max-line-length` (to 500), not the footer rule. These are squash commits whose bodies are bot PR descriptions, so they can't be fixed retroactively. Under the previous `HEAD~1..HEAD` on push they were never linted.

This won't block the merge — `main`'s ruleset requires only `Check PR Title`, `SonarCloud Code Analysis` and `codecov/project`, not `commitlint` — but it turns the check permanently red on every release, which is how a check that *is* required on `develop` loses its credibility.

Ordinary and bot PRs into `develop` are unaffected: I linted the dependabot branch commit from #5402 and the Snyk branch commit from #5374, and both pass.

maheshkutty:
Reproduced all of this locally — the finding holds.

**Verified:**

- `base.sha` *is* the merge-base for both release PRs, so the linted range is the whole release: `git merge-base` returns exactly `base.sha` for #5426 (47 commits) and #5413 (40 commits).
- Both ranges fail — `commitlint --from <base.sha> --to <head.sha>` exits 1 for each.
- Both named commits fail individually on `footer-max-line-length`: `a9beb357` and `0d52ebc1`. Cause is as described — `commitlint.config.js` overrides only `body-max-line-length` (500) and `type-enum`, so config-conventional's default footer limit of 100 still applies. The Snyk commit's footer carries a long `snyk:metadata:` line and tracking-pixel URLs; unfixable retroactively.

**One correction:** `main`'s required checks are `Check PR Title`, `Update Helm Charts For Staging and Create Pull Request`, `SonarCloud Code Analysis` and `codecov/project` — the Helm one is missing from your list. Doesn't change the conclusion (`commitlint` still isn't required on `main`, so it won't block).

**Also worth noting:** `develop` has *no* required status checks configured — its ruleset has an empty `required_status_checks`. So commitlint isn't currently blocking there either; the credibility argument still stands, but the check is advisory on both branches today.

On the fix — I'd rather not raise `footer-max-line-length` globally, since that disables a real rule for all future commits to paper over two historical bot commits. Options I see:

1. **`ignores` in `commitlint.config.js`** — skip commits whose body matches the Snyk/dependabot metadata signature. Targeted, keeps the rule live for humans.
2. **Skip the job on release PRs** — `if: !startsWith(github.head_ref, 'release/')`. Release branches are bot-assembled from already-merged commits that were linted on the way into `develop`, so re-linting them adds nothing.

Leaning towards 2 — it's the smaller change and matches where the linting value actually is. Happy to go with 1 if you'd rather keep coverage on release PRs. Will follow up in this PR either way rather than leave the check red.

shekhar-rudder:
> Leaning towards 2

LGTM

shekhar-rudder (approved):

PR #5399, opened 2026-07-26
chore: split rETL from event-stream in router transform for hubspot

achettyiitr:

Add a dedicated reverse-ETL (rETL) code path for the HubSpot destination, decoupled from the event-stream transform logic and gated per workspace so it can be rolled out safely.

Behaviour
- New modules under src/v0/destinations/hs/:
  - retl-transform.ts: rETL router entry (processBatchRouterRetl) plus the DEST_HS_RETL_SPLIT_WORKSPACE_IDS gate helpers.
  - retl-hs-transform-v3.ts: rETL identify (objects + associations) and batching, new/v3 API.
  - retl-hs-transform-v1.ts: rETL identify and batching, legacy API.
- The gate lives at the top of processBatchRouter in transform.ts: a router batch uses the rETL path only when its workspaceId is in DEST_HS_RETL_SPLIT_WORKSPACE_IDS AND its events are mappedToDestination (rETL). Everything else - event-stream events and every workspace not in the allow-list - keeps using the existing logic untouched. With the env var unset (the default), behaviour is byte-identical to before.
- The rETL path is self-contained: it only handles rETL identify (object create/update and associations) and rejects anything else instead of falling back to the event-stream handlers. The invalid-email validation from the v3 event-stream identify is preserved on the rETL path.
- Resolves INT-6781.


ItsSudip (approved):

src/v0/destinations/hs/transform.ts
```diff
@@ -65,6 +66,15 @@ const processBatchRouter = async (
   inputs: HubspotRouterRequest[],
   reqMetadata: NonNullable<unknown>,
 ): Promise<HubSpotBatchRouterResult> => {
+  // Workspace-gated rETL/event-stream split. A router call is homogeneous per
+  // workspace/source, so allow-listed workspaces' rETL (mappedToDestination)
+  // batches are handled by the dedicated rETL code path. Everything else - and
+  // every workspace not in the allow-list - keeps using the existing logic
+  // below, unchanged.
+  if (inputs.length > 0 && shouldUseHsRetlSplitPath(inputs[0])) {
```
krishna2020, line 74: Gating on `inputs[0]` is safe here: the webapp does not allow event-stream and rETL sources to be connected to the same destination, so a router batch is never a mix of the two.

PR #5400, opened 2026-07-26
fix(hubspot): populate endpointPath on delivery requests and normalize destType casing

achettyiitr:
## What & why

Two problems with `transformer_outgoing_request_count` for HubSpot:

1. **Every endpoint collapsed into `endpointPath="default"`.** HubSpot delivery requests never set the `endpointPath` field, so rudder-server substituted the literal `"default"` (`router/worker.go`) for all of them. Upsert, create, update, search, and track were indistinguishable on dashboards, and per-endpoint delivery throttling couldn't discriminate either.
2. **`destType` split across `hs` and `HS`.** rudder-server emits `destType` from `DestinationDefinition.Name` (`HS`), while the transformer's own HubSpot API calls hard-coded the lowercase literal `'hs'` — so the same destination showed up under two labels.
3. Resolves INT-6784.

## Changes

- **Populate `endpointPath` on each delivered request**, built the same way as `endpoint`: the processor sets the single-object path from a `_PATH` constant co-located with each endpoint constant in `config.ts`, and the batch step appends the `/batch/*` suffix (or, for fixed endpoints like contacts/track, sets the sibling `_PATH` constant directly). Object/association types are templated via a shared `:objectType` placeholder, and record IDs (`hsSearchId`, `contactId`) never enter the tag, so label cardinality stays bounded. Dashboards now get real, per-endpoint labels, e.g. `/crm/v3/objects/contacts/batch/upsert`, `/crm/v3/objects/lead/batch/create`, `/crm/v3/associations/:fromObjectType/:toObjectType/batch/create`, `/events/v3/send`.
- ** Normalise `destType`** from `'hs'` to the `DESTINATION` constant (`'HS'`) at the five `util.ts` stat-tag call sites, matching what rudder-server emits (`DestinationDefinition.Name`) and the cache tags.
- **Update the hs router component fixtures** with the new `endpointPath` values.

### DIFF
--- a/src/v0/destinations/hs/config.ts
+++ b/src/v0/destinations/hs/config.ts
@@ -1,6 +1,8 @@
 import { getMappingConfig } from '../../util';
 
 const BASE_ENDPOINT = 'https://api.hubapi.com';
+// Placeholder replaced with the real object type when building object endpoints/paths.
+const OBJECT_TYPE_PLACEHOLDER = ':objectType';
 
 // For fetching properties from HubSpot
 const CONTACT_PROPERTY_MAP_ENDPOINT = `${BASE_ENDPOINT}/properties/v1/contacts/properties`;
@@ -19,11 +21,13 @@ const IDENTIFY_CREATE_UPDATE_CONTACT = `${BASE_ENDPOINT}/contacts/v1/contact/cre
 // Identify Batch
 // Ref - https://legacydocs.hubspot.com/docs/methods/contacts/batch_create_or_update
 const BATCH_CONTACT_ENDPOINT = `${BASE_ENDPOINT}/contacts/v1/contact/batch/`;
+const BATCH_CONTACT_ENDPOINT_PATH = '/contacts/v1/contact/batch/';
 const MAX_BATCH_SIZE = 1000;
 
 // Track
 // Ref - https://legacydocs.hubspot.com/docs/methods/enterprise_events/http_api
 const TRACK_ENDPOINT = 'https://track.hubspot.com/v1/event';
+const TRACK_ENDPOINT_PATH = '/v1/event';
 
 /*
  * NEW API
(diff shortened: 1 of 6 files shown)


src/v0/destinations/hs/config.ts
saikumarrs, line 1: Now that we have created new constants for endpoint paths, we can refactor the constants and to re-use them everywhere.
We can delete endpoint constants and compute them on the fly with base endpoint and (placeholder value replaced) endpoint path.

```
const CRM_CREATE_UPDATE_ALL_OBJECTS_ENDPOINT_PATH = '...';

// ... somewhere in the code
endPointPath = CRM_CREATE_UPDATE_ALL_OBJECTS_ENDPOINT.replace(replace object type here);
endpoint = `${BASE_ENDPOINT}${endPointPath}`;

```

saikumarrs (changes requested):

src/v0/destinations/hs/config.ts
achettyiitr: Good point. Let me make the changes.

src/v0/destinations/hs/es-retl-v3.ts
```diff
@@ -236,19 +247,23 @@ const processIdentify = async (
     if (contactId) {
```
saikumarrs, line 249: endpointPath is not overridden in this if-else block. Is it intentional?

saikumarrs: Worth checking other parts of the code too.

src/v0/destinations/hs/es-retl-v1.ts
```diff
@@ -365,6 +381,7 @@ const legacyBatchEvents = (
           inputs: identifyResponseList,
         };
         batchEventResponse.batchedRequest.endpoint = `${ev.message.endpoint}/batch/create`;
+        batchEventResponse.batchedRequest.endpointPath = `${ev.message.endpointPath}/batch/create`;
```
ItsSudip: NitPick: We can add constants for /batch/create and /batch/update as well.

src/v0/destinations/hs/es-retl-v3.ts
achettyiitr: There were 4 places that were missed. Added those. However, the idea is not to exhaustively search for all places where the endpoint path is not set.

src/v0/destinations/hs/es-retl-v1.ts
```diff
@@ -88,13 +92,18 @@ const processLegacyIdentify = async (
...
     } else if (operation === 'updateObject' && getHsSearchId(message)) {
       const { hsSearchId } = getHsSearchId(message);
-      endpoint = `${CRM_CREATE_UPDATE_ALL_OBJECTS.replace(
-        ':objectType',
+      endpointPath = CRM_CREATE_UPDATE_ALL_OBJECTS_ENDPOINT_PATH.replace(
+        OBJECT_TYPE_PLACEHOLDER,
         objectType,
-      )}/${hsSearchId}`;
+      );
+      endpoint = `${BASE_ENDPOINT}${endpointPath}/${hsSearchId}`;
```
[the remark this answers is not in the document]
achettyiitr: It's better for objectType, we can have dynamic onces.

ItsSudip (approved):

saikumarrs (approved):

PR #5407, opened 2026-07-27
feat: hubspot retl for upsert

achettyiitr:
## Summary

- Adds upsert support to the HubSpot reverse-ETL (rETL) flow. When a record's identifier is a unique property on its object type, the record is written through HubSpot's v3 batch upsert endpoint (`/crm/v3/objects/:objectType/batch/upsert`) — a single create-or-update keyed on `idProperty` — instead of the existing search-then-create/update path. This removes the eventually-consistent CRM Search step (and the create-vs-update 409 races that come with it) for those objects. The change is deliberately narrow: it only applies on the workspace-gated rETL split path and only for the v3 API. The legacy (v1) rETL path, the event-stream path, and associations are all unchanged.
- Resolves INT-6788, INT-6789.

## What changed

- **Endpoint / config** — Added `CRM_UPSERT_ALL_OBJECTS` (`/crm/v3/objects/:objectType/batch/upsert`) and generalised the contacts-only properties URL to `CRM_V3_PROPERTIES_ENDPOINT` (`/crm/v3/properties/:objectType`).

- **Uniqueness check** — Generalised `isLookupFieldUnique` to accept any `objectType`: the cache is now keyed by `destination:objectType` and it fetches `/crm/v3/properties/:objectType`. `objectType` defaults to `contacts`, so the existing event-stream caller needs no change.

- **Router gate** — In `processBatchRouterRetl`, when `apiVersion` is v3 and the batch's `identifierType` is unique for its `objectType`, every event is tagged `upsertObject` and `splitEventsForCreateUpdate` (plus its Search chain) is skipped. Otherwise the batch flows through the existing split logic unchanged. The v3 guard is important: v1 has no upsert handler, so v1 always splits.

- **Payload** — Added an `upsertObject` branch in `processRetlIdentify` (v3), mirroring `processUpsertIdentify`: `identifierType` → `idProperty`, `destinationExternalId` → `id`, `jobId` → `objectWriteTraceId`, with properties built from the populated traits.

- **Batching** — Added an `upsertObject` bucket in `batchRetlEvents` / `batchIdentifyRetl`, modelled on `upsertContacts`: dedup by `id` + `idProperty`, merge properties, and emit against the batch upsert endpoint.

- **Types** — Added `'upsertObject'` to the `operation` union and to `context.hubspotOperation`.

- **Associations** — Unchanged (no upsert endpoint).


src/v0/destinations/hs/retl-transform.ts
```diff
@@ -87,8 +93,32 @@ const processBatchRouterRetl = async (
...
+      // Otherwise, unchanged.
+      const canUpsert =
+        destination.Config.apiVersion === API_VERSION.v3 &&
+        objectType &&
+        identifierType &&
+        (await isLookupFieldUnique(destination, identifierType, metadata, objectType));
+
+      if (canUpsert) {
+        tempInputs = tempInputs.map((input) => {
+          const taggedInput = input;
```
ItsSudip, line 111: `const taggedInput = input` copies the reference, not the object, so the `taggedInput.message.context = {...}` below mutates the caller's `inputs` elements in place. The processing loop at line 134 iterates `inputs` (not `tempInputs`), so the tagging only takes effect through that side-effect — which also makes the `tempInputs` reassignments (lines 110/120) effectively dead, since its only real reader is the catch block at line 127 where it equals `inputs` anyway.

Suggest a proper shallow copy and iterating `tempInputs`:

```ts
tempInputs = tempInputs.map((input) => ({
  ...input,
  message: { ...input.message, context: { ...input.message.context, hubspotOperation: 'upsertObject' as const } },
}));
// then: await Promise.all(tempInputs.map(...))
```

(`splitEventsForCreateUpdate` uses the same mutate-in-place pattern and would benefit from the same treatment, though that's pre-existing.)

ItsSudip (commented):
Review comments on the rETL upsert changes — details inline.

src/v0/destinations/hs/retl-hs-transform-v3.ts
```diff
@@ -103,23 +106,69 @@ const processRetlIdentify = async (
...
     response.source = RETL_SOURCE;
+    recordTransformFlow(destination, 'retl', 'retl', 'association');
     return addHsAuthentication(response, Config);
   }
 
-  // rETL object create/update — associations return above; objects require a resolved operation.
+  // rETL object create/update/upsert — associations return above; objects require a resolved operation.
   if (!operation) {
     throw new InstrumentationError('operation not found');
   }
```
ItsSudip, line 119: `objectType` is required for every path (association routing and object create/update/upsert), so consider moving this validation up right after it's derived:

```ts
const objectType = externalIdInfo?.objectType;
if (!objectType) {
  throw new InstrumentationError('objectType not found');
}
```

Then the association branch above (line 91) can drop its own `objectType &&` guard → `if (String(objectType).toLowerCase() === 'association' && externalIdObj)`, and this check here can be removed. Note this flips the surfaced error from `"operation not found"` → `"objectType not found"` when both are missing — both are `InstrumentationError` (400) and no test pins the message, so it's harmless.

src/v0/destinations/hs/retl-hs-transform-v3.ts
```diff
@@ -201,6 +251,38 @@ const batchIdentifyRetl = (
...
+          throw new TransformationError('rETL - Invalid payload for upsertObject batch');
+        }
+        const { id, idProperty, properties } = json;
+
+        // Deduplicate by id + idProperty (lookup value) - hubspot fails the batch
+        // upsert request if the same id appears more than once.
+        const existing = identifyResponseList.find(
+          (data): data is HubSpotUpsertPayload =>
+            hasUpsertPayloadShape(data) && data.id === id && data.idProperty === idProperty,
+        );
```
ItsSudip, line 271: This metric and `hs_batch_size` (line 203) are emitted from both this split path and the old combined path in `es-retl-v3.ts` (lines 349 / 467) with identical labels, so the two flows can't be distinguished in metrics. Suggest adding a low-cardinality `path` label — e.g. `path: 'split'` here and `path: 'combined'` in `es-retl-v3.ts` — hoisting the value into a named const per file rather than inlining the literal.

src/v0/destinations/hs/config.ts
```diff
@@ -59,6 +60,10 @@ const TRACK_CRM_ENDPOINT_PATH = '/events/v3/send';
 /* CRM ALL Objects */
 const CRM_CREATE_UPDATE_ALL_OBJECTS_ENDPOINT_PATH = '/crm/v3/objects/:objectType';
 
+// Batch upsert for any object type (create-or-update via a unique idProperty)
+// Ref - https://developers.hubspot.com/docs/api/crm/contacts#create-or-update-contacts-upsert
+const CRM_UPSERT_ALL_OBJECTS_ENDPOINT_PATH = '/crm/v3/objects/:objectType/batch/upsert';
+
 /* CRM Association v3 */
 // Ref - https://developers.hubspot.com/docs/api/crm/associations
 const CRM_ASSOCIATION_V3_ENDPOINT_PATH =
```
ItsSudip, line 69: `BATCH_CREATE_CUSTOM_OBJECTS_ENDPOINT_PATH` is defined here and exported (line 139) but not referenced anywhere in the PR or existing source. Suggest removing it and reintroducing it in the follow-up that actually uses it.

src/v0/destinations/hs/retl-hs-transform-v3.ts
```diff
@@ -103,23 +106,69 @@ const processRetlIdentify = async (
...
+    }
+
+    let properties = await populateTraits(propertyMap, traits, destination, metadata);
+    properties = removeHubSpotSystemField(properties);
+
+    // Ref: https://developers.hubspot.com/docs/api/crm/contacts#create-or-update-contacts-upsert
+    const upsertPayload = {
+      id: destinationExternalId,
+      idProperty: identifierType,
+      properties,
```
ItsSudip, line 141: `metadata?.jobId?.toString()` — `Metadata.jobId` is a required `number`, so the `?.` chain hides the real nullability and would silently yield `undefined` for `objectWriteTraceId` if the type ever loosens. Suggest `metadata.jobId.toString()`.

test/integrations/destinations/hs/router/data.ts
```diff
@@ -4836,4 +4838,9 @@ const baseData = [
 ];
 
 // Each rETL fixture is followed by its gated split-path duplicate (see retlSplitData.ts).
-export const data = withRetlSplitCases(baseData as Record<string, any>[]);
+// Dedicated rETL upsert fixtures are appended afterwards so they are NOT auto-derived
+// as behaviour-preservation split cases (see retlUpsertData.ts).
+export const data = [
+  ...withRetlSplitCases(baseData as Record<string, any>[]),
+  ...(retlUpsertData as Record<string, any>[]),
```
ItsSudip: `retlUpsertData as Record<string, any>[]` newly introduces an `any` cast. Suggest typing `retlUpsertData` with the same fixture interface the suite already uses and dropping the cast. (The adjacent `baseData` cast is pre-existing.)

src/v0/destinations/hs/util.test.ts
```diff
@@ -330,7 +330,7 @@ describe('isLookupFieldUnique utility test cases', () => {
     const result = await isLookupFieldUnique(mockDestination as any, 'email', mockMetadata as any);
```
ItsSudip, line 330: These `isLookupFieldUnique` tests all use the `contacts` default (3-arg form). The new `objectType` parameter — separate cache key + separate properties endpoint — has no unit coverage. Suggest adding: (1) `objectType = 'companies'` asserting the cache key is scoped per object type (`<destID>:companies`) and the URL targets `/crm/v3/properties/companies`, and (2) an explicit `contacts` call confirming the default is wired.

test/integrations/destinations/hs/router/retlUpsertData.ts
```diff
@@ -0,0 +1,339 @@
...
+  messageId: `msg-${email}`,
+  originalTimestamp: '2024-01-15T10:00:00.000Z',
+  sentAt: '2024-01-15T10:00:00.000Z',
+  integrations: { All: true },
+});
+
+const UPSERT_ENDPOINT = 'https://api.hubapi.com/crm/v3/objects/contacts/batch/upsert';
+const UPSERT_ENDPOINT_PATH = '/crm/v3/objects/contacts/batch/upsert';
+
+export const retlUpsertData = [
```
ItsSudip: All three upsert fixtures exercise the `hasUniqueValue: true` → upsert path. Consider adding a fixture where the properties mock returns `hasUniqueValue: false` and asserting the events fall back to the `splitEventsForCreateUpdate` create/update flow (not the upsert endpoint), guarding the fallback against regression.

src/v0/destinations/hs/retl-transform.ts
achettyiitr: We have a similar setup in [transform.ts](https://github.com/rudderlabs/rudder-transformer/blob/develop/src/v0/destinations/hs/transform.ts#L118) as well.

src/v0/destinations/hs/retl-hs-transform-v3.ts
achettyiitr: Don't think we need that distinction; I have already added a separate `recordTransformFlow` to understand different flows.

src/v0/destinations/hs/config.ts
achettyiitr: Already fixed it.

src/v0/destinations/hs/retl-hs-transform-v3.ts
achettyiitr: The existing flow has a [similar](https://github.com/rudderlabs/rudder-transformer/blob/feat.hs-upsert-retl/src/v0/destinations/hs/es-retl-v3.ts#L116) check. Basically, I wanted to avoid drift between the previous implementation and the new changes. We can take it as a separate refactoring.

src/v0/destinations/hs/retl-transform.ts
ItsSudip: Confirmed — `transform.ts` uses the same shape (loop over `inputs`, `tempInputs` only in the catch), so keeping it consistent here is the right call for this PR. The mutate-in-place + unused `tempInputs` cleanup would touch both the event-stream and rETL paths, so it's better as a separate follow-up than diverging one file. Fine to leave as-is. 👍

src/v0/destinations/hs/retl-hs-transform-v3.ts
ItsSudip: Makes sense — `hs_transform_flow` with `code_path` (`retl` vs `es_retl`) already captures the split-vs-combined distinction, with more context than a label would have. No need for `path` then. One minor note for later: it counts invocations by flow, so `hs_batch_size` / `hs_upsert_duplicate_trace_id` still aren't sliceable *by* flow — only relevant if we ever want batch-size distributions per flow. Not needed now. 👍

src/v0/destinations/hs/retl-hs-transform-v3.ts
ItsSudip: Fair — keeping it aligned with the existing `es-retl-v3.ts` check avoids drift. Agreed to handle the `?.` tightening as a separate refactor across both paths. 👍

ItsSudip (approved):

PR #5408, opened 2026-07-27
feat(braze_audience): add BRAZE_AUDIENCE router transform and network handler

bgolubovic-rs:
## Summary
- Net-new `BRAZE_AUDIENCE` destination: mirror-mode `record` → Braze `POST /users/track/bulk` boolean custom attributes.
- Router transform + network handler with partial-failure (`errors[]`) → per-job 200/400 mapping for transformerProxy.
- Soft-bounce empty/whitespace/`object` `external_id`; reject reserved `customAttributeName === 'external_id'`.
- Identity hard-bounce: Braze enums **or** live message patterns (e.g. `"'external_id' must be fewer than 988 bytes"`) — live Braze does not always return `EXTERNAL_USER_ID_TOO_LARGE`.
- [ACT2-387](https://linear.app/rudderstack/issue/ACT2-387) · parent [ACT2-384](https://linear.app/rudderstack/issue/ACT2-384).

## Merge order
1. rudder-integrations-config
2. **This PR (transformer)**
3. rudder-integrations-info (parallel OK)
4. rudder-webapp
5. rudderstack-operator transformerProxy **after this merges**

## Sibling PRs
| Repo | PR |
| --- | --- |
| integrations-config | https://github.com/rudderlabs/rudder-integrations-config/pull/2613 |
| transformer | https://github.com/rudderlabs/rudder-transformer/pull/5408 |
| integrations-info | https://github.com/rudderlabs/rudder-integrations-info/pull/683 |
| webapp | https://github.com/rudderlabs/rudder-webapp/pull/9847 |
| operator (proxy) | https://github.com/rudderlabs/rudderstack-operator/pull/1515 |
| specs (already open) | https://github.com/rudderlabs/rudder-specs/pull/108 |

## QA scorecard (this repo)
| Check | Result |
| --- | --- |
| Unit: routerTransform + utils (reserved attr / normalizeExternalId / identity hard-bounce patterns) | Pass |
| Integration: `braze_audience` router fixtures + component tests | Pass |
| Pre-commit: `test:ts` + `test:js` + eslint/prettier | Pass |

## Mini / Braze E2E (re-verified 2026-07-27 on mini transformer `:19090` + Braze US-03)
| Case | Result |
| --- | --- |
| Offline verify (`make offline`, 18 checks) | Pass |
| Router: INSERT+DELETE membership batch (`connection` required) → batched 200 + `/users/track/bulk` | Pass |
| Router: `customAttributeName=external_id` → 400 reserved | Pass |
| Router: object `external_id` → soft-bounce (missing/empty after trim) | Pass |
| Live proxy mixed batch: valid / oversized / valid → **200 / 400 / 200**; Braze error msg `"'external_id' must be fewer than 988 bytes"` | Pass (post live-message hard-bounce fix) |
| Test connection empty `{attributes:[]}` → **400** `No data parsed` (= auth OK) | Pass |
| D6 429 | Skipped |
| Hosted E6 (prod proxy) | Blocked until operator merges |

## Test plan
- [ ] CI green on develop
- [ ] Confirm proxy PR lands after this for production partial-failure reporting
- [x] Mini live proxy: oversized id hard-bounce uses live message pattern (not enum-only)


src/v0/destinations/braze_audience/types.ts
```diff
@@ -0,0 +1,57 @@
...
+});
+
+export const BrazeAudienceConnectionConfigSchema = z
+  .object({
+    customAttributeName: z
+      .string()
+      .min(1)
+      .max(255)
+      .regex(/^\w+$/, 'customAttributeName must be alphanumeric or underscore'),
+    syncMode: z.literal('mirror').optional(),
```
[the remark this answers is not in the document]
bgolubovic-rs: Fixed in latest push — addressed as discussed.

src/v0/destinations/braze_audience/utils.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const buildBulkBody = (
+  attributes: BrazeAudienceAttributePayload[],
+): { attributes: BrazeAudienceAttributePayload[] } => ({ attributes });
+
+/** Trim and require a non-empty external_id (sources already key identifiers by destination field). */
+export const normalizeExternalId = (raw: unknown): string | null => {
+  if (raw === null || raw === undefined) return null;
+  const value = String(raw).trim();
+  return value.length > 0 ? value : null;
```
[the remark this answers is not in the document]
bgolubovic-rs: Fixed in latest push — addressed as discussed.

src/v1/destinations/braze_audience/networkHandler.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  const failedByIndex = new Map<number, { type?: string; message: string }>();
+  for (const err of errors) {
+    if (typeof err.index === 'number') {
+      failedByIndex.set(err.index, {
+        type: err.type,
+        message: err.type || 'braze_partial_error',
+      });
+    }
+  }
```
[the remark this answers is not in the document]
bgolubovic-rs: Fixed in latest push — addressed as discussed.

src/v1/destinations/braze_audience/networkHandler.ts
```diff
@@ -0,0 +1,148 @@
...
+      /external_id|invalid.*(user|id)|user.?not.?found/i.test(fail.type);
+
+    if (isIdentity) {
+      stats.increment('braze_audience_hard_bounce', {
+        destinationId,
+        workspaceId,
+        reason: 'identity',
+      });
+      return { statusCode: 400, metadata, error: fail.message };
+    }
```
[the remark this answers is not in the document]
bgolubovic-rs: Fixed in latest push — addressed as discussed.

src/v1/destinations/braze_audience/networkHandler.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  if (!isHttpStatusSuccess(status)) {
+    const errorMessage = JSON.stringify(response?.message ?? response) || 'unknown error';
+    const responseWithIndividualEvents = rudderJobMetadata.map((metadata) => ({
+      statusCode: status,
+      metadata,
+      error: errorMessage,
+    }));
+
+    if (status === 401) {
```
shekhar-rudder: Wondering if we need destination-specific metrics here — `braze_audience_hard_bounce`/`sync_disabled` on 401 and `braze_audience_soft_bounce` on 429/5xx (lines 64-79) look like they overlap with the generic `outgoing_request_count{destType, statusCode}` metric and the retry classification `getDynamicErrorType(status)` already attaches to the thrown error below. Same goes for `braze_audience_batch_sent` at line 105, which looks like it duplicates the generic `output_batch_size` histogram already fired in `processBatchedDestination.ts`. Might be worth trimming these — generic success/failure/throughput counts probably belong at the framework layer rather than duplicated per destination.

src/v0/destinations/braze_audience/routerTransform.ts
```diff
@@ -0,0 +1,93 @@
...
+  private get connectionConfig(): BrazeAudienceConnectionConfig {
+    return this.connection!.config.destination;
+  }
+
+  transformEvent(
+    input: z.infer<typeof BrazeAudienceRouterRequestSchema>,
+  ): TransformedEvent<BrazeAudienceAttributePayload> {
+    const { message } = input;
+    const messageAction = message.action;
+    if (!messageAction) {
```
shekhar-rudder: nit: this check (and `attrValue` below) might not be reachable anymore — `RecordMessageSchema.action` is `z.nativeEnum(RecordAction)` (required) and gets validated before `transformEvent` runs, and `ACTION_ATTR_VALUE`'s keys line up exactly with `RecordAction`'s values. Could probably drop both.

src/v1/destinations/braze_audience/networkHandler.ts
```diff
@@ -0,0 +1,184 @@
...
+        });
+        return { statusCode: 500, metadata, error: unindexedMessage };
+      }
+      return { statusCode: 200, metadata, error: 'success' };
+    }
+
+    const isIdentity = isIdentityHardBounce(fail.type);
+
+    if (isIdentity) {
+      stats.increment('braze_audience_hard_bounce', {
```
shekhar-rudder: nit: small naming thing — "hard bounce"/"soft bounce" (here and lines 134/155) doesn't quite match how the rest of the repo talks about this (`tags.js` uses `RETRYABLE`/`ABORTED`). Might read more consistently as something like `braze_audience_aborted`/`braze_audience_retryable`.

src/v0/destinations/braze_audience/types.ts
```diff
@@ -0,0 +1,60 @@
...
+  })
+  .passthrough();
+
+export type BrazeAudienceConnectionConfig = z.infer<typeof BrazeAudienceConnectionConfigSchema>;
+
+const RecordMessageSchema = z
+  .object({
+    type: z.literal('record'),
+    action: z.nativeEnum(RecordAction),
+    identifiers: z.record(z.unknown()).optional(),
```
shekhar-rudder: nit: could add a `z.union([z.string(), z.number()])` type constraint on `external_id` here, so the wrong-type case gets caught at schema validation instead of inside `transformEvent`. The trim/empty/finite-number logic can stay right where it is in `normalizeExternalId` — no need for `.refine()`/`.transform()`.

src/v0/destinations/braze_audience/types.ts
```diff
@@ -0,0 +1,60 @@
...
+  .object({
+    customAttributeName: z
+      .string()
+      .min(1)
+      .max(255)
+      .regex(/^\w+$/, 'customAttributeName must be alphanumeric or underscore')
+      .refine((value) => value !== 'external_id', {
+        message: 'customAttributeName cannot be external_id',
+      }),
+    syncMode: z.literal('mirror').optional(),
```
shekhar-rudder: nit: `syncMode` (and `schemaVersion` below, plus `fields` on `RecordMessageSchema`) don't look like they're read anywhere in this PR. Worth removing if not needed soon, or a quick comment noting they're reserved for later.

src/v0/destinations/braze_audience/routerTransform.ts
```diff
@@ -0,0 +1,93 @@
...
+class BrazeAudienceIntegration extends BatchDestination<
+  BrazeAudienceAttributePayload,
+  typeof BrazeAudienceRouterRequestSchema
+> {
+  private readonly headers: Record<string, string>;
+
+  constructor(...args: ConstructorParameters<typeof BatchDestination>) {
+    super(...args);
+    if (!this.connection) {
+      throw new InstrumentationError(`Connection config is required for ${DESTINATION_TYPE}`);
```
shekhar-rudder: nit: `DESTINATION_TYPE` is only used once, right here — might not be worth its own exported constant. Could inline it or use `this.destination.DestinationDefinition?.Name` instead.

shekhar-rudder (commented):
A few small comments on the network handler / schema — nothing blocking.

src/v1/destinations/braze_audience/networkHandler.ts
```diff
@@ -0,0 +1,184 @@
+import { isHttpStatusSuccess } from '../../../v0/util/index';
+import { proxyRequest, prepareProxyRequest } from '../../../adapters/network';
+import { getDynamicErrorType, processAxiosResponse } from '../../../adapters/utils/networkUtils';
+import { TransformerProxyError } from '../../../v0/util/errorTypes';
+import tags from '../../../v0/util/tags';
+import type { DeliveryJobState, DeliveryV1Response, ProxyMetdata } from '../../../types';
+
+const stats = require('../../../util/stats');
```
shekhar-rudder: **[nit]** `networkHandler.ts:8` uses `const stats = require('../../../util/stats')` (CommonJS) while its own test file and every sibling networkHandler (`braze`, `iterable_audience`) use ES `import stats from '../../../util/stats'`. Functionally fine under the jest mock, but inconsistent with the rest of the codebase's style.

src/v0/destinations/braze_audience/routerTransform.ts
bgolubovic-rs: Good catch — dropped both unreachable checks. Also typed `ACTION_ATTR_VALUE` as `Record<RecordAction, boolean>` so the lookup is guaranteed after Zod validation.

src/v1/destinations/braze_audience/networkHandler.ts
bgolubovic-rs: Renamed — `braze_audience_aborted` / `braze_audience_retryable` (and matching helpers/comments) to align with `tags.js` `ABORTED`/`RETRYABLE`.

src/v0/destinations/braze_audience/types.ts
bgolubovic-rs: Done — `identifiers.external_id` is now `z.union([z.string(), z.number()]).optional()` (passthrough object). Wrong types fail at schema validation; trim/empty/finite still live in `normalizeExternalId`.

src/v0/destinations/braze_audience/types.ts
bgolubovic-rs: Removed — dropped unused `syncMode`, `schemaVersion`, and `fields` from the Zod schemas. `.passthrough()` still accepts them on the wire from webapp/setup.

src/v0/destinations/braze_audience/routerTransform.ts
bgolubovic-rs: Removed — inlined via `this.destination.DestinationDefinition.Name` and dropped the unused constant.

src/v1/destinations/braze_audience/networkHandler.ts
bgolubovic-rs: Fixed — switched to `import stats from '../../../util/stats'` to match sibling networkHandlers. `networkHandler` unit tests still green.

src/v1/destinations/braze_audience/networkHandler.ts
bgolubovic-rs: Agreed — trimmed the overlapping ones.

Kept only Braze-specific classification that framework metrics miss:
- `braze_partial_failure` (destination-tagged) on 2xx + `errors[]`
- per-record `braze_audience_aborted` / `braze_audience_retryable` with `reason: identity | partial | partial_unindexed`

Removed HTTP-status `aborted`/`retryable`/`sync_disabled` and `braze_audience_batch_sent` (covered by `outgoing_request_count` / `getDynamicErrorType` / `output_batch_size`). Observability spec updated to match.

src/v0/destinations/braze_audience/types.ts
```diff
@@ -0,0 +1,62 @@
...
+const RecordMessageSchema = z
+  .object({
+    type: z.literal('record'),
+    action: z.nativeEnum(RecordAction),
+    identifiers: z
+      .object({
+        external_id: z.union([z.string(), z.number()]).optional(),
+      })
+      .passthrough()
+      .optional(),
```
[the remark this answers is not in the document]
bgolubovic-rs: Good catch. `external_id` is now `z.unknown().optional()` so object/array/boolean values reach `normalizeExternalId` and soft-bounce with `InstrumentationError` per record. Regression test updated accordingly.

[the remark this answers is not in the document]
bgolubovic-rs: Leave `empty/whitespace/NaN` soft-bounce in `normalizeExternalId`

src/v0/destinations/braze_audience/types.ts
shekhar-rudder: The `identifiers.external_id` field is now typed via `z.union([z.string(), z.number()]).optional()` in `RecordMessageSchema`, and the only call site (`normalizeExternalId(message.identifiers?.external_id)` in `routerTransform.ts`) always passes a `string | number | undefined`. Could we tighten `normalizeExternalId`'s parameter from `unknown` to `string | number | undefined`?

src/v1/destinations/braze_audience/networkHandler.ts
```diff
@@ -0,0 +1,159 @@
...
+      return { statusCode: 200, metadata, error: 'success' };
+    }
+
+    const isIdentity = isIdentityAborted(fail.type);
+
+    if (isIdentity) {
+      stats.increment('braze_audience_aborted', {
+        destinationId,
+        workspaceId,
+        reason: 'identity',
```
shekhar-rudder: **[nit]** This is the only call site for `braze_audience_aborted`, so `reason: 'identity'` is always constant and adds no distinguishing information — remove the `reason` label here. (Note: `braze_audience_retryable` below has two distinct reason values — `'partial'` vs `'partial_unindexed'` — so that one should keep its `reason` label.)

src/v1/destinations/braze_audience/networkHandler.ts
```diff
@@ -0,0 +1,159 @@
...
+      '',
+      responseWithIndividualEvents,
+    );
+  }
+
+  const errors = response?.errors ?? [];
+
+  if (errors.length > 0) {
+    // Destination-tagged — never the untagged event-stream braze_partial_failure.
+    stats.increment('braze_partial_failure', {
```
shekhar-rudder: **[medium]** Rename this to `braze_audience_partial_failure` instead of disambiguating via the `destination` tag on the shared `braze_partial_failure` name (already used untagged by v0 `braze` and tagged by v1 `braze`). This also sidesteps a likely bug: `src/util/prometheus.js:629-633` pre-registers `braze_partial_failure` with `labelNames: []`, so tagged `increment` calls against the existing counter (this one and v1 braze's) may throw internally and get silently swallowed. A new metric name avoids that entirely.

src/v0/destinations/braze_audience/routerTransform.ts
```diff
@@ -0,0 +1,87 @@
...
+    const attrValue = ACTION_ATTR_VALUE[message.action];
+
+    const { customAttributeName } = this.connectionConfig;
+    if (!customAttributeName) {
+      throw new InstrumentationError('customAttributeName is required on connection config');
+    }
+
+    const externalId = normalizeExternalId(message.identifiers?.external_id);
+    if (!externalId) {
+      // Soft-bounce this record (BatchDestination maps InstrumentationError per-item).
```
shekhar-rudder: **[blocking]** This comment is incorrect — `InstrumentationError` sets status `400`, which this codebase's convention treats as a permanent failure (retryable is 429/5xx). `BatchDestination` will abort this record, not soft-bounce/retry it. Update the comment to say aborted.

src/v1/destinations/braze_audience/networkHandler.ts
bgolubovic-rs: Done — removed `reason` from `braze_audience_aborted`. Left it on `braze_audience_retryable` for `partial` vs `partial_unindexed`.

src/v0/destinations/braze_audience/routerTransform.ts
bgolubovic-rs: Done — comment now says abort (400 per-item), not soft-bounce.

src/v1/destinations/braze_audience/networkHandler.ts
bgolubovic-rs: Done — renamed to `braze_audience_partial_failure` (labels: `destinationId`, `workspaceId`) and registered it in `prometheus.js` alongside `braze_audience_aborted` / `braze_audience_retryable`. No longer touches the untagged event-stream `braze_partial_failure` counter.

src/v0/destinations/braze_audience/types.ts
bgolubovic-rs: Done — `normalizeExternalId` is now `(raw: string | number | undefined)`. Wrong shapes stay rejected at Zod; the helper handles empty / non-finite only.

krishna2020:
Could we add `test/integrations/destinations/braze_audience/dataDelivery/` (`data.ts` + `network.ts`), following `iterable_audience`?

**Reason:** all 8 cases in `networkHandler.test.ts` pass `status: 201`, so the `!isHttpStatusSuccess(status)` → `TransformerProxyError` branch (`networkHandler.ts:55-72`) has no coverage — that's the entire auth-failure / 5xx / rate-limit path.

Two gaps the direct-call unit test also can't reach by construction:

- **Handler resolution** — `networkHandlerFactory.js:30` `require`s the module inside a bare `try {} catch {}`, so a module-load error silently falls back to the generic handler with no signal.
- **Response shape** — `processAxiosResponse` (`networkUtils.js:140-157`) returns `response` as a string or `''` on failures, not the object shape `BrazeAudienceProxyParams` declares, so the hand-built fixture doesn't represent the error path.

Minimum ask is a couple of non-2xx cases; the full `dataDelivery` suite additionally covers the wiring and a nock-produced response shape. `iterable_audience/dataDelivery/network.ts` has the pattern to copy (`authErrorResponse` / `serverErrorResponse`).

bgolubovic-rs:
Added `test/integrations/destinations/braze_audience/dataDelivery/` (`data.ts` + `network.ts`) following `iterable_audience`.

Cases:
- 2xx empty `errors[]` → all jobs 200
- 2xx + indexed identity error → per-job 400
- 2xx + unknown indexed error → per-job 500
- 401 / 429 / 500 → fan-out with `aborted` / `throttled` / `retryable`

This exercises handler resolution through `networkHandlerFactory` and the `processAxiosResponse` shape on the non-2xx path.

shekhar-rudder (dismissed):

krishna2020 (dismissed):

test/integrations/destinations/braze_audience/live/api.ts
```diff
@@ -0,0 +1,115 @@
...
+  return value;
+};
+
+export const customAttributeName = (ctx: RunContext): string => {
+  const fromResource = ctx.liveSecret.resourceIds?.customAttributeName;
+  if (typeof fromResource === 'string' && fromResource.trim().length > 0) {
+    return fromResource;
+  }
+  return LIVE_MEMBERSHIP_ATTR;
+};
```
achettyiitr: Should we error out if we don't find it?

test/integrations/destinations/braze_audience/live/api.ts
```diff
@@ -0,0 +1,115 @@
...
+
+export const externalId = (ctx: RunContext): string => ctx.identity('user');
+
+const restApiKey = (ctx: RunContext): string => {
+  const fromReadback = ctx.liveSecret.readback?.restApiKey;
+  if (typeof fromReadback === 'string' && fromReadback.trim().length > 0) {
+    return fromReadback;
+  }
+  return readString(ctx.liveSecret.config?.restApiKey, 'config.restApiKey');
```
achettyiitr: Since both of readback and restApiKey are same, we can ignore it and use only restApiKey?

test/integrations/destinations/braze_audience/live/verify.ts
```diff
@@ -0,0 +1,10 @@
+import { RunContext } from '../../../live/types';
+import { fetchMembership } from './api';
+
+/** Assert export/ids shows the expected boolean membership custom attribute. */
+export const verifyMembership =
+  (expected: boolean) =>
+  async (ctx: RunContext): Promise<void> => {
+    const value = await fetchMembership(ctx);
+    expect(value).toBe(expected);
+  };
```
achettyiitr: Since it's only one function, can we keep within the spec.ts?

test/integrations/destinations/braze_audience/live/setup.ts
```diff
@@ -0,0 +1,19 @@
+import { RunContext } from '../../../live/types';
+import { fetchMembership, setMembership } from './api';
+import { pollUntil } from '../../../live/poll';
+
+/** Seed membership=true and wait until export/ids can see it (leave scenario precondition). */
+export const seedMembershipTrue = async (ctx: RunContext): Promise<void> => {
```
achettyiitr: We can even move this within setup.ts?

test/integrations/destinations/braze_audience/live/api.ts
bgolubovic-rs: Good call — now required via `resourceIds.customAttributeName` (throws if missing). No silent fallback.

test/integrations/destinations/braze_audience/live/api.ts
bgolubovic-rs: Simplified — live helpers now use only `config.restApiKey` (same key for track + export/ids).

test/integrations/destinations/braze_audience/live/verify.ts
bgolubovic-rs: Done — folded into `spec.ts` and removed `verify.ts`.

test/integrations/destinations/braze_audience/live/setup.ts
bgolubovic-rs: Folded `seedMembershipTrue` into `spec.ts` as well (same single-function rationale) and removed `setup.ts`.

achettyiitr (approved):

PR #5410, opened 2026-07-28
feat: send optional auth header on config backend fetches

akashrpo:

New `src/util/configBackend.js` exporting `CONFIG_BACKEND_URL` and `configBackendHeaders()`.
The helper builds a Basic auth header from `CONFIG_BACKEND_HOSTED_SECRET` (secret as username),
or returns `{}` when unset. All config-backend fetches now pass it: transformation code (v0 and v1),
transformation libraries, rudder libraries, tracking plans. Drops the `CONFIG_BACKEND_URL`
constant duplicated across three files.

Resolves RUD-3006

Hosted config backend is adding auth on these routes; without a credential the transformer
would get 401s once that lands. Dependent change: config backend must accept the secret via
Basic auth before the env var is set in hosted deployments.

None when `CONFIG_BACKEND_HOSTED_SECRET` is unset - no header added, requests unchanged.
Keeps self-hosted setups working as-is.

N/A

`src/util/configBackend.js` - centralises the config-backend base URL and header construction
so future callers import one helper.

Env var is read at module load, so changing it needs a process restart. Secret is sent as a
header only, never logged or in a URL.

@coderabbitai review

<hr>

### Developer checklist

### Reviewer checklist


src/util/configBackend.js
```diff
@@ -0,0 +1,14 @@
+const CONFIG_BACKEND_URL = process.env.CONFIG_BACKEND_URL || 'https://api.rudderlabs.com';
+
+// Sent to the config backend when it requires authentication (basic auth, secret as username).
+// Unset for self-hosted config backends that serve these routes without auth, so no header is added.
+const { CONFIG_BACKEND_HOSTED_SECRET } = process.env;
+
+// Empty when unset, so fetchWithProxy keeps the call unauthenticated (self-hosted).
+function configBackendRequestOptions() {
```
fracasula, line 7: We don't have any tests on this PR either. If we get it wrong every fetch is a  401.

- secret set → exactly `Basic base64("<secret>:")`, one colon.
- secret unset → `{}`, **not** `{ headers: {} }`.
   - this was already wrong in the first commit you did

src/util/configBackend.js
```diff
@@ -0,0 +1,14 @@
+const CONFIG_BACKEND_URL = process.env.CONFIG_BACKEND_URL || 'https://api.rudderlabs.com';
+
+// Sent to the config backend when it requires authentication (basic auth, secret as username).
+// Unset for self-hosted config backends that serve these routes without auth, so no header is added.
+const { CONFIG_BACKEND_HOSTED_SECRET } = process.env;
```
fracasula: This is not documented anywhere. It is absent from `sample.env`, `README.md`, `CONTRIBUTING.md`, `docker-compose.yml`, `Dockerfile`, `Dockerfile-ut-func`, and every workflow in `.github/`.

fracasula: Additionally, this secret is never trimmed nor validated. A trailing newline or space (trivially introduced via a k8s secret file or a `.env` line) would never match. Failing fast at boot beats a 401 discovered hours later in my opinion.

Also, if I'm not mistaken, a per-event 401 is aborted, hence never retried. That would lead to data loss, right?

fracasula (changes requested):
@akashrpo @ktgowtham if I understand this correctly, if we were to misconfigure secrets (e.g. trailing space, mistake while doing secret rotation, etc...), then we would return a 401 per event (HTTP response for the whole batch would still be a 200).

If that's the case we would effectively skip those events in the processor. I suggest we do the same as [809](https://github.com/rudderlabs/rudder-server/blob/3e126b2027150b3aaae85500a658e3ef9b9d724c/processor/internal/transformer/utils.go#L21) and return a special response that we can agree on so that these can be retried and we can set alerts on them as well so that:
* if it's an operator mistake, fix it
* if it's a malicious actor, we can look into it and manually disable the account, report it, etc...

Wdyt?

test/__tests__/util.utils.test.js
```diff
@@ -1,4 +1,37 @@
-const { constructValidationErrors } = require("../../src/util/utils");
+const { constructValidationErrors, responseStatusHandler } = require("../../src/util/utils");
+
+describe('responseStatusHandler', () => {
+  it('does not throw on 200', () => {
+    expect(() => responseStatusHandler(200, 'Transformation', 'v1', 'url')).not.toThrow();
+  });
+
+  it('throws a retriable 809 on 401 (config backend auth failure), so the event is retried not dropped', () => {
```
fracasula: @akashrpo my bad, I suggested 809 but that should be used only when the CP is actually down.

We should use `503` + `X-Rudder-Should-Retry: true` instead of `809`, so it matches `rudder-pytransformer` as well.

Same pattern we already use in `src/middlewares/memoryFencing.ts:186-190`:

```ts
ctx.set('X-Rudder-Should-Retry', 'true');
ctx.set('X-Rudder-Error-Reason', 'config_backend_auth_failed');
ctx.status = 503;
```

The headers are mandatory: rudder-server only retries a 503 that carries `X-Rudder-Should-Retry: true`. A bare 503 gets bounded retries and then a panic instead.

Use a distinct reason string (`config_backend_auth_failed`, not a generic one) so a wrong secret is
distinguishable from the config backend being down.

Note this repoints the alerting: `processor_control_plane_down` stops firing for this service, and
`transformer_client_perpetual_retry_count{reason=...}` becomes the signal cc @ktgowtham

fracasula (changes requested):

fracasula:
As per our conversation on Slack we should handle 403s as well. Once we flip the switch (i.e. BLOCK_PUBLIC_TRANSFORMATION_ROUTES=true) we'll start returning 403s. If that happens it would most likely be due to a misconfiguration on our side. Unfortunately 403s are dropped as of now, it'd be better to add a dedicated counter and alert on a config-backend 403 instead of silently dropping those events.

fracasula (commented):
@akashrpo I think once we add the 403s management here we're good. I also got this from a test subagent:

### No test proves the header reaches the wire

There are four unit tests on `configBackendRequestOptions()` itself, but **nothing asserts that any of the five call sites actually sends it**. `test/mocks/config-backend-server.ts` serves all three transformation routes (`:64`, `:95`, `:116`) with **no `Authorization` check**, and there is no 401/403 negative test anywhere in `test/`. The nock-based suites ignore request headers by default.

So a regression that dropped the options argument at a call site would be invisible. Add an assertion in the mock server plus one negative case. The mock also serves **no tracking-plan route at all**, so that call site has no e2e coverage regardless.

fracasula (dismissed):
LGTM, but the mocks still assert nothing about Authorization. There is no auth check and no 401/403 negative case, so a regression dropping the header remains invisible. I suggest doing this one last fix before merging.

fracasula (approved):

ktgowtham (approved):
LGTM. 
@fracasula @akashrpo 
do we know if the CONFIG_BACKEND_HOSTED_SECRET is already accessible by UT pods(both shared and dedicated). If not, we might need to do some devops, operator changes

Jayachand (approved):

ItsSudip (approved):

PR #5418, opened 2026-07-30
feat(nativeBatching): fold NetworkHandler into the BatchDestination class

shekhar-rudder:
Moves delivery response handling from a separate `networkHandler.ts` file onto the destination's own `BatchDestination`-family class, and stops expressing it in terms of the delivery API.

Design, evidence and per-destination parity tables: `docs/superpowers/specs/2026-07-30-network-handler-abstraction-design.md`.

## The contract

An integration declares a **single `delivery` object** — a status-keyed map plus an optional error-message extractor. Applying it is framework-owned: `handleDeliveryResponse(Class, ctx)` is a free function in `delivery.ts`, so there is no member to override and no `super` to call.

```ts
type StatusOverride = (ctx: DeliveryContext, fallback: () => Verdict) => HandleResponseResult;

type DeliverySpec = {
  statusOverrides?: Partial<Record<StatusKey, StatusOverride>>;
  failureReason?: (ctx: DeliveryContext) => string;
};

static readonly delivery: DeliverySpec;
```

Both are grouped under `delivery` because everything else on a `BatchDestination` — `transformEvent`, `getBatchStrategy`, `getInputSchema` — is about turning an event into a request, while these two are about reading the response that comes back: a different phase, behind a different flag. As bare statics they read as more transform surface. Each destination's `delivery.ts` exports one spec object, so the class carries one line for delivery however much that contract grows.

Keys are an exact status or a class (`'2xx'` / `'4xx'` / `'5xx'`), exact winning. A handler that owns only *some* of the responses carrying its status — mixpanel's `400` applies to `/import` but not `/engage` — calls `fallback()`, so the return type stays total with no sentinel value. The spec is **resolved down the prototype chain** by `resolveDeliverySpec`: static properties are shadowed rather than merged, so a subclass declaring one would otherwise silently drop an ancestor's entries with no type or runtime error. `statusOverrides` merges key-by-key child-over-parent; `failureReason` takes the nearest declaration, which is what static-method shadowing did.

Verdicts are `success` / `abort` / `retry`, with throttling and the two auth categories as refinements, because that is what rudder-server can act on — `REFRESH_TOKEN` ends as a 500 and retries, `AUTH_STATUS_INACTIVE` ends as a 400 and aborts (§2.1 traces this).

**It never infers an auth verdict from a status code.** `getAuthErrCategoryFromStCode`'s unconditional 401→`REFRESH_TOKEN` is wrong more often than right — a 401 from an API-key destination is a bad key, not a refreshable token — and guessing costs a real control-plane token refresh. A destination that genuinely wants the mapping declares it; `gaec` derives its category from the response body, as it always did.

## What the response looks like

Shaped to what rudder-server actually reads, verified against its source rather than assumed:

| field | behaviour | why |
|---|---|---|
| `status` | `ctx.status`, passed through | `ProxyResponseV1` has no `status` field (`transformer_proxy_adapter.go:40-44`) — disposition comes from each job state. Relabelling a mixed batch `207` would invent a difference from the legacy handler for no reader. `am/networkHandler.ts:75` records the same fact in a comment. |
| `message` | the shared failure reason when every failure agrees on one; `<n> of <m> events failed; see per-event errors` when they differ | quoting the first of several would present one job's error as the batch's. Also the only place the batch-level reason survives, which a fixed "partial failures" string discarded. |
| `statTags` | the whole tag set when every job failed the same way; absent otherwise | drives `integration.failure_detailed` (`processor/integrations/integrations.go:29-37`), which tags the response as a whole — only honest when the response as a whole is one failure. The bridge supplies `errorCategory`/`errorType`, `deliver()` merges the identifying tags from the same `getTags` metadata `postTransformation` uses for a thrown error, so returned and thrown failures are tagged identically. |
| `destinationResponse` | not echoed | confirmed unread: not a field on `ProxyResponseV1`; live events and the debugger ship the per-job `error` (`routerJobResponseBodys`); the reporting extractor's `destinationResponse` lookup runs against that stored per-job string. |

Error messages are **strings, not JSON blobs**, and **each integration owns its own extraction**. The per-job `error` is what the v1 adapter stores as the job's response body (`transformer_proxy_adapter.go:169`) — what live events show and what error reporting groups on — so a string is returned bare rather than JSON-quoted.

The framework itself does not parse the body. `delivery.failureReason(ctx)` defaults to the status-only string `genericNetworkHandler` already produces (`[Generic Response Handler] Request failed with status: <n>`), and a destination that wants its API's error text declares one against that API's shape.

An earlier revision had a shared `messageFromResponse` helper that searched the body for `message`, then `msg`, then `error`, then `description`. It was withdrawn: there is no shared convention on the legacy path to generalise — customerio joins `reason`/`field`/`message` from its own `errors[]`, gaec reads `error.message`, braze_audience does `JSON.stringify(response?.message ?? response)`, iterable_audience uses `params ?? msg ?? message`, and `genericNetworkHandler` reads nothing. A key-ordered search looks like it unifies those but only imposes one destination's field order on every other, and a quirk added for one leaks into all: Iterable's `params`-before-`msg` precedence meant an object-valued early key silently masked a good message on a later key for *every* destination on the default list, so `{ error: {}, description: 'real message' }` returned `{}`.

## Destinations migrated

Every batching-framework destination that has its own network handler, because gating delivery on the flag switches the path for all of them at once (§4.0):

| destination | overrides | parity |
|---|---|---|
| `customerio` | `207` | 5/5 branches vs the legacy handler |
| `gaec` | `'2xx'`, `'4xx'` | 8/8 vs `gaecResponseHandler` |
| `iterable_audience` | `'2xx'` | 7/7 vs `AudienceListStrategy` |
| `braze_audience` | `'2xx'` | 15/15 vs the legacy handler, plus counter-for-counter |
| `test_destination` | none needed | 6/6 statuses vs `genericNetworkHandler` |

Each per-item list is indexed off the **posted request body**, not off the response, because that is the array guaranteed to line up with the job list. `gaec` drove its list off `results` at first, which diverged whenever Google omitted or truncated it — the existing `gaec_v1_scenario_2` fixture does exactly that — and turned an abort into a batch retry that could never converge, since re-uploading accepted adjustments returns duplicate-enhancement failures on every attempt.

`iterable_audience` is the only Pattern D case — Iterable reports failures by *identity*, not index. Both of its deliberate-success branches carry over: a GDPR-forgotten user is counted and accepted rather than aborted, and `notFound` on an unsubscribe is a no-op success.

`test_destination` needs nothing — it composes `genericNetworkHandler` and replaces only `proxy`, which the framework default already reproduces. A test holds that equivalence so it fails if the two ever drift.

## Behaviour changes

Two, both `customerio`, both in §4.1:

1. **The top-level `message` text.** Framework-generated; no per-job field derives from it.
2. **The status-inferred `authErrorCategory` is dropped.** It is unreachable for a Basic-auth destination — `OAuthTransport.RoundTrip` returns at `if !isOauthDestination` before `postRoundTrip`, so neither the token refresh nor the 401→500 rewrite could ever fire. This takes the *transformer's own* HTTP status for those responses from 401/403 to 200, which is read in exactly one place: the `statusCode` label on the `transformer_outgoing_request` metric (`router/worker.go:528`). **No job outcome changes** — the per-job `statusCode` stays 401/403.

`gaec`'s 401 moves `errorType` from `aborted` to `retryable`, and that one is a correction: `gaec` *is* OAuth-backed (`auth: { type: 'OAuth', rudderScopes: ['delivery'] }`), so a successful refresh makes rudder-server return 500 and retry the batch (`services/oauth/v2/http/transport.go:195`) — the legacy tag described the opposite of what happens.

## Gated behind its own flag, default off

`isBatchingFrameworkDeliveryEnabled(destType, workspaceId)`, env `{DEST}_BATCHING_FRAMEWORK_DELIVERY_ENABLED_WORKSPACE_IDS`. Two deliberate properties: **no GA map**, so already-GA destinations keep the legacy handler until a workspace is named; and it **requires the transform flag**, because the delivery path has to interpret a payload built by the matching transform path. Nothing changes in production until a workspace opts in.

The branch is guarded by an `isProxyV1Request` type predicate rather than a version string — the array-ness of `metadata` is what the code actually depends on, and narrowing on it removes the `ProxyV1Request` casts. `ProxyV1RequestSchema` was rejected: it is unused anywhere and marks `secret`/`dontBatch` required, so a real payload omitting either would fail validation and fall through to the legacy handler silently.

## Parity is tested, not asserted

Each destination's `delivery.test.ts` drives **both** the new path and the existing handler over the same responses and compares per-job codes and errors. For `braze_audience` it additionally compares `stats.increment` call lists, so all three of its counters are held to parity too.

All five destinations set the delivery flag via `envOverrides` on their `dataDelivery` component tests, so **CI exercises the framework path** rather than dead code behind an off flag. `gaec`'s v0 scenarios get the flag too and are unaffected, which is itself the assertion that the shape guard works. `braze_audience`'s live tests run on it as well, via a new `envOverrides` hook on `LiveSpec`.

## Skills updated

`batching-framework`, `vdm-next-audience-integration` and `vdm-v2-object-destination` all described delivery as a separate `networkHandler.ts`, which would have sent the next integration down the path this PR replaces.

The contract lives in **one new skill**, `batching-framework-delivery` — verdict builders, `statusOverrides` resolution, indexing `perItem` off the posted request body rather than the response, the no-auth-inference rule, the delivery flag, and that transport stays in `networkHandler.ts` with the legacy handler retained until GA. The other three point at it and keep only what is specific to them: the single-item transport audit (`batching-framework`), identity-keyed failures (`vdm-next-audience-integration`), CustomerIO's 207 (`vdm-v2-object-destination`).

`live-integration-test` also documents `LiveSpec`, which gained `envOverrides` here, so it now covers naming a rollout flag — without one a live run exercises the legacy path.

## Bugs found while surveying (not fixed here)

- `iterable` / `iterable_audience` can emit `metadata: undefined` job states. rudder-server detects this as a non-fatal in/out breach and backfills the job to 500, so no data is lost — but the reported outcome is discarded and the event is redelivered. Migrating `iterable_audience` fixes it as a side effect, since the framework bounds-checks `perItem`.
- `TransformerProxyError`'s 6th argument is written and never read; ~50 handlers compute and pass it for nothing.
- Of the 20 v1 handlers that can produce a per-job response array, only `gaec` and `gaoc` set `statTags` on a returned response — the same hand-rolled block copied between them, both hardcoding `errorType: 'aborted'` and `status: 400` on a *partial* failure.

### DIFF
--- a/.claude/skills/batching-framework-delivery/SKILL.md
+++ b/.claude/skills/batching-framework-delivery/SKILL.md
@@ -101,6 +101,12 @@ So: if your destination's errors are worth reading, write the extractor in your 
 
 `gaec` originally indexed off the response's `results` and hit exactly this: whenever Google omitted `results`, an abort became a batch retry that could never converge, since re-uploading accepted adjustments returns duplicate-enhancement failures on every attempt.
 
+**`ctx.request.body?.JSON` only holds the batch on `BodyFormat.JSON`.** `mapSuccessPayloadToServerFormat` writes the batch to `body[strategy.bodyFormat]` and hard-sets the other three to `{}` (`processBatchedDestination.ts`), so a destination on `JSON_ARRAY`, `FORM` or `XML` reading `body.JSON` gets nothing on every response — success included — and the mismatch guard answers each one with an N-job 500. Read the key your `getBatchStrategy` actually returns. All six spec'd destinations are on `JSON` today, which is why no test catches this for you.
+
+**Decide what an unreadable posted array should mean before you write the loop.** `perItem([])` is a length mismatch, so the bridge retries the whole batch — and that retry reposts the same body and mismatches again, which never converges. `braze_audience` avoids it by falling back to `ctx.jobs`, which the framework builds 1:1 with the posted array. Taking the `fallback` parameter and returning a whole-batch verdict is the other option.
+
+**A delivery spec and an array-returning `transformEvent` are mutually exclusive.** `ctx.jobs` carries one entry per job, so a job contributing two body items puts the two lengths permanently out of step and the batch retries forever. `batchDestination.ts` and the VDM V2 dispatch table both permit `transformEvent` to return an array; a destination that does must not call `perItem`.
+
 ### Never infer auth from a status code
 
 A 401 from an API-key destination is a bad key, not a refreshable token, and guessing costs a real control-plane token refresh. Derive the category from the **response body** — `gaec` does, via `getAuthErrCategory` — or leave it alone. `getAuthErrCategoryFromStCode`'s unconditional 401→`REFRESH_TOKEN` is the anti-pattern this replaces.
--- a/.claude/skills/batching-framework-delivery/SKILL.md
+++ b/.claude/skills/batching-framework-delivery/SKILL.md
@@ -1,0 +1,144 @@
+---
+name: batching-framework-delivery
+description: Delivery (response handling) for batching-framework destinations. Declare a static delivery spec on the BatchDestination class instead of writing a networkHandler responseHandler — the framework derives status codes, statTags and the response shape.
+---
+
+# Batching Framework — Delivery
+
+**Objective:** Handle a destination's delivery response by declaring what should happen to the batch, and let the framework turn that into the delivery API response. Integrations never build a `DeliveryV1Response`, choose a status code, or throw.
+
+For the router-transform half of the framework, see `.claude/skills/batching-framework/SKILL.md`.
+
+## When you need this
+
+**Most destinations need nothing.** The framework default reproduces `genericNetworkHandler`: 2xx is a success, 429 is throttled, other retryable statuses retry, everything else aborts. `posthog` and `custom_audience` declare no overrides at all.
+
+Add a `delivery.ts` only when the destination's response handling genuinely differs:
+
+- a **partial-failure body** — some records rejected inside an otherwise-successful response
+- a **2xx that isn't a success** — the failure is in the body, not the status
+- **identity-keyed failures** — the response names *which* records failed rather than indexing them
+- a **real auth signal** in the body that should drive token refresh
+
+## Reference
+
+- `src/services/destination/nativeBatching/delivery.ts` — verdicts, `DeliverySpec`, the bridge
+- `src/v0/destinations/customerio/v2/delivery.ts` — 207 multi-status, positional index
+- `src/v0/destinations/braze_audience/delivery.ts` — partial failure on a 2xx, positional index
+- `src/v0/destinations/iterable_audience/delivery.ts` — failures keyed by identity, not index
+- `src/v0/destinations/google_adwords_enhanced_conversions/delivery.ts` — partial failure + body-derived OAuth
+- `docs/superpowers/specs/2026-07-30-network-handler-abstraction-design.md` — design, evidence, per-destination parity tables
+
+## Declaring the delivery spec
+
+Everything an integration says about delivery lives in **one object**, exported from its
+`delivery.ts` and attached to the class as `static readonly delivery`. Grouping it is the point: a
+`BatchDestination`'s other members are all about transforming an event, and a bare `statusOverrides`
+or `failureReason` beside them reads as more of the same.
+
+```typescript
+// delivery.ts
+import {
+  abort, perItem, success, retry, throttled, authExpired, authRevoked,
+  type DeliverySpec, type StatusOverrideMap,
+} from '../../../services/destination/nativeBatching/batchDestination';
+
+const myStatusOverrides: StatusOverrideMap = {
+  207: (ctx) => perItem(items.map((item, i) => (failed(i) ? abort(reason) : success()))),
+  '4xx': (ctx, fallback) => (isAuthBody(ctx.response) ? authRevoked('grant gone') : fallback()),
+};
+
+export const myDelivery: DeliverySpec = {
+  statusOverrides: myStatusOverrides,
+  // Optional: the reason on a failure verdict. Default is status-only and never reads the body.
+  failureReason: (ctx) => extractMyErrorMessage(ctx.response),
+};
+
+// routerTransform.ts
+class MyIntegration extends BatchDestination<TBody> {
+  static readonly delivery = myDelivery;
+}
+```
+
+- Both members are **optional**. A destination needing neither declares no `delivery` at all.
+- Status keys are an **exact status** or a class (`'2xx'` / `'4xx'` / `'5xx'`); exact wins.
+- The spec is **resolved down the prototype chain**: `statusOverrides` merges key-by-key (child
+  winning) and `failureReason` takes the nearest declaration. Declaring a spec in a subclass
+  therefore does not drop what a parent declared — static properties are shadowed, not merged, and
+  that failure would otherwise be invisible.
+- `fallback()` gives you the framework's own classification. Call it for the responses your override
+  does not own, rather than re-deriving a verdict — the return type stays total, with no sentinel.
+- Applying the spec is **framework-owned**: `handleDeliveryResponse(Class, ctx)` is a free function
+  in `delivery.ts`, so there is nothing to override and no `super` to call.
+
+`DeliveryContext` carries `status`, `response` (parsed body), `jobs` (one `ProxyMetdata` per job, in order), `request` (what was sent) and `destinationConfig`.
+
+## Error messages belong to the integration
+
+`failureReason`'s default is status-only — `[Generic Response Handler] Request failed with status: <n>` — and **never inspects the response body**. That mirrors `genericNetworkHandler`, the fallback every unmigrated destination already gets, which likewise leaves the body to travel in `destinationResponse`.
+
+Do not add body-parsing to the framework. There is no shared convention to generalise: on the legacy path every destination extracts its message itself, against the shape its own API returns (`buildErrorMessage` joining `reason`/`field`/`message` for customerio, `error.message` for gaec, `params ?? msg ?? message` for iterable_audience). A generic "look for `message`, then `msg`, then `error`" helper looks like it unifies them but does not — it just imposes one destination's field order on every other, and any quirk added for one leaks into all of them.
+
+So: if your destination's errors are worth reading, write the extractor in your own `delivery.ts` and point `delivery.failureReason` at it. Return the string **bare** — the per-job `error` is what live events display and what error reporting groups on, so `Invalid API key` beats `"Invalid API key"`.
+
+## Verdicts
+
+| Builder | Meaning | Result |
+| --- | --- | --- |
+| `success()` | delivered | per-job `200` |
+| `abort(reason)` | permanent failure | per-job `400` |
+| `retry(reason, { dontBatch })` | transient failure | per-job `500` |
+| `throttled(reason)` | rate limited | per-job `429` |
+| `authExpired(reason)` | token stale but recoverable | `REFRESH_TOKEN` → refresh + retry |
+| `authRevoked(reason)` | grant gone | `AUTH_STATUS_INACTIVE` → abort |
+| `perItem([...])` | one verdict per request-body item | per-job, positionally |
+
+`ItemVerdict` (what `perItem` accepts) deliberately excludes the two auth refinements: rudder-server overwrites the status code of *every* job in a batch when an `authErrorCategory` is present, so a per-item auth verdict cannot be represented.
+
+### `perItem` is positional and 1:1
+
+**Index it off the posted request body (`ctx.request.body?.JSON`), not off the response.** The posted array is the one guaranteed to be the same length as the job list; a response array can be truncated or omitted entirely. A length mismatch makes the framework retry the whole batch rather than misattribute — correct, but a worse outcome than reading the body you sent.
+
+`gaec` originally indexed off the response's `results` and hit exactly this: whenever Google omitted `results`, an abort became a batch retry that could never converge, since re-uploading accepted adjustments returns duplicate-enhancement failures on every attempt.
+
+### Never infer auth from a status code
+
+A 401 from an API-key destination is a bad key, not a refreshable token, and guessing costs a real control-plane token refresh. Derive the category from the **response body** — `gaec` does, via `getAuthErrCategory` — or leave it alone. `getAuthErrCategoryFromStCode`'s unconditional 401→`REFRESH_TOKEN` is the anti-pattern this replaces.
+
+## What the framework derives
+
+Don't set these yourself:
+
+- **per-job `statusCode`** — from the verdict kind
+- **top-level `status`** — `ctx.status`, passed through
+- **`message`** — the shared failure reason when every failure agrees on one, otherwise a count
+- **`statTags`** — the whole tag set, when every job failed the same way
+- **`authErrorCategory`** — from the auth refinements only
+
+## Transport stays in `networkHandler.ts`
+
+Only *response handling* moves. A destination that builds its HTTP request at delivery time (SDK call, URL derived from `params`, custom `processAxiosResponse`) keeps `proxy` / `prepareProxy` / `processAxiosResponse` in its handler.
+
+**Keep the legacy `networkHandler.ts` until GA.** Workspaces not yet enrolled still transform through `processRouterDest` and are delivered by that handler; both are deleted together at GA.
+
+## Enabling it
+
+`{DEST_NAME_UPPER}_BATCHING_FRAMEWORK_DELIVERY_ENABLED_WORKSPACE_IDS` — comma-separated workspace IDs or `ALL`. Two things to know:
+
+- There is **no GA map** for it — being in `batchedDestinationsMap` does not enable framework delivery. It stays on the legacy handler everywhere until a workspace is named.
+- It **requires the transform flag** for the same workspace and returns false without it. The delivery path interprets a payload built by the matching transform path; an unenrolled workspace's events are still built by `processRouterDest`, and pairing those with framework response handling would misread them.
+
+Set both in your `dataDelivery` component tests via `envOverrides` (and in `live.ts` via the `envOverrides` field on `LiveSpec`), or CI keeps exercising the legacy handler.
+
+## Testing
+
+Drive **both** paths over the same responses and compare, rather than asserting the new one in isolation — that is what makes it a parity test:
+
+```typescript
+const viaFramework = (ctx) =>
+  toDeliveryV1Response(handleDeliveryResponse(Integration, ctx), ctx, DEST);
+const viaLegacy = (ctx) => legacyResponseHandler({ ... });
+// compare per-job statusCodes and errors for each response shape
+```
+
+**Reference:** `src/v0/destinations/braze_audience/delivery.test.ts` — also compares `stats.increment` call lists, so counters are held to parity too.
--- a/src/constants/batchedDestinationsMap.ts
+++ b/src/constants/batchedDestinationsMap.ts
@@ -4,18 +4,28 @@ import { getBatchingFrameworkGaDestinations } from '../features';
 // Once a destination is added here, it always uses the new path regardless of env var.
 export const batchedDestinationsMap: Record<string, true> = getBatchingFrameworkGaDestinations();
 
-// Per-destination env var: {DEST}_BATCHING_FRAMEWORK_ENABLED_WORKSPACE_IDS
+// Per-destination env var: {DEST}_{SUFFIX}
 // Values: comma-separated workspace IDs, or 'ALL' for all workspaces
 // Example: POSTHOG_BATCHING_FRAMEWORK_ENABLED_WORKSPACE_IDS="ws-1,ws-2" or "ALL"
 // If not set or empty → disabled for that destination (legacy path)
-const getEnabledWorkspaceIds = (destType: string): string[] => {
-  const envKey = `${destType.toUpperCase()}_BATCHING_FRAMEWORK_ENABLED_WORKSPACE_IDS`;
+const getEnabledWorkspaceIds = (destType: string, suffix: string): string[] => {
+  const envKey = `${destType.toUpperCase()}_${suffix}`;
   return (
     process.env[envKey]
       ?.split(',')
       ?.map((s) => s.trim())
       ?.filter((s) => s) ?? []
   );
+};
+
+const matchesWorkspace = (enabledWorkspaceIds: string[], workspaceId: string): boolean => {
+  if (enabledWorkspaceIds.length === 0) {
+    return false;
+  }
+  if (enabledWorkspaceIds.includes('ALL')) {
+    return true;
+  }
+  return enabledWorkspaceIds.includes(workspaceId?.trim());
 };
 
 // OR logic:
@@ -30,12 +40,38 @@ export const isBatchingFrameworkEnabled = (destType: string, workspaceId: string
   }
 
   // Pre-GA: check per-destination env var
-  const enabledWorkspaceIds = getEnabledWorkspaceIds(upperDestType);
-  if (enabledWorkspaceIds.length === 0) {
+  return matchesWorkspace(
+    getEnabledWorkspaceIds(upperDestType, 'BATCHING_FRAMEWORK_ENABLED_WORKSPACE_IDS'),
+    workspaceId,
+  );
+};
+
+/**
+ * Whether the batching framework also owns *delivery* (response handling) for this destination and
+ * workspace, rather than the destination's own networkHandler.
+ *
+ * Per-destination env var: {DEST}_BATCHING_FRAMEWORK_DELIVERY_ENABLED_WORKSPACE_IDS
+ * Values: comma-separated workspace IDs, or 'ALL'. Unset or empty → the legacy networkHandler.
+ *
+ * Two deliberate properties:
+ *
+ *  - **No GA map.** Unlike the transform flag there is no destination list that force-enables this,
+ *    so delivery stays on the legacy handler everywhere until a workspace is named explicitly.
+ *  - **Requires the transform flag.** The delivery path has to interpret a payload built by the
+ *    matching transform path — an unenrolled workspace's events are still built by the legacy
+ *    `processRouterDest`, and its response must be read by the legacy handler. Enabling delivery
+ *    for a workspace that is not on the batching-framework transform would pair the two halves
+ *    incorrectly, so this returns false regardless of the delivery env var.
+ */
+export const isBatchingFrameworkDeliveryEnabled = (
+  destType: string,
+  workspaceId: string,
+): boolean => {
+  if (!isBatchingFrameworkEnabled(destType, workspaceId)) {
     return false;
   }
-  if (enabledWorkspaceIds.includes('ALL')) {
-    return true;
-  }
-  return enabledWorkspaceIds.includes(workspaceId.trim());
+  return matchesWorkspace(
+    getEnabledWorkspaceIds(destType, 'BATCHING_FRAMEWORK_DELIVERY_ENABLED_WORKSPACE_IDS'),
+    workspaceId,
+  );
 };
--- a/src/constants/batchedDestinationsMap.ts
+++ b/src/constants/batchedDestinationsMap.ts
@@ -4,12 +4,15 @@ import { getBatchingFrameworkGaDestinations } from '../features';
 // Once a destination is added here, it always uses the new path regardless of env var.
 export const batchedDestinationsMap: Record<string, true> = getBatchingFrameworkGaDestinations();
 
-// Per-destination env var: {DEST}_{SUFFIX}
+// Per-destination env var: {DEST}_{FEATURE}_ENABLED_WORKSPACE_IDS
 // Values: comma-separated workspace IDs, or 'ALL' for all workspaces
 // Example: POSTHOG_BATCHING_FRAMEWORK_ENABLED_WORKSPACE_IDS="ws-1,ws-2" or "ALL"
 // If not set or empty → disabled for that destination (legacy path)
-const getEnabledWorkspaceIds = (destType: string, suffix: string): string[] => {
-  const envKey = `${destType.toUpperCase()}_${suffix}`;
+const getEnabledWorkspaceIds = (
+  destType: string,
+  feature: 'BATCHING_FRAMEWORK' | 'BATCHING_FRAMEWORK_DELIVERY',
+): string[] => {
+  const envKey = `${destType.toUpperCase()}_${feature}_ENABLED_WORKSPACE_IDS`;
   return (
     process.env[envKey]
       ?.split(',')
@@ -40,10 +43,7 @@ export const isBatchingFrameworkEnabled = (destType: string, workspaceId: string
   }
 
   // Pre-GA: check per-destination env var
-  return matchesWorkspace(
-    getEnabledWorkspaceIds(upperDestType, 'BATCHING_FRAMEWORK_ENABLED_WORKSPACE_IDS'),
-    workspaceId,
-  );
+  return matchesWorkspace(getEnabledWorkspaceIds(upperDestType, 'BATCHING_FRAMEWORK'), workspaceId);
 };
 
 /**
@@ -71,7 +71,7 @@ export const isBatchingFrameworkDeliveryEnabled = (
     return false;
   }
   return matchesWorkspace(
-    getEnabledWorkspaceIds(destType, 'BATCHING_FRAMEWORK_DELIVERY_ENABLED_WORKSPACE_IDS'),
+    getEnabledWorkspaceIds(destType, 'BATCHING_FRAMEWORK_DELIVERY'),
     workspaceId,
   );
 };
--- a/src/services/destination/nativeBatching/delivery.ts
+++ b/src/services/destination/nativeBatching/delivery.ts
@@ -10,6 +10,7 @@ import { TransformerProxyError } from '../../../v0/util/errorTypes';
 import { isHttpStatusSuccess, isHttpStatusRetryable } from '../../../v0/util';
 import tags from '../../../v0/util/tags';
 import stats from '../../../util/stats';
+import logger from '../../../logger';
 import {
   REFRESH_TOKEN,
   AUTH_STATUS_INACTIVE,
@@ -41,10 +42,16 @@ export type Verdict = SuccessVerdict | AbortVerdict | RetryVerdict;
  * The subset of verdicts expressible per item. The auth refinements are absent deliberately:
  * rudder-server overwrites the status code of *every* job in a batch whenever an
  * `authErrorCategory` is present, so a per-item auth verdict cannot be represented.
+ *
+ * Both refinements are excluded by construction, not by omission. `as?: 'throttled'` already makes
+ * `authExpired()` a type error, but a plain `{ kind: 'abort'; reason: string }` would *accept*
+ * `authRevoked()` — a returned value gets no excess-property check, so the extra `auth: 'revoked'`
+ * would be silently dropped and the record would abort as an ordinary 400. `auth?: never` closes
+ * that, so both builders fail at the call site rather than one of them degrading quietly.
  */
 export type ItemVerdict =
   | SuccessVerdict
-  | { kind: 'abort'; reason: string }
+  | { kind: 'abort'; reason: string; auth?: never }
   | { kind: 'retry'; reason: string; as?: 'throttled'; dontBatch?: boolean };
 
 export const success = (): SuccessVerdict => ({ kind: 'success' });
@@ -93,6 +100,13 @@ export type PerItemVerdicts = { kind: 'perItem'; verdicts: ItemVerdict[] };
  * One verdict per **request body item**, in the order they were sent. Positional and 1:1 — the
  * framework carries no item→job map, so this is only correct when each job contributed exactly
  * one body item. A length mismatch degrades to a whole-batch verdict rather than misattributing.
+ *
+ * That makes a delivery spec and an **array-returning `transformEvent`** mutually exclusive.
+ * `ctx.jobs` is one entry per job (`processBatchedDestination` builds it from a `Set<number>` of
+ * job ids), so one job contributing two body items puts the two lengths permanently out of step:
+ * the guard retries the batch, the retry reproduces the same mismatch, and it never converges.
+ * `batchDestination.ts`'s `transformEvent` and the VDM V2 dispatch table both permit an array, so
+ * a destination that returns one must not declare `statusOverrides` that call `perItem`.
  */
 export const perItem = (verdicts: ItemVerdict[]): PerItemVerdicts => ({
   kind: 'perItem',
@@ -174,12 +188,20 @@ export const statusClassOf = (status: number): StatusKey | undefined => {
 export const defaultFailureReason = (status: number): string =>
   `[Generic Response Handler] Request failed with status: ${status}`;
 
-/** The framework's own status classification. Mirrors genericNetworkHandler's responseHandler. */
-export const classifyByStatus = (status: number, reason: string): Verdict => {
+/**
+ * The framework's own status classification. Mirrors genericNetworkHandler's responseHandler.
+ *
+ * `reason` is a thunk so the success path costs nothing. `fallback()` is called for every response
+ * with no override — including every plain 2xx — and an eagerly-evaluated reason would run the
+ * destination's extractor over the whole body just to discard the result. `braze_audience`'s falls
+ * through to `JSON.stringify(response)` when there is no `message`, which on a large batch response
+ * is the expensive kind of nothing.
+ */
+export const classifyByStatus = (status: number, reason: () => string): Verdict => {
   if (isHttpStatusSuccess(status)) return success();
-  if (status === 429) return throttled(reason);
-  if (isHttpStatusRetryable(status)) return retry(reason);
-  return abort(reason);
+  if (status === 429) return throttled(reason());
+  if (isHttpStatusRetryable(status)) return retry(reason());
+  return abort(reason());
 };
 
 // ---------------------------------------------------------------------------
@@ -271,7 +293,7 @@ export function handleDeliveryResponse(klass: unknown, ctx: DeliveryContext): Ha
   const statusClass = statusClassOf(ctx.status);
   const override =
     statusOverrides[ctx.status] ?? (statusClass ? statusOverrides[statusClass] : undefined);
-  const fallback = () => classifyByStatus(ctx.status, failureReason(ctx));
+  const fallback = () => classifyByStatus(ctx.status, () => failureReason(ctx));
   return override ? override(ctx, fallback) : fallback();
 }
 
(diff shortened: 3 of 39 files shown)


src/constants/batchedDestinationsMap.ts
```diff
@@ -4,12 +4,12 @@ import { getBatchingFrameworkGaDestinations } from '../features';
...
 export const batchedDestinationsMap: Record<string, true> = getBatchingFrameworkGaDestinations();
 
-// Per-destination env var: {DEST}_BATCHING_FRAMEWORK_ENABLED_WORKSPACE_IDS
+// Per-destination env var: {DEST}_{SUFFIX}
 // Values: comma-separated workspace IDs, or 'ALL' for all workspaces
 // Example: POSTHOG_BATCHING_FRAMEWORK_ENABLED_WORKSPACE_IDS="ws-1,ws-2" or "ALL"
 // If not set or empty → disabled for that destination (legacy path)
-const getEnabledWorkspaceIds = (destType: string): string[] => {
-  const envKey = `${destType.toUpperCase()}_BATCHING_FRAMEWORK_ENABLED_WORKSPACE_IDS`;
+const getEnabledWorkspaceIds = (destType: string, suffix: string): string[] => {
```
saikumarrs: **nit:** `suffix` is a raw env-key fragment, so both call sites repeat `_ENABLED_WORKSPACE_IDS` and a typo in either just reads an unset key — the feature silently reads as off, with no type or runtime error. Let's take the feature and keep the convention in one place:

```ts
const getEnabledWorkspaceIds = (
  destType: string,
  feature: 'BATCHING_FRAMEWORK' | 'BATCHING_FRAMEWORK_DELIVERY',
): string[] => {
  const envKey = `${destType.toUpperCase()}_${feature}_ENABLED_WORKSPACE_IDS`;
```

Same for the comment on L7 — `{DEST}_{SUFFIX}` documents less than the literal it replaced.

src/v0/destinations/braze_audience/delivery.ts
```diff
@@ -0,0 +1,127 @@
...
+};
+
+const brazeAudienceStatusOverrides: StatusOverrideMap = {
+  '2xx': (ctx, fallback) => {
+    const rawErrors = (ctx.response as { errors?: unknown } | undefined)?.errors;
+    const errors: BrazeAudienceError[] = Array.isArray(rawErrors) ? rawErrors : [];
+    if (errors.length === 0) return fallback();
+
+    const destinationId = ctx.jobs[0]?.destinationId ?? '';
+    const workspaceId = ctx.jobs[0]?.workspaceId ?? '';
```
saikumarrs: These two lines already exist three times — here, `iterable_audience/delivery.ts:104-105`, and the framework's own `nativeBatching/delivery.ts:312-313`. Let's put `destinationId` and `workspaceId` on `DeliveryContext` and have `handleDeliveryResponse` fill them from `ctx.jobs[0]` once, so a spec that needs them for stats isn't reaching into the job list to rebuild what the framework already computed.

src/v0/destinations/google_adwords_enhanced_conversions/delivery.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export const extractGaecErrorMessage = (response: unknown): string =>
+  (response as { error?: { message?: string } })?.error?.message || 'unknown error';
+
+const gaecStatusOverrides: StatusOverrideMap = {
+  '2xx': (ctx, fallback) => {
+    const body = ctx.response as
+      | { partialFailureError?: { code?: number; message?: string }; results?: unknown[] }
+      | undefined;
+    const partialFailureError = body?.partialFailureError;
```
saikumarrs, line 41: Test gap: none of the eight `parityCases` sends a non-object 2xx body. The v0 handler guards this with `isPartialFailureBody` because "a successful 200 may also have an empty/undefined body" (`v0/destinations/google_adwords_enhanced_conversions/networkHandler.ts:204`); here that job falls entirely to `body?.partialFailureError`. It holds — a string or array body has no such key — but nothing pins it. Add `''` and `null` response cases.

src/services/destination/nativeBatching/delivery.ts
```diff
@@ -0,0 +1,442 @@
...
+ *
+ * Static properties are inherited but *shadowed*, not merged: a subclass declaring its own
+ * `delivery` would otherwise silently drop everything an ancestor declared, with no type or
+ * runtime error. `statusOverrides` is therefore merged key-by-key, so a family-level map on e.g.
+ * VDMV2ObjectDestination keeps working when a concrete destination adds an entry of its own, and
+ * `failureReason` resolves to the nearest declaration — which is what a plain static override
+ * would have done. Not cached: the walk is three levels of small objects, nothing next to the HTTP
+ * call it accompanies, and a cache would go stale whenever a test swaps a spec.
+ */
+export function resolveDeliverySpec(klass: unknown): ResolvedDeliverySpec {
```
saikumarrs, line 267: `resolveDeliverySpec`'s prototype-chain merge has no production caller. All four declarations are leaf classes — `gaec`, `braze_audience`, `customerio` and `iterable_audience` `routerTransform.ts` — and no family class declares `delivery`, so the walk is only ever reached from `delivery.test.ts`. That is a merge rule with no counterpart elsewhere on `BatchDestination`, carried for a shadowing hazard nothing can hit yet. [Non-blocking] — it only starts paying once the VDM family specs exist.

src/services/destination/nativeBatching/delivery.ts
```diff
@@ -0,0 +1,442 @@
...
+import tags from '../../../v0/util/tags';
+import stats from '../../../util/stats';
+import type {
+  DeliveryJobState,
+  DeliveryV1Response,
+  ProxyMetdata,
+  ProxyV1Request,
+} from '../../../types';
+
+const {
```
maheshkutty: nit: you can use import instead require for consistancy

src/services/destination/nativeBatching/delivery.ts
```diff
@@ -0,0 +1,442 @@
...
+
+  return {
+    // The destination's own status, always. `ProxyResponseV1` has no `status` field
+    // (`router/transformer/transformer_proxy_adapter.go:40-44`), so rudder-server takes disposition
+    // from each job state and never reads this; substituting 207 for a status the destination did
+    // not send would invent a difference from the legacy handler that nothing benefits from.
+    status: ctx.status,
+    message,
+    ...statTags,
+    response,
```
maheshkutty, line 512: As part of this pr we are getting rid of destinationResponse, that we get from destination?

src/constants/batchedDestinationsMap.ts
shekhar-rudder: Done — `getEnabledWorkspaceIds` now takes a typed `'BATCHING_FRAMEWORK' | 'BATCHING_FRAMEWORK_DELIVERY'` feature and builds `{DEST}_{FEATURE}_ENABLED_WORKSPACE_IDS` itself, so both call sites can't drift on the suffix. Comments on L7 updated to match.

src/v0/destinations/braze_audience/delivery.ts
shekhar-rudder: Done — destinationId/workspaceId are now required fields on DeliveryContext, derived once by a firstJobIdentity(jobs) helper at the single production construction site (nativeIntegration.ts) and by each test's ctxFor. braze_audience, iterable_audience and the framework's own mismatch-stat call in nativeBatching/delivery.ts all just read ctx.destinationId/ctx.workspaceId now instead of re-deriving from jobs[0].

src/v0/destinations/google_adwords_enhanced_conversions/delivery.ts
shekhar-rudder: Added both — '2xx, empty string body' and '2xx, null body' cases to parityCases, asserting the framework and the legacy handler agree (both read as plain success, since `body?.partialFailureError` is undefined on a non-object body either way).

src/services/destination/nativeBatching/delivery.ts
shekhar-rudder: Agreed this is real but out of scope for this pass since you flagged it non-blocking — leaving resolveDeliverySpec's merge as-is for now rather than pulling out the unreached branch. Happy to revisit once a VDM family class actually declares delivery, or sooner if you'd rather I simplify it now.

src/services/destination/nativeBatching/delivery.ts
shekhar-rudder: Done — swapped the require() for a named ES import of authConstants (there's already precedent for that in cdk/v2/destinations/zoho/transformRecordV2.ts).

src/services/destination/nativeBatching/delivery.ts
shekhar-rudder: Yes, deliberately — `destinationResponse` is echoed on failures (it rides the `TransformerProxyError` throw, `{status: ctx.status, response: ctx.response}`, same as today) but omitted on a returned success, per the design doc's §3.7 (docs/superpowers/specs/2026-07-30-network-handler-abstraction-design.md:431-433).

Checked it against rudder-server before deciding: `destinationResponse` on the v1 proxy response has exactly one consumer, `enterprise/reporting/error_extractor.go:326-327`, and that path is failure-gated (`shouldReport` only admits `StatusDetail.StatusCode >= 400`, `error_reporting.go:246`). Nothing reads it off a success response, so dropping it there is a no-op behaviorally.

It is a real response-shape change for two of the five migrated destinations though — `customerio`'s legacy success path already omits it, but `iterable_audience`'s currently includes it (`v1/destinations/iterable_audience/strategies/audience-list.ts:113-118`) and `braze_audience`'s does too (`v1/destinations/braze_audience/networkHandler.ts:120`). Both are called out explicitly in the parity tables (design doc lines 711-713, 743-745) as a known, accepted diff — gated behind the per-workspace delivery flag, so nothing changes until a workspace opts in.

src/services/destination/nativeBatching/delivery.ts
shekhar-rudder: Circling back on this — gave VDMV2ObjectDestination its own `static readonly delivery: DeliverySpec = {}` (deliberately empty, no behaviour change). resolveDeliverySpec's walk for `CustomerIOIntegration` now finds a real family-level declaration between the leaf and `BatchDestination`'s root default, instead of jumping straight from leaf to root. Doesn't change any merge output today since it's empty, but it's a real site for a future VDM v2 object destination to actually share something at, and it's a genuine class in the walk rather than a hypothetical.

src/services/destination/nativeBatching/delivery.ts
```diff
@@ -0,0 +1,456 @@
...
+ * string and leaves the body to travel separately in `destinationResponse`
+ * (`adapters/networkhandler/genericNetworkHandler.js`). There is no shared body-parsing
+ * convention on the legacy path to inherit: every destination that wants a message out of its
+ * body writes that itself, against the shape its own API returns.
+ *
+ * So the framework stays out of it, and an integration that wants better reads its own body in its
+ * own `delivery.failureReason`.
+ */
+export const defaultFailureReason = (status: number): string =>
+  `[Generic Response Handler] Request failed with status: ${status}`;
```
maheshkutty, line 190: Do need to add prefix as part of error message "[Generic Response Handler]", this prefix looks like technical message which is internal to us

shekhar-rudder: This isn't new — it's copied verbatim from the existing `genericNetworkHandler.js:27`, which is already the live fallback for every destination without its own network handler today. Customers already see this exact string for delivery failures; this default just preserves parity with that (called out explicitly in the design doc, §2.3 around `docs/superpowers/specs/2026-07-30-network-handler-abstraction-design.md:230`).

It only fires on the unclassified fallback path too — all four destinations this PR migrates (braze_audience, iterable_audience, gaec, customerio) already supply their own `failureReason` with real API error text, so this string only shows up when nothing better exists, same as today.

saikumarrs (commented):
Went through the delivery bridge and the four destination specs against the legacy handlers they replace.

Two correctness items: A1 — `gaec` still falls back to indexing off the response, which this PR's own `SKILL.md:100-102` calls out and names `gaec` for; and A2 — `ItemVerdict` excludes `authExpired` but not `authRevoked`, so `perItem([authRevoked(...)])` compiles and drops the category. Both have suggestions inline.

Three heads-ups are latent rather than live — nothing reaches them with the current override set, but the framework invites them: `statTags` on a mixed non-2xx, the `ctx.request.body?.JSON` contract vs `bodyFormat`, and positional `perItem` vs an array-returning `transformEvent`.

Two are questions rather than claims: the missing `fallback()` in `iterable_audience`, and whether skipping `ErrorReportingService.reportError` on the returned path was deliberate given you matched the statTags parity right above it.

Rest are nits and one test gap.

src/v0/destinations/google_adwords_enhanced_conversions/delivery.ts
```diff
@@ -0,0 +1,83 @@
...
+    // guaranteed to be the same length as the job list — which keeps the bridge's attribution
+    // guard out of the picture entirely.
+    //
+    // Reading a missing `results` entry as failed is deliberate and reproduces the legacy handler's
+    // `results?.[i] ?? {}`. Google omits or truncates `results` on some partial failures, and the
+    // alternative — losing attribution and retrying the batch — would re-upload adjustments Google
+    // has already accepted, which come back as duplicate-enhancement failures on every attempt.
+    const items = (ctx.request.body?.JSON as { conversionAdjustments?: unknown[] })
+      ?.conversionAdjustments;
+    const positions = Array.isArray(items) ? items : results;
```
saikumarrs, line 61: The fallback indexes off the response. `.claude/skills/batching-framework-delivery/SKILL.md:100-102` — added in this PR — says to index off the posted body, and names `gaec` as the destination that hit exactly this. The comment right above gives the reason itself: the posted array is the one guaranteed to be the same length as the job list. `results` carries no such guarantee — Google truncates it on partial failures.

```suggestion
    const positions = Array.isArray(items) ? items : ctx.jobs;
```

Unreachable today — `networkHandler.ts:164` dereferences `conversionAdjustments!` before delivery runs — so it's cheap to correct now. `braze_audience/delivery.ts:91` already falls back to `jobs`.

src/services/destination/nativeBatching/delivery.ts
```diff
@@ -0,0 +1,456 @@
...
+export type Verdict = SuccessVerdict | AbortVerdict | RetryVerdict;
+
+/**
+ * The subset of verdicts expressible per item. The auth refinements are absent deliberately:
+ * rudder-server overwrites the status code of *every* job in a batch whenever an
+ * `authErrorCategory` is present, so a per-item auth verdict cannot be represented.
+ */
+export type ItemVerdict =
+  | SuccessVerdict
+  | { kind: 'abort'; reason: string }
```
saikumarrs: `ItemVerdict` only excludes one of the two auth refinements the comment above (and `SKILL.md:96`) says are absent. `perItem([authExpired('x')])` is a type error — `as: 'authExpired'` isn't assignable to `as?: 'throttled'`. `perItem([authRevoked('x')])` compiles: `{kind:'abort'; reason:string; auth:'revoked'}` is assignable to `{kind:'abort'; reason:string}`, and there's no excess-property check on a returned value. The `auth: 'revoked'` is then dropped and the record aborts as a plain 400.

```suggestion
  | { kind: 'abort'; reason: string; auth?: never }
```

Checked with `tsc` both ways — that makes `authRevoked` fail at the call site and leaves `abort()` untouched.

src/services/destination/nativeIntegration.ts
```diff
@@ -214,6 +233,61 @@ export class NativeIntegrationDestinationService implements DestinationService {
...
+          frameworkResponse.statTags = {
+            ...frameworkResponse.statTags,
+            ...deliveryMetaTO.errorDetails,
+          };
+        }
+        return frameworkResponse;
+      }
+
       let rudderJobMetadata =
         version.toLowerCase() === 'v1'
```
saikumarrs, line 309: The catch at line 335 still branches on `version.toLowerCase() === 'v1'` while the framework branch above narrows on `isProxyV1Request` — and its comment argues that the array-ness of `metadata` is the honest thing to narrow on. A request satisfying one and not the other returns a `DeliveryV1Response` when the destination succeeds and a v0 body when it fails. Let's use the same predicate in both places.

(Commenting here since 335 isn't in the diff.)

src/services/destination/nativeBatching/delivery.ts
```diff
@@ -0,0 +1,456 @@
...
+  // only honest when the response as a whole is one failure: every job failed, the same way. A
+  // partially-succeeded batch has no single errorType, and labelling one 'aborted' — as gaec's
+  // legacy handler did — counts a batch that partly delivered as a whole-batch abort.
+  //
+  // Only the error-describing half is set here, matching what the throw path hands
+  // `TransformerProxyError`. The identifying half — destType, destinationId, workspaceId, module,
+  // implementation, feature — is merged in by the caller from the same `getTags` metadata that
+  // `postTransformation` merges for a thrown error, so both paths emit one tag set from one source.
+  const allFailed = failures.length === verdicts.length && failures.length > 0;
+  const statTags =
```
saikumarrs, line 486: Heads-up: `statTags` is gated on `uniform && allFailed`, so a part-failed per-item list on a non-2xx returns a non-2xx `status` with no `statTags` — which `validateStatTags` rejects (`zodTypes.ts:169-174,220`). That's the mixpanel `/import`-on-400 case the exclusion comment at 362-364 names as the reason this path exists.

Nothing reaches it today — every current override is 2xx-keyed except gaec's `4xx`, which returns whole-batch verdicts — so the first destination that does fails its component test rather than shipping.

.claude/skills/batching-framework-delivery/SKILL.md
```diff
@@ -0,0 +1,144 @@
...
+| `throttled(reason)` | rate limited | per-job `429` |
+| `authExpired(reason)` | token stale but recoverable | `REFRESH_TOKEN` → refresh + retry |
+| `authRevoked(reason)` | grant gone | `AUTH_STATUS_INACTIVE` → abort |
+| `perItem([...])` | one verdict per request-body item | per-job, positionally |
+
+`ItemVerdict` (what `perItem` accepts) deliberately excludes the two auth refinements: rudder-server overwrites the status code of *every* job in a batch when an `authErrorCategory` is present, so a per-item auth verdict cannot be represented.
+
+### `perItem` is positional and 1:1
+
+**Index it off the posted request body (`ctx.request.body?.JSON`), not off the response.** The posted array is the one guaranteed to be the same length as the job list; a response array can be truncated or omitted entirely. A length mismatch makes the framework retry the whole batch rather than misattribute — correct, but a worse outcome than reading the body you sent.
```
saikumarrs, line 100: Heads-up: this holds only for `BodyFormat.JSON`. `mapSuccessPayloadToServerFormat` writes the batch to `body[strategy.bodyFormat]` and hard-sets the other three to `{}` (`processBatchedDestination.ts:164-167`), so a destination on `JSON_ARRAY` or `FORM` that follows this rule gets `items = []` on every response — success included — and an N-job 500 retry, with only the mismatch counter to show for it.

All six destinations default to JSON today so nothing breaks. Worth either naming the constraint here or giving `ctx` a `bodyFormat`-aware accessor.

src/services/destination/nativeBatching/delivery.ts
```diff
@@ -0,0 +1,456 @@
...
+): { kind: 'abort'; reason: string; auth: 'revoked' } => ({
+  kind: 'abort',
+  reason,
+  auth: 'revoked',
+});
+
+export type PerItemVerdicts = { kind: 'perItem'; verdicts: ItemVerdict[] };
+
+/**
+ * One verdict per **request body item**, in the order they were sent. Positional and 1:1 — the
```
saikumarrs, line 101: Heads-up: worth stating here that a delivery spec and an array-returning `transformEvent` are mutually exclusive. `perItem` is positional against the posted body while `ctx.jobs` comes from a `Set<number>` (`types.ts:52`), so one job contributing two items puts the two lengths permanently out of step — the guard retries the batch, the retry reproduces the mismatch, and it never converges.

All four spec'd destinations return a single event today, but `batchDestination.ts:105` and the VDM V2 dispatch table both permit an array.

src/v0/destinations/iterable_audience/delivery.ts
```diff
@@ -0,0 +1,130 @@
...
+  const body = response as { params?: unknown; msg?: unknown; message?: unknown } | undefined;
+  // The trailing `?? response` is the framework default the legacy handler lacked: a body with
+  // none of these fields is shown whole rather than replaced by a placeholder that says nothing.
+  const message = body?.params ?? body?.msg ?? body?.message ?? response;
+  if (typeof message === 'string') return message || 'unknown error format';
+  return JSON.stringify(message) ?? 'unknown error format';
+};
+
+const iterableAudienceStatusOverrides: StatusOverrideMap = {
+  '2xx': (ctx) => {
```
saikumarrs, line 74: `braze_audience` and `gaec` both take `fallback` and return it when the array they'd index is empty; this override doesn't take the parameter at all, so `subscribers ?? []` on line 78 goes straight to `perItem([])` and the mismatch guard turns a batch Iterable accepted into an N-job 500.

Is `subscribers` guaranteed present on every 2xx that reaches here, or should this degrade the same way? `customerio/v2/delivery.ts:32` has the same `?? []` shape.

src/services/destination/nativeIntegration.ts
```diff
@@ -214,6 +233,61 @@ export class NativeIntegrationDestinationService implements DestinationService {
...
+            ctx.workspaceId,
+            tags.FEATURES.DATA_DELIVERY,
+          );
+          frameworkResponse.statTags = {
+            ...frameworkResponse.statTags,
+            ...deliveryMetaTO.errorDetails,
+          };
+        }
+        return frameworkResponse;
+      }
```
saikumarrs, line 306: This returns without going through `DestinationPostTransformationService`, so `ErrorReportingService.reportError` (`postTransformation.ts:230`) never runs for a returned failure — attribution-lost retries, `dontBatch` retries, and any future non-2xx per-item failure.

The block just above deliberately merges `getTags` so `integration.failure_detailed` matches the thrown path. Was leaving `reportError` out of that same parity deliberate?

test/integrations/destinations/google_adwords_enhanced_conversions/dataDelivery/data.ts
```diff
@@ -1,9 +1,29 @@
...
+ * handler and continue to assert its behaviour.
+ */
+const batchingFrameworkDelivery = {
+  GOOGLE_ADWORDS_ENHANCED_CONVERSIONS_BATCHING_FRAMEWORK_ENABLED_WORKSPACE_IDS: 'ALL',
+  GOOGLE_ADWORDS_ENHANCED_CONVERSIONS_BATCHING_FRAMEWORK_DELIVERY_ENABLED_WORKSPACE_IDS: 'ALL',
+};
 
 export const data = [
   ...v0oauthScenarios,
   ...v1oauthScenarios,
```
saikumarrs: Test gap: this enrols `gaec_v1_oauth_scenario_1/2` in the delivery flag, but both still assert the legacy handler's `'... during Google_adwords_enhanced_conversions response transformation'` and `errorType: 'aborted'` (`oauth.ts:211,214,264,267`). They pass because `gaecProxyRequest` throws in `getConversionActionId` (`networkHandler.ts:156`) before the delivery branch runs — so nothing here covers the framework's 401/403 path.

Either drop them from the `.map` or add a case that reaches `toDeliveryV1Response`.

test/integrations/destinations/google_adwords_enhanced_conversions/dataDelivery/data.ts
```diff
@@ -1,9 +1,29 @@
 import { v0oauthScenarios, v1oauthScenarios } from './oauth';
 import { testScenariosForV0API, testScenariosForV1API } from './business';
+import { EnvOverride } from '../../../envUtils';
+
+/**
+ * gaec is not GA for the batching framework, so the transform flag has to be set alongside the
```
saikumarrs: **nit:** this has it backwards — gaec *is* GA. `features.ts:29` is `{ routerTransform: true, batching: true }` and `getBatchingFrameworkGaDestinations()` reads `batching`, so `isBatchingFrameworkEnabled` returns true before the env var is read. `GOOGLE_ADWORDS_ENHANCED_CONVERSIONS_BATCHING_FRAMEWORK_ENABLED_WORKSPACE_IDS` on line 14 is a no-op; only the `_DELIVERY_` one is load-bearing.

The next reader copying this pattern sets a var that does nothing. Same claim is in the PR description.

src/services/destination/nativeBatching/delivery.ts
```diff
@@ -0,0 +1,456 @@
...
+ * to override or to call `super` on. The resolution order — exact status, then status class, then
+ * the framework's own classification — *is* the contract; a destination that wants different
+ * behaviour declares an override for the status it cares about.
+ */
+export function handleDeliveryResponse(klass: unknown, ctx: DeliveryContext): HandleResponseResult {
+  const { statusOverrides, failureReason } = resolveDeliverySpec(klass);
+  const statusClass = statusClassOf(ctx.status);
+  const override =
+    statusOverrides[ctx.status] ?? (statusClass ? statusOverrides[statusClass] : undefined);
+  const fallback = () => classifyByStatus(ctx.status, failureReason(ctx));
```
saikumarrs: **nit:** `failureReason(ctx)` is evaluated before `classifyByStatus` decides it isn't needed, and the next line calls `fallback()` for every override-less response — so a 2xx pays a full extractor call whose result is discarded. `extractBrazeAudienceErrorMessage` on a body without `message` reaches `JSON.stringify(response)` over the whole payload.

Taking a thunk keeps the success path free:

```ts
export const classifyByStatus = (status: number, reason: () => string): Verdict => {
  if (isHttpStatusSuccess(status)) return success();
  if (status === 429) return throttled(reason());
  if (isHttpStatusRetryable(status)) return retry(reason());
  return abort(reason());
};
```

src/services/destination/nativeBatching/delivery.ts
```diff
@@ -0,0 +1,456 @@
...
+    perItemPreserved = true;
+  } else {
+    // Never index past the end — that is how a job state with `metadata: undefined` gets emitted,
+    // which rudder-server reports as a non-fatal in/out breach and then redelivers. Degrade to a
+    // whole-batch verdict instead, and make the mismatch visible.
+    stats.counter('batch_delivery_per_item_mismatch', 1, {
+      destType,
+      destinationId: ctx.destinationId,
+      workspaceId: ctx.workspaceId,
+      items: String(result.verdicts.length),
```
saikumarrs: **nit:** `items` and `jobs` are Prometheus label values, and `MAX_BATCH_SIZE` is 1000 for both `braze_audience` and `iterable_audience` — a 1000x1000 label space per (destType, destinationId, workspaceId). This counter fires exactly when a destination is stuck in the mismatch/retry loop, so the guard that surfaces the problem is what takes out the metrics backend.

Let's bucket them, or move the raw counts into the log line.

src/services/destination/nativeBatching/delivery.ts
saikumarrs: The empty family-level `delivery` gives the merge a real caller — fine to close this.

Separate thing on the same function: the walk continues only while `typeof current === 'function'`, so a missing or renamed `Integration` export resolves to the empty spec and the destination silently degrades to status-only classification. `braze_audience`, `customerio` and `iterable_audience` all report partial failures on a 2xx, so every rejected record would come back `statusCode: 200, error: 'success'` — dropped with no throw, no log, no metric. The transform path throws on the same mistake (`processBatchedDestination.ts:206`). Let's throw here too.

src/services/destination/nativeBatching/delivery.ts
shekhar-rudder: Done — `{ kind: 'abort'; reason: string; auth?: never }`.

Confirmed your reading exactly. The value comes back from a function call rather than a fresh object literal, so no excess-property check runs and `authRevoked()` was structurally assignable to the old arm. The `auth: 'revoked'` then went nowhere: the per-item mapping branches on `kind` alone, and the block that reads `auth === 'revoked'` to set `authErrorCategory` sits behind `!perItemPreserved`. So it degraded to a plain 400 with no signal to the author that their intent had been dropped.

Verified with `tsc` both ways — `perItem([authRevoked('x')])` and `perItem([authExpired('x')])` now both error, while `success()`/`abort()`/`retry()`/`throttled()` still compile. Whole-batch `authRevoked()` is unaffected: `AbortVerdict` keeps `auth?: 'revoked'`, so the restriction applies only to the per-item channel.

src/services/destination/nativeBatching/delivery.ts
shekhar-rudder: Done — `classifyByStatus(status: number, reason: () => string)`, with `handleDeliveryResponse` passing `() => failureReason(ctx)`.

Checked all four `failureReason` extractors are pure before making it lazy (gaec, braze_audience, iterable_audience, and the framework default — no stats, no side effects), so the verdicts and messages produced are identical; the only difference is that the discarded call no longer happens. Added a test asserting a 200 does not evaluate the reason and a 500 does.

src/services/destination/nativeBatching/delivery.ts
shekhar-rudder: Done — the counter now carries only `destType`/`destinationId`/`workspaceId`, and the two counts moved into a `logger.error` emitted alongside it.

Agreed on the failure mode, which is what made this worth fixing rather than bounding: the counter fires precisely when a destination is stuck in the mismatch/retry loop, so the guard that surfaces the problem would have been the thing that took the metrics backend down with it. A test pins the label set so it can't drift back.

.claude/skills/batching-framework-delivery/SKILL.md
shekhar-rudder: Good catch — documented in the skill.

`mapSuccessPayloadToServerFormat` writes the batch to `body[strategy.bodyFormat]` and hard-sets the other three to `{}`, so a destination on `JSON_ARRAY`, `FORM` or `XML` that reads `body.JSON` gets nothing on every response — success included — and the mismatch guard answers each one with an N-job 500. The skill now says to read the key your own `getBatchStrategy` returns, and calls out that all six spec'd destinations being on `JSON` today is exactly why no test would catch it.

src/services/destination/nativeBatching/delivery.ts
shekhar-rudder: Done — stated both at the `perItem` declaration and in the skill.

`ctx.jobs` is one entry per job, so a job contributing two body items puts the positional list and the job list permanently out of step: the guard retries the batch, the retry reproduces the same mismatch, and it never converges. Worth naming precisely because `batchDestination.ts` and the VDM V2 dispatch table both permit `transformEvent` to return an array, so nothing stops someone pairing the two.

src/services/destination/nativeIntegration.ts
shekhar-rudder: Partly done, and chasing it turned up something worth flagging.

I first did exactly what you suggested — used `isProxyV1Request` in the catch — and it broke `intercom_v2`. `INTERCOM_V2_v0_oauth_scenario_1` declares `version: 'v0'` but builds its body with `generateProxyV1Payload`, so it is a v0-route request carrying an array `metadata`. Route and array-ness genuinely disagree in the suite today.

Which means the hazard also runs the other way, on the branch this PR adds: shape-alone would take a v0-route request carrying an array down the framework branch and answer it with a `DeliveryV1Response` — a body rudder-server's v0 reader cannot parse. So I have tightened the framework guard to require **both** the v1 route and the array, and left the catch on `version === 'v1'` untouched.

That gives the property this comment is really after. The framework branch is now strictly narrower than the catch's test, so any request the framework branch accepts is one the catch also treats as v1 — a single call can no longer answer with a v1 body when the destination succeeds and a v0 body when it fails.

Leaving the catch alone is deliberate: changing it would only affect a v1 route with non-array `metadata`, which rudder-server cannot send and which already panics inside `handlevV1DeliveriesFailureEvents`. Separately, the `intercom_v2` fixture looks wrong to me — a v0 scenario with a v1 payload — but I have left it alone as out of scope here.

src/services/destination/nativeIntegration.ts
shekhar-rudder: Deliberate, and it is parity rather than a divergence.

On the legacy path `deliver` returns `networkHandler.responseHandler(...)` straight through, and no network handler calls `reportError` either — grep over `src/adapters/` and `src/v1/destinations/` comes back empty. Reporting fires only from `postTransformation`, which is to say only on the throw path, for the framework and the legacy handlers alike. A returned per-job failure is a delivery outcome the router acts on, not a transformer error.

The `getTags` merge just above is different in kind, which is why it is there and this is not: without it `integration.failure_detailed` would carry a bare `errorType` with no destination or workspace to attribute it to, so returned and thrown failures had to be made to agree. Routing every returned per-job failure into error reporting would instead be new behaviour for the 20 v1 handlers this replaces.

Added a comment at the return site recording the reasoning.

test/integrations/destinations/google_adwords_enhanced_conversions/dataDelivery/data.ts
shekhar-rudder: Confirmed, and fixed by dropping them from the `.map`.

You are right that they never reach the delivery branch: the mocked 401/403 sit on `googleAds:searchStream`, so `gaecProxyRequest` throws inside `getConversionActionId` before the upload ever happens. Enrolling them advertised framework coverage that did not exist.

The framework auth path stays covered where it is actually reachable — `gaec_v1_scenario_4` mocks the 401 on `uploadConversionAdjustments` and asserts the framework's REFRESH_TOKEN response with `errorType: 'retryable'`, and `delivery.test.ts` covers the 403 → AUTH_STATUS_INACTIVE branch against the legacy handler.

test/integrations/destinations/google_adwords_enhanced_conversions/dataDelivery/data.ts
shekhar-rudder: You are right — removed the var and rewrote the comment.

`features.ts:29` is `{ routerTransform: true, batching: true }` and `getBatchingFrameworkGaDestinations()` reads `batching`, so `isBatchingFrameworkEnabled` returns true before the env var is ever read. Only `_DELIVERY_` is load-bearing here. The comment now states that outright, and explains why a genuinely pre-GA destination *would* need both — `isBatchingFrameworkDeliveryEnabled` short-circuits on the transform flag.

Noting that the PR description carries the same claim; flagging it for a separate correction rather than folding a description edit into this change.

src/v0/destinations/google_adwords_enhanced_conversions/delivery.ts
shekhar-rudder: Agreed on the principle — `results` is the array with no length guarantee, and the comment directly above it argues against indexing off exactly that.

Not taking it in this PR, though. It changes what the override returns for the unreadable-body case, from a lost-attribution retry to per-adjustment aborts, and this PR is scoped to moving the handler onto the class without changing how a response is read. As you note it is unreachable today — `networkHandler.ts` dereferences `conversionAdjustments!` before delivery runs — so nothing is exposed in the meantime.

Deferred with the other delivery-logic items; they are listed at the bottom of the commit message so they do not get lost.

src/services/destination/nativeBatching/delivery.ts
shekhar-rudder: Confirmed, and it is worse than a mistagged counter: `DeliveryV1ResponseSchema` refines on `validateStatTags` (`zodTypes.ts:169-174,220`), so a non-2xx response with absent `statTags` is rejected outright — and that is precisely the mixpanel `/import`-on-400 case the `perItemPreserved` exclusion exists to let through.

Not fixing it here. A floor is new response-shape logic and this PR is meant to be a refactor. Your read of the reachability matches mine: every current override is 2xx-keyed except gaec's `4xx`, which returns whole-batch verdicts, so the first destination that hits this gets a failing component test rather than a shipped bug.

Deferred with the other delivery-logic items, listed in the commit message.

src/v0/destinations/iterable_audience/delivery.ts
shekhar-rudder: Answering the question first: `subscribers` is guaranteed present today, because the framework builds this body itself — `mapSuccessPayloadToServerFormat` writes it to `body.JSON`, and iterable_audience is on the default `JSON` body format. So the `?? []` is unreachable rather than load-bearing, which is why it has not bitten.

It should still degrade rather than hand `perItem` an empty list, and I had it doing that, but I have backed it out: it changes what the override returns, and this PR is scoped to the refactor. Deferred with the rest.

One note for whoever picks it up. iterable cannot degrade to `ctx.jobs` the way `braze_audience` and `gaec` can — it is content-keyed, matching response identities against each subscriber's `email`/`userId`, and job metadata carries neither. So it needs a whole-batch verdict rather than a job-list fallback, which is a real difference from the other two and not just a stylistic one.

src/services/destination/nativeBatching/delivery.ts
shekhar-rudder: Confirmed the mechanism. `MiscService.getBatchDestinationHandler` is `require(...).Integration`, so a missing or renamed export arrives as `undefined`, the walk resolves to the empty spec, and that is indistinguishable from a destination that genuinely classifies on status alone. For the three destinations that report partial failures on a 2xx, every rejected record would come back `statusCode: 200, error: 'success'` — dropped with no throw, no log, no metric.

I had it throwing and have backed that out for the same reason as the other items: it adds a new runtime failure mode, and this PR is meant to be a refactor rather than a behaviour change. Unreachable for the five wired destinations, so deferring is safe.

Listed in the commit message with the rest so it is easy to pick up.

src/services/destination/nativeBatching/delivery.ts
shekhar-rudder: Reversing my previous reply — this one is in, as of b2981d1f.

The deciding factor was that the fix is entirely framework-level: `statTags` is built only in `toDeliveryV1Response`, no destination spec touches it (grep across every `delivery.ts` returns nothing), and the identifying-tag merge in `nativeIntegration.ts` is already conditional on `statTags` being present, so it picks the new cases up unchanged. All 3930 component tests pass with no fixture edits, which is the same thing your reachability analysis predicted.

The gate is now `describesOneFailure || !isHttpStatusSuccess(ctx.status)`, and where the failures do not agree the errorType comes from a failing verdict rather than from `first`, which can be a success on a mixed batch. A 2xx is deliberately untouched, so a partly-delivered batch still carries no whole-response tag set — the gaec-legacy-handler behaviour the original rule exists to avoid.

One correction to what I wrote earlier: I said the response would be "rejected outright", which overstates it. `DeliveryV1ResponseSchema` is only ever parsed by the test harnesses — `testUtils.ts:628` and `live/routerProxyRequests.ts:56` — with no reference anywhere in `src/`. So the real cost was a failing component/live test for the next destination that returns per-item detail on a non-2xx, which is what you said in the first place.

src/services/destination/nativeBatching/delivery.ts
shekhar-rudder: Reversing my previous reply — this one is in too, as of b2981d1f.

Same reasoning as the `statTags` thread: the change is confined to `resolveDeliverySpec`, no destination participates, and nothing else moves. `resolveDeliverySpec` now throws a `PlatformError` when handed anything that is not a class, naming the likely cause (a `routerTransform` module not exporting `Integration`), with a test covering `undefined`, `null` and a plain object.

Unreachable for the five wired destinations, so no current behaviour changes — it just means the failure surfaces at the point of the mistake instead of as silent `statusCode: 200, error: 'success'` on every rejected record.

src/services/destination/nativeBatching/delivery.ts
shekhar-rudder: Reverting this again, as of 147da143 — and this time with the reasoning corrected rather than the scope.

I justified the floor on the claim that a non-2xx top-level `status` already tells rudder-server the response failed, so tagging it merely agreed with what we were already sending. That claim does not survive checking:

- On the v1 route the controller returns **HTTP 200** for this response unless `authErrorCategory` is set — `deliverToDestinationV1` calls `deliveryPostProcess(ctx)` with no status argument in every other case (`controllers/delivery.ts`). The v0 route passes the body status through unconditionally; v1 does not.
- `ProxyResponseV1` has no `status` field at all — `router/transformer/transformer_proxy_adapter.go:41-45` declares only `message`, `response` and `authErrorCategory`, so the field is dropped on unmarshal.

So rudder-server never learns of a non-2xx here. It sees 200 and takes disposition from each job's `statusCode`, exactly as the design intended.

That leaves `validateStatTags` as a refinement in our own zod schema, parsed only by the test harnesses, enforcing on v1 an invariant that is genuinely load-bearing on v0 — where the body status *does* become the wire status — but not here. Satisfying it would be a test-harness accommodation rather than a correctness fix, which is not something a refactor should carry.

Your original framing was the accurate one throughout: the cost is that the first destination returning per-item detail on a non-2xx fails its component test. That is a real thing to fix, but the right fix is probably to scope `validateStatTags` to the version where the field is actually read, which is well outside this PR.

The original rule is restored unchanged — a whole-response tag set only when the response as a whole is one failure. The `resolveDeliverySpec` guard from the same commit stays in; its justification never depended on any of this.

saikumarrs (approved):

PR #5424, opened 2026-07-30
feat(braze): add live integration test suite

achettyiitr:
## Description
- Enrolls Braze into the live (real-destination) integration suite covering every deliverable transform code path — and extends the shared live harness to support them.
- Resolves INT-6898

### Braze scenarios (`test/integrations/destinations/braze/live/`)

**Event-stream**
- identify by `external_id`, anonymous identify (`rudder_id` user_alias), identity resolution (`/users/identify`), custom track event, order completed (legacy purchases), group attribute, `dontBatch`

**Recommended ecommerce** (`useEcommerceRecommendedEvents`)
- `product_viewed`, `order_placed`, `cart_updated` (add + remove), `checkout_started`, `order_refunded`, `order_cancelled`

**Config / message variants**
- `supportDedup` (fresh + existing-user reduce), `enableNestedArrayOperations`, `sendPurchaseEventWithExtraProperties`, multi-product / product-level currency / sku fallback, `brazeExternalId` external id, `integrations.Braze.alias` custom label, RETL (`mappedToDestination`) identify, and per-job delivery mapping (`BRAZE_PER_JOB_DELIVERY_MAPPING`) for the track and merge branches

**Other endpoints**
- alias → `/users/merge` (delivery-only; Braze processes merges asynchronously, no read-back SLA)
- group → `/v2/subscription/status/set` — parked (`enabled: false`) until a `subscription_group_id` is supplied via `resourceIds`

Read-backs use `/users/export/ids` and `/subscription/user/status`; cleanup via `/users/delete`. Auth is a single `restApiKey` in `LIVE_SECRET_BRAZE.config`.

### Shared harness (`test/integrations/live/`)

- Explicit `MetadataOverride` (`{ workspaceId?, dontBatch? }`) and `DestinationOverride` (`{ WorkspaceID? }`) step-option types, replacing loose `Record` types.
- `destinationOverride` threaded through `buildRouterTransformBody` so transforms gated on `destination.WorkspaceID` (Braze per-job mapping) are exercisable without conflating it with `metadata.workspaceId`. `destination.Config` stays owned by `resolveConfig` + `configOverride`.
- Scenario `verify` checks are single-shot; the framework owns retry/backoff (fixes a compounding poll that could exceed the 60s per-scenario timeout).

🤖 Generated with

### DIFF
--- a/test/integrations/destinations/braze/live/api.ts
+++ b/test/integrations/destinations/braze/live/api.ts
@@ -1,0 +1,201 @@
+import axios from 'axios';
+import { Agent } from 'https';
+import { RunContext } from '../../../live/types';
+import { RUDDER_ALIAS_LABEL } from './profiles';
+
+// keepAlive:false so read-back/cleanup sockets don't linger as open handles when the suite finishes.
+const brazeAgent = new Agent({ keepAlive: false });
+
+// Fields the read-back requests from Braze's export endpoint.
+const FIELDS_TO_EXPORT = [
+  'external_id',
+  'email',
+  'first_name',
+  'last_name',
+  'custom_attributes',
+  'user_aliases',
+];
+
+// Credentials for the direct Braze API helpers (read-back + cleanup). Same restApiKey/dataCenter the
+// transform authenticates with — the key just needs users.export.ids + users.delete (+ subscription
+// scopes) in addition to the delivery scopes.
+const apiCreds = (ctx: RunContext): { restApiKey: string; dataCenter: string } => {
+  const config = (ctx.liveSecret.config ?? {}) as Record<string, unknown>;
+  const restApiKey = config.restApiKey as string | undefined;
+  const dataCenter = config.dataCenter as string | undefined;
+  if (!restApiKey) {
+    throw new Error('Braze restApiKey missing from LIVE_SECRET_BRAZE.config');
+  }
+  if (!dataCenter) {
+    throw new Error('Braze dataCenter missing from LIVE_SECRET_BRAZE.config');
+  }
+  return { restApiKey, dataCenter };
+};
+
+// Mirrors src/v0/destinations/braze/util.ts getEndpointFromConfig so read-back/cleanup hit the same
+// cluster the transform delivered to.
+export const restEndpoint = (dataCenter: string): string => {
+  const [region, number] = dataCenter.trim().toLowerCase().split('-');
+  switch (region) {
+    case 'eu':
+      return `https://rest.fra-${number}.braze.eu`;
+    case 'us':
+      return `https://rest.iad-${number}.braze.com`;
+    case 'au':
+      return `https://rest.au-${number}.braze.com`;
+    default:
+      throw new Error(`Invalid Braze Data Center: ${dataCenter} (valid regions: EU, US, AU)`);
+  }
+};
+
+const authHeaders = (restApiKey: string) => ({
+  'Content-Type': 'application/json',
+  Accept: 'application/json',
+  Authorization: `Bearer ${restApiKey}`,
+  Connection: 'close' as const,
+});
+
+export type BrazeUserProfile = {
+  external_id?: string;
+  email?: string;
+  first_name?: string;
+  last_name?: string;
+  custom_attributes?: Record<string, unknown>;
+  user_aliases?: Array<{ alias_name: string; alias_label: string }>;
+};
+
+type ExportSelector = {
+  externalIds?: string[];
+  userAliases?: Array<{ alias_name: string; alias_label?: string }>;
+};
+
+// POST /users/export/ids — the read-back. Returns the matched user profiles (empty array if none).
+export const exportUsers = async (
+  ctx: RunContext,
+  selector: ExportSelector,
+): Promise<BrazeUserProfile[]> => {
+  const { restApiKey, dataCenter } = apiCreds(ctx);
+  const body: Record<string, unknown> = { fields_to_export: FIELDS_TO_EXPORT };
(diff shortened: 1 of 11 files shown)


test/integrations/destinations/braze/live/api.ts
```diff
@@ -0,0 +1,201 @@
...
+  }
+  try {
+    await axios.post(`${restEndpoint(dataCenter)}/users/delete`, body, {
+      headers: authHeaders(restApiKey),
+      httpsAgent: brazeAgent,
+      timeout: 15000,
+    });
+  } catch (err) {
+    // eslint-disable-next-line no-console
+    console.error('[live:braze] teardown (users/delete) failed', err);
```
maheshkutty: **Leaks the REST API key into CI logs.** `console.error` on an `AxiosError` inspects the error's own enumerable properties, which include `config.headers.Authorization: Bearer <restApiKey>` (plus a large `request`/`response` blob). The harness already redacts `metadata.secret` in `runPipelineStep`'s job states, so this is inconsistent with the surrounding convention as well as risky on a public CI log.

Suggest logging only the message and status:

```ts
} catch (err) {
  const status = (err as { response?: { status?: number } })?.response?.status;
  // eslint-disable-next-line no-console
  console.error(
    `[live:braze] teardown (users/delete) failed${status ? ` (status ${status})` : ''}:`,
    err instanceof Error ? err.message : String(err),
  );
}
```

test/integrations/destinations/braze/live/verify.ts
```diff
@@ -0,0 +1,95 @@
...
+  const key = groupAttributeKey(ctx);
+  const profile = await exportUserByExternalId(ctx, ctx.identity('user'));
+  expect(profile).not.toBeNull();
+  expect(profile?.custom_attributes ?? {}).toMatchObject({ [key]: true });
+};
+
+// Alias merge: after /users/merge collapses the source (previousId) into the kept (userId) profile,
+// the source-only marker attribute must surface on the kept profile. Braze processes merges
+// asynchronously, so the scenario gives this a wider retry window (MERGE_READBACK in spec.ts).
+export const verifyMerge = async (ctx: RunContext): Promise<void> => {
```
maheshkutty: `verifyMerge` is exported but never referenced — `braze-alias-merge` deliberately ships without a read-back. Either drop it (git remembers it), or keep it and make the dead-code status explicit in the doc comment; right now a reader has to grep to find out it isn't wired up.

If it goes, `mergeSourceMarker` in `profiles.ts` and the marker attribute written by `createMergeUsers` are only there to support it — worth pruning together or noting why they stay.

test/integrations/destinations/braze/live/verify.ts
```diff
@@ -0,0 +1,95 @@
...
+import {
+  BrazeUserProfile,
+  exportUserByAlias,
+  exportUserByExternalId,
+  getSubscriptionGroups,
+  subscriptionGroupId,
+} from './api';
+
+// Each check here is a SINGLE-SHOT assertion. The scenario-level `verify` block owns the
+// retry/backoff for Braze's eventually-consistent export (see READBACK/MERGE_READBACK in spec.ts),
```
maheshkutty: `MERGE_READBACK` doesn't exist in `spec.ts` (only `READBACK` does) — same stale reference on line 68. Please drop the `/MERGE_READBACK` here and reword line 68, otherwise the next reader goes looking for a constant that was never added.

test/integrations/destinations/braze/live/spec.ts
```diff
@@ -0,0 +1,900 @@
...
+            },
+          }),
+        },
+      ],
+      verify: { check: verifyProfileByExternalId(ecomCartUpdatedTraits), ...READBACK },
+    },
+
+    // ─── alias → /users/merge ───
+    // Distinct endpoint. Setup creates the kept (userId) and source (previousId) profiles; the alias
+    // merges the source into the kept, and the read-back asserts the source-only marker attribute
```
maheshkutty: This block comment says "the read-back asserts the source-only marker attribute surfaced on the kept profile (Braze merges asynchronously — the verify polls a wide window)", but the scenario immediately below has no `verify`, and the comment at line 379 says exactly the opposite. Please reconcile — keep the line 379 explanation (delivery-only because merges are async) and fix this one.

test/integrations/destinations/braze/live/setup.ts
```diff
@@ -0,0 +1,43 @@
...
+export const createMergeUsers = async (ctx: RunContext): Promise<void> => {
+  const keepId = ctx.identity('user');
+  const mergeId = ctx.identity('merge-user');
+  await createUserWithAttributes(ctx, keepId, { first_name: 'CI-Merge-Keep' });
+  await createUserWithAttributes(ctx, mergeId, {
+    first_name: 'CI-Merge-Source',
+    ...mergeSourceMarker(ctx),
+  });
+  ctx.register({ type: 'braze-user', id: keepId });
+  ctx.register({ type: 'braze-user', id: mergeId });
```
maheshkutty: Nit: these `ctx.register(...)` calls are unused bookkeeping — teardown for this scenario goes through `deleteMergeUsers` in `spec.ts`, and nothing reads `ctx.resources` for `braze-user`. `createSubscriptionUser` just below doesn't register at all, so the two setups disagree. Pick one convention (either drop these, or register in both and note that registration is informational here).

maheshkutty (commented):
Two follow-ups from reviewing the live suite: a secret-redaction issue in the teardown logging, and some dead code / comments that contradict the code around the alias-merge scenario. Details inline.

maheshkutty (approved):

PR #5445, opened 2026-08-06
chore: cleaning up openfaas

fracasula:

* First commit removes OpenFaaS
* Second commit fixes some typos that I came across too while refactoring

Fixes [PIPE-3222](https://linear.app/rudderstack/issue/PIPE-3222/remove-openfaas-code-from-rudder-transformer)


mihir20 (dismissed):
LGTM, we can delete this file as well: test/__tests__/data/user_transformation_pycode_test_output.json

mihir20 (approved):

src/util/customTransformationsStore-v1.js
```diff
@@ -4,37 +4,12 @@ const { responseStatusHandler } = require('./utils');
...
 const rudderLibraryCache = {};
 
-const getTransformationURL = `${CONFIG_BACKEND_URL}/transformation/getByVersionId`;
 const getLibrariesUrl = `${CONFIG_BACKEND_URL}/transformationLibrary/getByVersionId`;
 const getRudderLibrariesUrl = `${CONFIG_BACKEND_URL}/rudderstackTransformationLibraries`;
 
-// Gets the transformation from config backend.
-// Stores the transformation object in memory with time to live after which it expires.
-// VersionId is updated any time user changes the code in transformation, so there wont be any stale code issues.
-async function getTransformationCodeV1(versionId) {
```
Jayachand, line 18: why not keep fetching transformation logic in this file itself along with libraries and remove file `customTransformationsStore.js‎`

Jayachand (approved):

src/util/customTransformationsStore-v1.js
fracasula: `getLibraryCodeV1` and `getRudderLibByImportName` (i.e. `customTransformationsStore-v1.js`) are both called only from ivmFactory, which only the v1 executor uses. Only `getTransformationCode` (i.e. `customTransformationsStore.js`) serves both  v0 and v1.

If you don't want to make the v0<->v1 distinction anymore and consolidate everything into one file (e.g. `customTransformationsStore.js`) then it might be a good opportunity to reconsider caching as well.

The `getTransformationCode` in `customTransformationsStore.js` has a 24h cache, unlike the one that I removed which had a perpetual cache with no expiry and it was used for OpenFaaS only.

Same for `getLibraryCodeV1` and `getRudderLibByImportName`, the cache is unbounded.

So if we were to consolidate everything into one file we could use the 24h cache for everything (libraries, code and rudder libraries) instead of the unbounded one (i.e. one `NodeCache` for each function).

It's out of scope for this PR but I could pick it up here anyway.

Jayachand: yeah, can be picked in other PR.

krishna2020 (approved):

shekhar-rudder (approved):

src/util/customTransformationsStore-v1.js
fracasula: @Jayachand follow up PR [here](https://github.com/rudderlabs/rudder-transformer/pull/5462).

PR #5449, opened 2026-08-09
feat: live integrations tests for criteo audience

achettyiitr:

Adds a live (real-destination) integration test suite for **criteo_audience**, plus the OAuth harness the suite needs to mint real access tokens via rudder-auth, and the CI wiring to run it.

- **criteo_audience live suite** (`test/integrations/destinations/criteo_audience/live.ts`): scenarios covering the Audience Type options — `add-email`, `add/remove-madid`, `add-identityLink`, and a RETL `email` flow. Each seeds an event, drives `/routerTransform → /proxy` against the real Criteo API, and cleans up via a real remove PATCH. Audience IDs come from the secret's `resourceIds` (`emailAudienceId`, `maidAudienceId`, `identityLinkAudienceId`).
- **OAuth harness** for `authType: 'oauth'` destinations:
  - `live/rudderAuthContainer.ts` — starts the rudder-auth container via `testcontainers` (image pulled from ECR), waits on `/health`, exposes a dynamic mapped port, and forwards credential-shaped env vars (`CLIENT_ID|CLIENT_SECRET|CONSUMER_KEY|CONSUMER_SECRET`) into the container so tokens can be minted with a test app instead of the image defaults.
  - `live/oauthTokenResolver.ts` — posts the refresh token to `/tokens/destination/<dest>/refresh` and returns the access token.
  - `component.live.test.ts` — starts the container once when any enrolled destination is OAuth, mints the token per destination, and injects it into `metadata.secret.accessToken`.
- **CI** (`.github/workflows/live-integration-tests.yml`): `criteo_audience` added to the matrix. A destination opts into the rudder-auth path purely by declaring its OAuth app-cred Vault mappings in the matrix `appSecrets` field; a single Vault step pulls both the destination secret and those app creds, and the ECR login/pre-pull steps gate on `appSecrets` being set (apiKey destinations skip them).

- Resolves INT-6946

### DIFF
--- a/.github/workflows/live-integration-tests.yml
+++ b/.github/workflows/live-integration-tests.yml
@@ -46,8 +46,8 @@ jobs:
           - destination: vero
           - destination: criteo_audience
             appSecrets: |
-              control-plane/data/external-services CRITEO_AUDIENCE_CLIENT_ID_DESTINATION
-              control-plane/data/external-services CRITEO_AUDIENCE_CLIENT_SECRET_DESTINATION
+              control-plane/data/external-services CRITEO_AUDIENCE_CLIENT_ID_DESTINATION ;
+              control-plane/data/external-services CRITEO_AUDIENCE_CLIENT_SECRET_DESTINATION ;
     steps:
       - name: Harden the runner (Audit all outbound calls)
         uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
@@ -82,7 +82,7 @@ jobs:
           jwtGithubAudience: ${{ vars.VAULT_ADDR }} # MUST match bound_audiences, else auth fails
           exportToken: false # keep the raw Vault token out of env
           secrets: |
-            engineering_shared/data/integrations_team/e2e_test/rudder-transformer ${{ steps.secret.outputs.name }} | ${{ steps.secret.outputs.name }}
+            engineering_shared/data/integrations_team/e2e_test/rudder-transformer ${{ steps.secret.outputs.name }} | ${{ steps.secret.outputs.name }} ;
             ${{ matrix.appSecrets }}
 
       - name: Configure AWS credentials for ECR (OIDC)
--- a/.github/workflows/live-integration-tests.yml
+++ b/.github/workflows/live-integration-tests.yml
@@ -40,7 +40,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        destination: [hs, braze_audience, braze, vero]
+        include:
+          - destination: hs
+          - destination: braze_audience
+          - destination: vero
+          - destination: criteo_audience
+            appSecrets: |
+              control-plane/data/external-services CRITEO_AUDIENCE_CLIENT_ID_DESTINATION
+              control-plane/data/external-services CRITEO_AUDIENCE_CLIENT_SECRET_DESTINATION
     steps:
       - name: Harden the runner (Audit all outbound calls)
         uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
@@ -64,7 +71,7 @@ jobs:
         id: secret
         run: echo "name=LIVE_SECRET_$(echo '${{ matrix.destination }}' | tr 'a-z' 'A-Z')" >> "$GITHUB_OUTPUT"
 
-      - name: Import destination secret from dev Vault (OIDC/JWT)
+      - name: Import secrets from dev Vault (OIDC/JWT)
         id: vault
         uses: hashicorp/vault-action@892a26828f195e65540a40b4768ae4571f51ebfc # v4.0.0
         with:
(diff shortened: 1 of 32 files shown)


test/integrations/live/rudderAuthContainer.ts
```diff
@@ -0,0 +1,75 @@
...
+  // These override the app creds baked into the image (needed when the refresh token was issued by
+  // a different app than the image ships). Only the destinations whose CI matrix entry supplies
+  // these vars have them in env, so the forwarding is scoped without any per-destination code.
+  private credentialEnv(): Record<string, string> {
+    const env: Record<string, string> = {};
+    for (const [key, value] of Object.entries(this.env)) {
+      if (value === undefined) {
+        continue;
+      }
+      if (/(CLIENT_ID|CLIENT_SECRET|CONSUMER_KEY|CONSUMER_SECRET)/.test(key)) {
```
saikumarrs: What are `CONSUMER` envs for?

test/integrations/live/oauthTokenResolver.ts
```diff
@@ -0,0 +1,114 @@
...
+      );
+    }
+    const v1 = await this.tryV1(secret);
+    if (v1.ok) {
+      return v1.secret;
+    }
+    const legacy = await this.tryLegacy(destination, refreshToken);
+    if (legacy.ok) {
+      return legacy.secret;
+    }
```
saikumarrs: Can't we define which route that destination supports?

.github/workflows/live-integration-tests.yml
```diff
@@ -40,7 +40,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        destination: [hs, braze_audience, braze, vero]
+        include:
+          - destination: hs
+          - destination: braze_audience
+          - destination: vero
+          - destination: criteo_audience
```
saikumarrs: This will not be sustainable as we add more destinations. Let's add a step to first resolve the destination names based on the directory names of the live tests.

.github/workflows/live-integration-tests.yml
```diff
@@ -40,7 +40,14 @@ jobs:
...
+            appSecrets: |
+              control-plane/data/external-services CRITEO_AUDIENCE_CLIENT_ID_DESTINATION ;
+              control-plane/data/external-services CRITEO_AUDIENCE_CLIENT_SECRET_DESTINATION ;
```
saikumarrs: There is a way to automatically infer the secret names from the destination names. This will avoid manually defining them in the workflow.
You can refer to rudder-auth implementation.

.github/workflows/live-integration-tests.yml
```diff
@@ -75,7 +82,23 @@ jobs:
...
+          role-to-assume: ${{ vars.AWS_ECR_IAM_ROLE_ARN }}
+          aws-region: us-east-1
+
+      - name: Log in to Amazon ECR
+        if: matrix.appSecrets != ''
+        uses: aws-actions/amazon-ecr-login@d539f0932e70871a027e9d5a9d8fc38589180a64 # 2.1.6
+
+      - name: Pre-pull rudder-auth image (surface ECR/pull errors before the test run)
+        if: matrix.appSecrets != ''
+        run: docker pull 422074288268.dkr.ecr.us-east-1.amazonaws.com/rudderstack/rudder-auth:develop
```
saikumarrs, line 112: Leaving this comment as a reminder to fix the tag from `develop` to the new default branch.

.github/workflows/live-integration-tests.yml
saikumarrs: [Blocker] Before we proceed, let's make sure client credentials for criteo audience in dev vault are not the same as the production env.
If that's the case, we should first create a dev OAuth app and use it.

We can't risk exposing the production credentials.

test/integrations/live/rudderAuthContainer.ts
achettyiitr: Some have consumer secrets, refer https://github.com/rudderlabs/rudder-auth/blob/develop/config/default.yml#L286

test/integrations/live/oauthTokenResolver.ts
achettyiitr: This would require keeping some hard-coded mappings, which is not needed. We can simply make both api calls, as it is just for tests to simplify.

.github/workflows/live-integration-tests.yml
achettyiitr: If it is just about adding destinations, then we are good, but for these destinations, some have secrets, and these secrets also vary.

> This will not be sustainable as we add more destinations.

- Adding a single line here for setting up the integration test would be sufficient I think.

.github/workflows/live-integration-tests.yml
achettyiitr: It's not just CLIENT_ID, secrets can be CONSUMER* or DEVELOPER_TOKEN* as well so we can't generalise.

.github/workflows/live-integration-tests.yml
achettyiitr: Once we completely migrate to main, we will update here.

.github/workflows/live-integration-tests.yml
achettyiitr: Both apps are different.

test/integrations/live/oauthTokenResolver.ts
saikumarrs: We already do a great deal of hard-coding for defining the destinations. IMO this change would not be make huge difference.

I'm concerned if we would accidentally match any left over legacy implementation.

.github/workflows/live-integration-tests.yml
saikumarrs: I didn't get you. Can you please elaborate?

Each destination has a dedicated secret key, right? If we follow the convention, it should be straightforward.

> Adding a single line here for setting up the integration test would be sufficient I think.

That'd be even worse on top of having to define the destinations itself.

.github/workflows/live-integration-tests.yml
saikumarrs: Yeah, let's add them too. There are not a lot of variations.
Everything is keyed by the "destination" name.

.github/workflows/live-integration-tests.yml
achettyiitr: It's not about destinations; for each, we are also pointing to app secrets here.

> That'd be even worse on top of having to define the destinations itself.

Not sure why it's worse. Here we are just registering these destinations. Later in future if we need to club few destiantions to run within one jobs, we only need changes here.

.github/workflows/live-integration-tests.yml
achettyiitr: Do you mean blindly pass  *CONSUMER* or DEVELOPER_TOKEN* ?

.github/workflows/live-integration-tests.yml
saikumarrs: My point is that we have to avoid hard-coding the destinations to test and their secrets mapping.

Grouping destinations in jobs is a different concern which can be solved by another custom step, if needed. Even then you'd want to avoid the hard coding.

.github/workflows/live-integration-tests.yml
saikumarrs: If there is secret defined in the vault for that destination, it should be mapped, right? I'm assuming it won't function properly otherwise.

Can you confirm? If not, needs a secret clean up I guess.

saikumarrs (dismissed):

saikumarrs (approved):

PR #5459, opened 2026-08-12
fix(fb_custom_audience): treat non-2xx proxy responses without an error body as failures

shekhar-rudder:
## Summary
- `errorResponseHandler` (shared by `fb_custom_audience`, `facebook_pixel`, `facebook_conversions`) only checked for `response.error` to decide success vs. failure. When Facebook returned a non-2xx status (observed: `500`) with a body that lacked the usual `{ error: {...} }` shape, the response was silently treated as **successful** despite the failing status.
- Fix: when the status is non-2xx and there's no `error` field, throw a `NetworkError` (classified via the existing `getDynamicErrorType`) so it flows through the normal failure path and produces a well-formed per-job response instead of a false "success".

## Root cause investigation (verified against live logs/metrics — no guesswork)
Traced with real Grafana/Loki/Mimir data for `destinationId=xxx` (`FB_CUSTOM_AUDIENCE`, `entusmtprod`), confirming the exact mechanism end-to-end:

1. `router_transformerproxy_invalid_response{reason="missing output", destType="FB_CUSTOM_AUDIENCE", destinationId="xxxx"}` incremented exactly 3 times at 01:47:56.190Z, 01:48:03.047Z and 01:48:32.096Z UTC on 2026-08-12 (Mimir datasource `xxxxxx`), matching rudder-server's own `[TransformerProxy] proxy response missing output` warn logs, each carrying `statusCode: 500` and a **6000-jobID** batch.
2. rudder-transformer's own logs for the same 3 timestamps (`xxxxxx` / `delivery-transformer`, pod `xxxx`) show the real failure: `RangeError: Invalid string length` thrown at `JSON.stringify` inside Koa's `response.js:263` `get length` getter, called from `addRequestSizeMiddleware` (`src/middleware.ts:41`) — a stats-only middleware that measures `ctx.response.length` purely to emit an `http_response_size` histogram.
3. Why the body was that large: because `errorResponseHandler` didn't throw, the batch was classified as "success". `NativeIntegrationDestinationService.deliver`'s v0→v1 adaptation (`src/services/destination/nativeIntegration.ts`) then builds one job-state entry **per job in the batch**, each carrying a **full duplicate** of the raw destination response — i.e. the payload gets multiplied 6000×, large enough to exceed V8's max string length.
4. `errorHandlerMiddleware` (`src/middlewares/errorHandler.ts`) catches the `RangeError`, overwrites `ctx.body` with a generic `{ error, message }` (no `output` key) and sets `ctx.status = 500` — this is exactly what rudder-server sees as "missing output".

This fix directly addresses step 3's trigger condition (non-2xx + no `error` field), by routing it to the failure path instead. On the failure path, `DestinationPostTransformationService.handlevV1DeliveriesFailureEvents` does **not** duplicate the raw response the same way — it reads `error.destinationResponse?.response`, which for Facebook's typical (object/string) error bodies resolves to `undefined` and falls back to the short `errObj.message` string, so per-job entries stay small regardless of batch size. Verified with a live 5-job batch (see Test plan).

Fixes INT-6978 (https://linear.app/rudderstack/issue/INT-6978)

## Known review findings (not fixed by this PR)
The response-duplication behavior in step 3 above is a separate, still-open structural issue: `NativeIntegrationDestinationService.deliver`'s v0→v1 adaptation duplicates the **entire raw destination response** once per job with no size cap, for **any** non-throwing ("success") response — including a genuine 2xx. If a real Facebook success response is ever large enough (e.g. a verbose batch ack) combined with a large enough job batch, the same `RangeError: Invalid string length` → "missing output" failure mode can still occur, independent of this fix. Fixed (not just characterized) in a dedicated follow-up, since it's a framework-level concern (applies to any v0-handler destination adapted to v1, not just `fb_custom_audience`): **#5481**.

## Test plan
- [x] Added a regression test reproducing the exact failure mode (non-2xx status, no `error` field) in `networkHandler.test.ts`
- [x] Added a batched (v1, 5-job) integration regression test reproducing the real incident's shape in `test/integrations/destinations/fb_custom_audience/dataDelivery/business.ts` + `network.ts` (`fbca_v1_scenario_batch_infra_failure`) — verifies every job in the batch is marked failed with status 500 instead of silently succeeding
- [x] `npx jest src/v0/util/facebookUtils` — all passing
- [x] `npx jest --testPathPattern="nativeIntegration|facebookUtils|fb_custom_audience"` — 101/101 passing
- [x] `npm run test:ts -- component --destination=fb_custom_audience` — 51/51 passing
- [x] `npm run test:ts -- component --destination=facebook_pixel` — 59/59 passing (shares the same handler)
- [x] `npm run test:ts -- component --destination=facebook_conversions` — 16/16 passing (shares the same handler)
- [x] `npm run lint` clean

### DIFF
--- a/src/services/destination/__tests__/nativeIntegration.test.ts
+++ b/src/services/destination/__tests__/nativeIntegration.test.ts
@@ -3,11 +3,9 @@ import {
   ProcessorTransformationOutput,
   ProcessorTransformationRequest,
   ProcessorTransformationResponse,
-  ProxyV1Request,
 } from '../../../types/index';
 import { NativeIntegrationDestinationService } from '../nativeIntegration';
 import { DestinationPostTransformationService } from '../postTransformation';
-import networkHandlerFactory from '../../../adapters/networkHandlerFactory';
 
 afterEach(() => {
   jest.clearAllMocks();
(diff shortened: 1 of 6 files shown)


maheshkutty (approved):

maheshkutty:
@shekhar-rudder You can remove destinationId and other infra details as part of PR

PR #5460, opened 2026-08-12
chore: onboard customerio to live integration test

ItsSudip:

Enrolls `customerio` in the live (real-destination) integration suite —
`test/integrations/destinations/customerio/live.ts` plus a `live/` module
(`spec` / `api` / `setup` / `verify` / `profiles`). The registry and the CI matrix
discover it automatically, so no workflow changes are needed.

12 scenarios, each asserting the real delivery verdict **and** a field-level read-back
of the resulting destination state:

| Scenario | Read-back |
| --- | --- |
| identify by userId | person attributes |
| identify by email only | attributes via email → cio_id |
| identify with an email-shaped userId | attributes via email → cio_id |
| identify → track → identify (email-shaped userId) | one profile, update applied, event present |
| track / page / screen | activity by name |
| alias merge | secondary gone, primary intact |
| group object identify | object attributes + person→object relationship |
| group add_relationships | relationship exists |
| device register then delete | device token present, then absent |
| RETL (mappedToDestination) | attributes on the externalId-keyed person |

Resolves INT-6981

The mocked component suite verifies the request the transformer builds; it cannot
verify that CustomerIO accepts it. These scenarios drive seeded events through the
real `/routerTransform → /proxy` path against a live account and assert both the
delivery verdict and the resulting destination state, read back independently
through CustomerIO's App API.

The suite is written to be agnostic to which transform path runs, so it doubles as
the acceptance gate for the batching-framework (v2) rollout:

- no scenario references a v1 endpoint or the v2 batch envelope;
- each pipeline step seeds exactly one event, so `expectedOutputs` /
  `expectedProxyRequests` of 1/1 hold on both paths;
- activity read-backs match on the event **name**, never the activity `type` — v1
  records a screen as an `event`, v2 as a `screen`.

Both paths were exercised by pinning `CUSTOMERIO_BATCHING_FRAMEWORK_ENABLED_WORKSPACE_IDS`;
all 14 seeds transform to 1 output / 1 proxy request on each, and the scenarios were
run end to end against a live sandbox.

None. Test-only; no `src/` changes.

N/A

N/A — every helper is scoped to this destination's `live/` folder.

- Requires `LIVE_SECRET_CUSTOMERIO` in Vault under
  `engineering_shared/data/integrations_team/e2e_test/rudder-transformer` (single-line
  JSON). It needs **two** credentials: the Track API `siteID`/`apiKey` the transform
  delivers with, and a separate App API bearer key under `readback.appApiKey`
  (read scope: customers, activities, objects) used only by the read-backs.
- US region only; an EU account would additionally need the `track-eu` / `api-eu` hosts.
- rETL `record` events are deliberately excluded: they exist only on the v2 path, so
  they cannot hold across both.

### DIFF
--- a/test/integrations/destinations/customerio/live.ts
+++ b/test/integrations/destinations/customerio/live.ts
@@ -1,0 +1,1 @@
+export { live, default } from './live/spec';
--- a/test/integrations/destinations/customerio/live/api.ts
+++ b/test/integrations/destinations/customerio/live/api.ts
@@ -1,0 +1,279 @@
+import axios from 'axios';
+import { Agent } from 'https';
+import type { RunContext } from '../../../live/types';
+
+// keepAlive:false so read-back/cleanup sockets don't linger as open handles when the suite finishes.
+const cioAgent = new Agent({ keepAlive: false });
+const TIMEOUT_MS = 15000;
+
+// Two different APIs, two different credentials:
+//  - Track API  Basic siteID:apiKey  — what the transform delivers to. Used here only for
+//    setup/teardown side effects (create a person, delete a person/object).
+//  - App API    Bearer App API Key   — read-only; every verify reads destination state back through
+//    it, independently of the transform.
+// US hosts only: the suite runs against a US sandbox. CustomerIO's EU region uses separate hosts
+// (track-eu / api-eu), so enrolling an EU account means selecting them here too.
+const TRACK_BASE = 'https://track.customer.io';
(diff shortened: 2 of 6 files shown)


krishna2020:
https://github.com/rudderlabs/rudder-transformer/pull/5482

ItsSudip:
Already, integration tests were added through [this](https://github.com/rudderlabs/rudder-transformer/pull/5482) pr.

PR #5477, opened 2026-08-13
fix(ga4_v2): use Promise.allSettled so a single bad mapping does not drop all valid mappings

### LINKED ISSUE #5471: ga4_v2: Promise.all in handleCustomMappings drops all valid mappings when any single mapping throws
## What's happening

`handleCustomMappings` in the GA4 v2 transformer uses `Promise.all` over an array of async mapping callbacks. If any one mapping throws (invalid event name, reserved name, validation failure), `Promise.all` immediately rejects and discards the results of all other mappings. The entire event gets marked as errored even if only one of several custom mappings was bad.

## Where

`src/v0/destinations/ga4_v2/customMappingsHandler.js`, lines 110-151:

```js
const processedPayloads = await Promise.all(
  validMappings.map(async (mapping) => {
    const eventName = mapping.destEventName;
    if (isReservedEventName(eventName)) {
      throw new InstrumentationError(   // one bad mapping kills all the others
        `[GA4]:: Reserved event name: ${eventName} are not allowed`
      );
    }
    validateEventName(eventName);
    // ...
  })
);
```

## Failure scenario

Customer sets up three custom mappings for "Purchase": A → `purchase`, B → `purchase` (reserved name, typo), C → `buy_item`.

1. Mapping B throws `InstrumentationError` at line 115
2. `Promise.all` rejects immediately
3. Mappings A and C - both valid - produce no GA4 events
4. The entire event fails with an error about mapping B
5. Customer loses all GA4 purchase tracking until they find and fix mapping B

## Fix

Use `Promise.allSettled` and handle failures per-mapping:

```js
const settled = await Promise.allSettled(validMappings.map(async (mapping) => { ... }));
const processedPayloads = settled
  .filter(r => r.status === 'fulfilled')
  .map(r => r.value);
const errors = settled
  .filter(r => r.status === 'rejected')
  .map(r => r.reason);
// log errors and continue with processedPayloads
```

## Environment

rudder-transformer `main` branch (2026-08-13), Node.js.

harsh4vardhan:
## Summary

Resolves #5471.

`handleCustomMappings` uses `Promise.all` to evaluate all
`validMappings` entries concurrently. If any one mapping throws
(e.g. a reserved GA4 event name, a failed name validation, or a
bad custom-mapping expression), `Promise.all` immediately rejects
and the results of every other mapping are discarded. The entire
event fails with a single-mapping error, even when the majority of
mappings are perfectly valid.

### Change

Replace `Promise.all` with `Promise.allSettled`:

- **Fulfilled** results are collected into `processedPayloads` and
  returned as GA4 events.
- **PlatformErrors** (sandbox unavailable, fail-closed) are
  rethrown immediately so the event stays retryable. This
  preserves the existing behaviour for transient infrastructure
  failures.
- **InstrumentationError / ConfigurationError** from individual
  mappings are silently dropped, allowing the rest to proceed.
- If **all** mappings fail (and none is a PlatformError), the
  first rejection is surfaced so the customer is not left with a
  silent no-op.

### Tests

Added `ga4_custom_mapping_allsettled_all_fail`: two mappings both
using reserved GA4 event names (`first_open` and `app_update`).
The test confirms the first `InstrumentationError` is surfaced when
every mapping fails, validating the all-fail path of the new logic.

harsh4vardhan

### DIFF
--- a/src/v0/destinations/ga4_v2/customMappingsHandler.js
+++ b/src/v0/destinations/ga4_v2/customMappingsHandler.js
@@ -107,7 +107,7 @@ const handleCustomMappings = async (message, Config, workspaceId) => {
     return buildDeliverablePayload(rawPayload, Config);
   }
 
-  const processedPayloads = await Promise.all(
+  const settled = await Promise.allSettled(
     validMappings.map(async (mapping) => {
       const eventName = mapping.destEventName;
       // reserved event names are not allowed
@@ -149,6 +149,22 @@ const handleCustomMappings = async (message, Config, workspaceId) => {
       return ga4MappedPayload;
     }),
   );
+
+  const processedPayloads = [];
+  for (const result of settled) {
+    if (result.status === 'fulfilled') {
+      processedPayloads.push(result.value);
+    } else if (result.reason instanceof PlatformError) {
+      throw result.reason;
+    }
+  }
+
+  if (processedPayloads.length === 0 && settled.length > 0) {
+    const firstRejected = settled.find((r) => r.status === 'rejected');
+    if (firstRejected) {
+      throw firstRejected.reason;
+    }
+  }
 
   return processedPayloads.map((processedPayload) =>
     buildDeliverablePayload(processedPayload, Config),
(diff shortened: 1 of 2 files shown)


krishna2020:
this destination support got discontinued

PR #5482, opened 2026-08-14
feat(customerio): env flag to gate event-stream events on the V2 batching path

shekhar-rudder:
## Summary
- Enabling the batching framework for CustomerIO (`isBatchingFrameworkEnabled`) is what's needed to unlock record-event support, since the legacy `processRouterDest` has no concept of record events. But it also moves event-stream events (identify/track/page/screen/group/alias) onto the new V2 request shape/endpoint, which is a breaking change for existing customers relying on the current V1 request format.
- Adds `CUSTOMERIO_EVENT_STREAM_BATCHING_FRAMEWORK_ENABLED` (default off/disabled). While off, event-stream events keep using the legacy V1 request shape — reusing `transform.ts`'s existing per-event builder as-is — even when the batching framework is on for the workspace. When enabled, event-stream events use the new V2 code path, same as record events.
- `getBatchStrategy(endpoint)` now branches on the endpoint: the V2 batch endpoint is chunked/batched exactly as before; any legacy V1 endpoint gets a pass-through strategy (one request per event, matching `processRouterDest`'s current behaviour — CustomerIO's V1 API endpoints aren't batched today except "group" calls, which already target the same V2 batch endpoint and continue to batch as before).
- All changes are confined to `src/v0/destinations/customerio/` and its test directory — no shared/framework files were touched.

## Test plan
- [x] `npm test -- --testPathPattern=customerio --no-coverage` — 68 passed
- [x] `npm run test:ts -- component --destination=customerio` — 125 passed (added `CUSTOMERIO_EVENT_STREAM_BATCHING_FRAMEWORK_ENABLED: 'true'` to the existing V2 router test fixtures in `dataV2.ts` so they keep exercising the new code path)
- [x] New unit tests in `routerTransform.test.ts` covering: legacy V1 shape by default, V2 shape when the flag is enabled, no batching for legacy per-user endpoints, and legacy group events still batching via the shared V2 endpoint
- [x] `tsc --noEmit`, eslint, prettier all clean


src/v0/destinations/customerio/v2/config.ts
```diff
@@ -40,6 +40,17 @@ const RECORD_ACTION_MAP = {
...
 
+// Enabling the batching framework for CustomerIO (isBatchingFrameworkEnabled) is what's
+// needed to unlock record-event support, since the V1 processRouterDest has no concept of
+// record events. But it also moves event-stream events (identify/track/page/screen/group/
+// alias) onto this V2 code path, which changes their request shape/endpoint — a breaking
+// change for existing customers. This flag lets event-stream events opt into the V2 path
+// independently: while disabled (the default), they keep using the V1 processRouterDest's
+// request shape even when the batching framework is on for the workspace. See
+// routerTransform.ts's transformEventStream/getBatchStrategy.
+const isEventStreamV2APIEnabled = (): boolean =>
```
maheshkutty, line 51: To enable batching for record event type we need to set below envs `CUSTOMERIO_EVENT_STREAM_V2_API_ENABLED, CUSTOMERIO_BATCHING_FRAMEWORK_ENABLED_WORKSPACE_IDS`

maheshkutty (approved):
LGTM

PR #5483, opened 2026-08-14
fix(braze): stop router in/out mismatch from duplicated job metadata

shekhar-rudder:
## What

Fixes the P0 `router-transformer-invalid-response` alert (reason `in out mismatch`, destType `braze`) firing on `multi-tenant-eu` / `proeumt`.

## Root cause

`chunkTaggedItems` sorted tagged track items by `externalId`, then recovered per-job groups from **contiguous** runs of `sourceJobIndex`. That assumes every item of a job shares one `externalId`. It doesn't:

- `getUserAttributesObject` copies traits verbatim and `external_id` is **not** in `reservedKeys` (`transform.ts:157-196`), so a `traits.external_id` lands on the attributes item.
- When the message has no `userId`, `setExternalIdOrAliasObject` (`util.ts:1512`) takes the alias branch — it sets `user_alias` and never overwrites the attributes item's `external_id`, while the events item is built fresh and gets **no** `external_id`.

So that job's two items sort to opposite ends of the batch, form two groups, and can land in two chunks. `trackChunkResponse` then emits the job's metadata once per chunk.

rudder-server (`router/transformer/transformer.go:325-366`) builds `in` as a deduped set of input jobIds and `out` by appending once per (output element x distinct jobId), so a jobId in two outputs makes `len(out) > len(in)`. On mismatch it **discards the whole transformer response** and retries every job in the batch as a 500 (`transformer.go:368-389`), with no log line — only the stat and the job status.

## Evidence

Reporting DB (`errors` table, EU TimescaleDB) for destination `xxxxxxxx`:

```
Transformer returned invalid output size: 69 for input size: 68
Transformer returned invalid output size: 75 for input size: 74
Transformer returned invalid output size: 67 for input size: 66
```

Always over by **exactly one** — one straddling job per batch, i.e. a duplicate, not a drop.

Regression correlation: workspace `xxxxxxx` was added to `BRAZE_PER_JOB_DELIVERY_MAPPING_WORKSPACE_IDS` in rudder-devops `ffbee0a84` ("enable new batching strategy of Braze for starter customers", #7628) at 2026-08-12 10:29 UTC; the first such error for this destination is 2026-08-12 ~11:34 UTC. No occurrences before that, and no other destination affected EU-wide in 45 days. The affected code (`chunkTaggedItems` / `groupBySourceJob`) came in with `b76933ce6` (2026-07-27) and is reachable only on the flagged path.

## Fix

Group by `sourceJobIndex` first (a `Map`, so grouping no longer depends on sort order), then order whole *groups* by their leading `externalId`. Same-user locality is preserved so the V1 per-chunk `externalId` cap still bites late, but a job's items can no longer be split across chunks whatever external_ids they carry.

Because `fits()` is evaluated against a whole group and `addGroupToChunk` commits the whole group, a chunk boundary can now only fall *between* jobs. A chunk may therefore close under the cap (e.g. at 74 of 75) rather than split the job that overflows it.

## Tests

Four new tests in `braze.util.test.ts`, in two blocks. All four fail on the parent commit and pass on the fix.

`router in/out job accounting (ON path)`
- a job whose items carry different `external_id`s is not emitted in two chunks;
- out-jobId count equals in-jobId count, mirroring rudder-server's exact counting rule. Pre-fix: `Expected: 81, Received: 82` — the production symptom reproduced.

`chunk boundaries never split a source job`
- a chunk closes under the per-type cap rather than splitting the overflowing job. Pre-fix this finds a 75-attribute chunk, i.e. the job really was split across the boundary;
- a deterministic randomised sweep (120 jobs, mixed and absent `external_id`s) asserts no jobId is ever emitted twice. Pre-fix a jobId appears twice.

Verification (re-run after merging `develop`):
- `npx jest src/v0/destinations/braze` — 262 passed (6 suites)
- Braze component tests — 108 passed
- `tsc --noEmit -p tsconfig.test.json` clean (the root `tsconfig.json` excludes `*.test.ts`, so it does not cover this file); prettier clean

The counts are higher than at first review because `develop` has since gained #5488's `validateDestinationConfig` tests; nothing in this branch changed.

## Packing cost

Group-atomic packing can leave slots unused, so this was measured against the old code on identical workloads (request counts out of `processBatchWithDeliveryMapping`):

| jobs/batch | mixed-id jobs | old | new | extra |
|---|---|---|---|---|
| 74 | 0-100% | 2 | 2 | +0 |
| 300 | 0-50% | 9 | 9 | +0 |
| 300 | 100% | 8 | 9 | +1 |
| 1000 | 1-100% | 27 | 28 | +1 |
| 5000 | 50-100% | 134 | 136 | +2 |

The cost is a small absolute number of extra requests, not a proportional one, so it shrinks as batches grow. At this destination's actual batch sizes (66-75 jobs) it is zero. The bound is `(max group size - 1)` unused slots per chunk — 1 of 75 (1.33%) for the standard attributes+events shape. Where all of a job's items share one `external_id` the two algorithms produce identical output, since the old sort already made them adjacent.

Separately: the group-preserving chunker has real overhead versus theoretically optimal packing for jobs with many purchases (up to ~2x at ~38 items/job). That is pre-existing from `b76933ce6` — an identical sweep over group sizes k=1..74 gives byte-identical counts on both versions.

## Notes

- Sub/merge chunking uses positional `_.chunk`, which would have the same straddle hazard — but each message yields exactly one subscription group (`transform.ts:462`) or one merge update, so it is not reachable today. Left unchanged deliberately.
- The stale `debug/braze-in-out-mismatch` branch targets *dropped* jobs (`processBatch`'s missing `else`); that is a different, real defect but not the cause of this alert, which is a duplicate.
- Third commit is test-only cleanup: removes `any` from the blocks added here (real types, a `jobId` type-predicate narrowing, and dropping two `as any` casts that turned out to be unnecessary). The file's pre-existing `any` usages are untouched.
- Fourth commit merges `develop`. The only conflict was an import line in `braze.util.test.ts`: #5488 added `ConfigurationError` where this branch had added `Metadata`; both are kept. `util.ts` auto-merged — #5488's `restApiKey` guard and this PR's chunker sit in disjoint regions of the file, and the diff of `util.ts` against `develop` is still exactly the `groupBySourceJob`/`chunkTaggedItems` change.

### DIFF
--- a/src/v0/destinations/braze/braze.util.test.ts
+++ b/src/v0/destinations/braze/braze.util.test.ts
@@ -2911,3 +2911,129 @@ describe('router in/out job accounting (ON path)', () => {
     expect(outJobIdCount(result)).toBe(jobs.length);
   });
 });
+
+// ---------------------------------------------------------------------------
+// Chunk-boundary invariant: chunking packs whole source-job groups, never
+// individual items, so a boundary can only fall between two jobs. A chunk may
+// therefore close under the per-type cap rather than split a job.
+// ---------------------------------------------------------------------------
+
+describe('chunk boundaries never split a source job', () => {
+  const destination = brazeDestFor();
+
+  type ItemSpec = { type: 'attributes' | 'events' | 'purchases'; externalId?: string };
+
+  const jobFromSpecs = (jobId: number, specs: ItemSpec[]): BrazeTransformedEvent => {
+    const bucket: Record<string, any[]> = { attributes: [], events: [], purchases: [] };
+    specs.forEach((spec, k) => {
+      const base: Record<string, unknown> = { tag: `${jobId}-${k}` };
+      if (spec.externalId !== undefined) {
+        base.external_id = spec.externalId;
+      }
+      if (spec.type === 'events') {
+        bucket.events.push({ ...base, name: 'e', time: 't' });
+      } else if (spec.type === 'purchases') {
+        bucket.purchases.push({
+          ...base,
+          product_id: `p${jobId}-${k}`,
+          price: 1,
+          currency: 'USD',
+          quantity: 1,
+          time: 't',
+        });
+      } else {
+        bucket.attributes.push(base);
+      }
+    });
+    return {
+      destination,
+      statusCode: 200,
+      batchedRequest: {
+        version: '1',
+        type: 'REST',
+        method: 'POST',
+        endpoint: '',
+        headers: {},
+        params: {},
+        body: {
+          JSON: {
+            attributes: bucket.attributes,
+            events: bucket.events,
+            purchases: bucket.purchases,
+          },
+        },
+        files: {},
+      } as any,
+      metadata: [{ jobId, workspaceId: 'workspace-non-mau' }],
+    };
+  };
+
+  const jobIdOccurrences = (result: any[]): Map<number, number> => {
+    const counts = new Map<number, number>();
+    for (const out of result) {
+      for (const jobId of new Set((out.metadata ?? []).map((m: any) => m.jobId))) {
+        counts.set(jobId as number, (counts.get(jobId as number) ?? 0) + 1);
+      }
+    }
+    return counts;
+  };
+
+  test('closes a chunk under the per-type cap rather than splitting the job that overflows it', () => {
+    // 37 two-attribute jobs fill 74 of the 75 attribute slots. The 38th job also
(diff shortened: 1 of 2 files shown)


saikumarrs (dismissed):
@shekhar-rudder let's remove the workspace ID from the PR description.

ItsSudip (dismissed):

ItsSudip (approved):

PR #5484, opened 2026-08-17
test(customerio): assert destination state in the live spec, not just 2xx

shekhar-rudder:
Stacked on #5482. Answers "are the live tests just making API calls, or should they verify?" — they were only asserting that CustomerIO returned 2xx.

Both rollout scenarios seeded identical events and asserted identically, so nothing in the suite could tell the two states of `CUSTOMERIO_EVENT_STREAM_V2_API_ENABLED` apart. The one claim the flag makes was the one thing untested.

## What's added

Restructured into the layout `live/README.md` prescribes (`live/{spec,api,verify,profiles}.ts`, `live.ts` as a re-export), plus App API read-backs:

- identify traits land on the profile
- track / page / screen activities exist, with their properties
- the device-related track event registers a device (previously asserted by nothing)
- the group object is **linked** to the person, and carries its traits
- the record insert creates a profile with its identifiers
- the alias merge actually combines the two profiles
- **cross-scenario parity** — both flag states must produce the same observable CustomerIO state

Read-back needs an App API key (Bearer, `api.customer.io`); the Track credentials the transform delivers with are write-only. Read from `LIVE_SECRET_CUSTOMERIO.readback.appApiKey`, which already exists on the shared Vault path. Absent, it fails loudly rather than skipping.

## ⚠️ Needs a product decision: screen events change type when the flag flips

Confirmed live, and isolated by mutation-testing the parity check — the two states are identical except:

```
- event:Viewed Home Screen      (flag OFF)
+ screen:Viewed Home Screen     (flag ON)
```

`transform.ts:107` sets `evType = 'event'` for `EventType.SCREEN`; `v2/util.ts:137` sends `action: 'screen'`. With the flag off the profile has no `screen` activity at all.

For a customer enabling the flag, any CustomerIO segment, campaign trigger or report matching "Viewed X Screen" as an event stops matching, and vice versa. That's a migration hazard rather than a transport detail. It's normalised as the single tolerated exception in the parity check so every *other* difference still fails — but whether the divergence itself is acceptable isn't a test decision.

## Two bugs the read-backs caught

1. **The alias step never exercised a merge.** CustomerIO's merge keeps the primary and folds the secondary into it, but nothing created the primary — so the merge was a no-op returning 2xx. Fixed by identifying the merge primary first (with disjoint traits, so the read-back can prove the profiles combined) and settling before the merge via `delayBeforeMs`: a merge whose primary hasn't materialised is silently dropped.

2. **The live runner reported skipped steps as passed.** `component.live.test.ts` returned early on short-circuit, which jest records as ✓ — so after any failure every later step *and* the scenario read-back showed green without running (6 phantom passes in the first run). They now fail with an explanatory message.

## Pinned by live probes, not assumed

- `page` activities carry no `name` — CustomerIO discards it, so a page's name isn't observable on either path
- activity `data` preserves types (`revenue: 42` stays a number); person attributes are stringified
- `object_type_id` comes back as a string despite the OpenAPI spec typing it an integer
- merge moves the secondary's attributes onto the primary, conflicts keep the primary's, secondary → 404

## Verification

- Live suite: **24/24, exit 0** (both scenarios, real CustomerIO account)
- Parity assertion mutation-tested — fails with a precise A/B diff when the states differ
- Mocked component + customerio unit suites: **3997 passed, 2 skipped, 0 failed**
- `tsc --noEmit -p tsconfig.test.json` clean; prettier clean
- `/security-review`: no findings — the App API key is never interpolated into errors or logs, path params are encoded, fixtures use `maskedSecrets`


shekhar-rudder:
Superseded — folding this into #5482 directly as a commit on its branch rather than keeping a stacked PR.

PR #5488, opened 2026-08-17
fix(braze): fail fast when restApiKey is missing instead of sending Bearer undefined

shekhar-rudder:
## What

Braze never validated `restApiKey` before putting it on the wire. When the key is absent, `` `Bearer ${destination.Config.restApiKey}` `` stringifies to the literal **`Bearer undefined`**. This adds a single guard that throws `ConfigurationError` instead.

## Why

Braze echoes back the credential it received, so a missing key produces:

```
401 {"message":"Invalid API key: undefined"}
```

That reads like a rotated or revoked *customer* credential and sends the responder to Braze's status page — not to our own config plumbing. It is genuinely misleading: in a recent P1 the initial triage concluded "Braze-side auth incident across 6 destinations" when the real cause was internal and the key had simply not reached the data plane.

It also costs real data. 4xx is terminal in rudder-server — `isJobTerminated` (`router/misc.go`) returns true for everything 200–499 except 429/408 — so **every event in flight is aborted with no retry**.

## Change

`validateDestinationConfig(destination)` in `util.ts` throws `ConfigurationError('Rest API Key not found. Aborting')` when `Config.restApiKey` is missing or empty. It runs once at each transformation entry point:

| entry point | note |
|---|---|
| `process` | processor path |
| `processRouterDest` | runs **before** the dedup lookup, which would otherwise spend a Braze round trip on a request that cannot authenticate |

`process` and `processRouterDest` are the only exports of `transform.ts`, so every path into the destination is covered. The remaining `Bearer ${destination.Config.restApiKey}` interpolations (`transform.ts`, `util.ts`, `identityResolutionUtils.ts`) are left untouched and assume a valid key, per review feedback.

A throw from `processRouterDest` is caught in `nativeIntegration.ts:149` and mapped across the whole `destInputArray`, so every event for that destination gets the 400 — which is correct, since none of them could have been delivered.

The message follows the existing convention: ga4 throws `'API Secret not found. Aborting'`, facebook_pixel `'Access token not found. Aborting'`. `deleteUsers.ts` already guarded this (it takes `config` rather than `destination`, so it keeps its own check) and `braze_audience` validates via zod — this closes the remaining delivery path.

## Testing

- `validateDestinationConfig` unit tests (`it.each`): key absent / empty string / `undefined` / `Config` absent → throws `ConfigurationError`; key present → does not throw.
- Component test: a destination with no `restApiKey` now returns `400 'Rest API Key not found. Aborting'` with `errorType: 'configuration'` instead of reaching Braze.
- **258/258** braze unit tests, **98/98** braze component tests, `tsc --noEmit` clean, eslint clean (0 errors).
- All 97 pre-existing component tests still pass unchanged, confirming no existing fixture depended on a missing key.

## Note for reviewers

This changes the observable failure for an unconfigured destination from a **delivery-stage 401** to a **transformation-stage 400**. Same terminal outcome (the events were already 100% lost), but dashboards or alerts keyed on the Braze 401 will see it move. That is the intended effect — the failure is now attributed to our config rather than to Braze.

Paired with rudderlabs/rudder-integrations-config#2645, which fixes the schema hole that let a Braze destination be saved without a `restApiKey` in the first place (`connectionMode` `cloud`/`warehouse` were missing from the `anyOf` gating the requirement).

### DIFF
--- a/src/v0/destinations/braze/transform.ts
+++ b/src/v0/destinations/braze/transform.ts
@@ -19,6 +19,7 @@ import {
   handleReservedProperties,
   getEndpointFromConfig,
   formatGender,
+  validateDestinationConfig,
 } from './util';
 import type {
   BrazeDestination,
@@ -534,6 +535,7 @@ async function process(
 ): Promise<ProcessorTransformationOutput> {
   let response;
   const { message, destination } = event;
+  validateDestinationConfig(destination);
   const messageType = message.type.toLowerCase();
 
   let category = ConfigCategory.DEFAULT;
@@ -618,6 +620,9 @@ const processRouterDest = async (
   const userStore = new Map<string, BrazeUser>();
   let failedLookupIdentifiers = new Set<string>();
   const { destination } = inputs[0];
+  // Checked before the dedup lookup below, which would otherwise spend a Braze
+  // round trip on a request that cannot authenticate.
+  validateDestinationConfig(destination);
   if (destination.Config.supportDedup) {
     let lookupResult: { users: BrazeUser[]; failedIdentifiers: Set<string> } | undefined;
     try {
--- a/src/v0/destinations/braze/transform.ts
+++ b/src/v0/destinations/braze/transform.ts
@@ -19,7 +19,7 @@ import {
   handleReservedProperties,
   getEndpointFromConfig,
   formatGender,
-  getAuthHeader,
+  validateDestinationConfig,
 } from './util';
 import type {
   BrazeDestination,
@@ -86,7 +86,7 @@ function buildResponse(
     headers: {
       'Content-Type': JSON_MIME_TYPE,
       Accept: JSON_MIME_TYPE,
-      Authorization: getAuthHeader(destination),
+      Authorization: `Bearer ${destination.Config.restApiKey}`,
     },
     userId: message.userId || message.anonymousId,
   };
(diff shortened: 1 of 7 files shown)


src/v0/destinations/braze/transform.ts
```diff
@@ -85,7 +86,7 @@ function buildResponse(
     headers: {
       'Content-Type': JSON_MIME_TYPE,
       Accept: JSON_MIME_TYPE,
-      Authorization: `Bearer ${destination.Config.restApiKey}`,
+      Authorization: getAuthHeader(destination),
```
saikumarrs: Since we require the auth header for successful delivery, how about validating it at the transformation method entry level and rest of the implementation assumes `destination.Config.restApiKey` is valid?

shekhar-rudder: Done in 7df5f81ad — reverted the `getAuthHeader` refactor.

Validation is now a single `validateDestinationConfig(destination)` at the two transformation entry points (`process`, `processRouterDest` — the only exports of `transform.ts`), and every `Bearer ${destination.Config.restApiKey}` interpolation is back to what it was, assuming a valid key.

In `processRouterDest` the check sits before the dedup lookup, so an unconfigured destination no longer spends a Braze round trip on a request that cannot authenticate.

saikumarrs (approved):

PR #5490, opened 2026-08-18
fix(braze): scope 296 to ecommerce schema rejections on /users/track

ItsSudip:

The v1 Braze networkHandler emitted `296` (delivered-with-warning) for **any** per-item error it
could correlate back to a job on a `/users/track` partial-failure response. This narrows it.

A correlated hit now warrants 296 only when all three hold:

1. `errors[i].input_array === 'events'`
2. the item we sent at `errors[i].index` is a recommended-ecommerce event (matched on its Braze event name)
3. `errors[i].type` is a JSON-pointer schema rejection (`/^The property '#\//`)

| Correlated hits | Status |
| --- | --- |
| none | 200, full response body |
| all qualify | 296, joined `error.type` strings |
| any does not | 400, joined `error.type` strings |

Abort outranks warning on a mixed job; `error` still carries every hit, joined with `; `.

Also fixes `braze_partial_failure`, which was registered in `prometheus.js` with `labelNames: []`.
prom-client throws on undeclared labels and the stats wrapper swallows the throw into an error log,
so that counter has never incremented at all. It now declares `destinationId`/`workspaceId`.

Resolves INT-7010

Braze validates recommended-ecommerce events against its ecommerce event schema and rejects
individual items:

```
The property '#/' contains additional properties ["sku_abcd"] outside of the schema when none are allowed
The property '#/product_id' of type integer did not match the following type: string
The property '#/' did not contain a required property of 'product_id'
The property '#/price' of type string did not match the following type: number
```

For these Braze accepts the batch and drops only the offending item — a genuine
delivered-with-warning. Every other correlated failure (missing identifier, bad attribute, invalid
purchase) is a real drop and should abort the job so it surfaces as a failure.

The event-name check is what makes this precise: `events[]` is not purely ecommerce. Unmapped track
events fall through to the legacy path, and chunking merges several jobs' events into one array — so
gating on the `useEcommerceRecommendedEvents` config alone would mis-classify legacy events as
warnings.

Two:

1. **Jobs previously receiving 296 for non-ecommerce failures now abort with 400.** This covers
   `'external_id' is required` and every `attributes`/`purchases` failure. Reason: those are
   permanent per-item drops, and reporting them as delivered-with-warning hid real data loss.
2. **`braze_delivered_with_warning` label names move from snake_case to camelCase**
   (`destinationId`/`workspaceId`), matching the neighbouring `braze_audience_*` metrics. Existing
   queries on that metric need updating. `braze_partial_failure` never emitted and
   `braze_delivery_aborted` is new, so neither has dependants.

N/A

- `isBrazeEcommerceEventName` (`src/v0/destinations/braze/ecommerceUtil.ts`) — derived from the
  existing `ECOMMERCE_ROUTED_ENTRIES`, so adding an event to `ConfigCategory` extends it for free.
- `BrazeProxyV1Request` (`src/v0/destinations/braze/types.ts`) — narrows the proxy request's
  `body.JSON` to `BrazeTrackRequestBody`. Braze-local, not a framework change.
- `src/util/prometheus.js` is shared, but the edit is confined to the `braze_partial_failure`
  entry's `labelNames`.

Correlation reads `body.JSON.events[index]` off the proxy request, relying on the same positional
alignment that `destInfo` correlation already assumes. If alignment ever failed, the item would read
as non-ecommerce and the job would abort with 400 rather than be mis-attributed a 296 — the safe
failure direction.

The body is unvalidated wire input, so runtime guards stay in place (`Array.isArray` on `events`,
`typeof name === 'string'` inside the predicate). `BrazeProxyV1Request` documents intent; it does
not prove shape.

<hr>

### Developer checklist

### DIFF
--- a/src/v1/destinations/braze/networkHandler.ts
+++ b/src/v1/destinations/braze/networkHandler.ts
@@ -5,19 +5,26 @@ import { TAG_NAMES } from '../../../v0/util/tags';
 import { isHttpStatusSuccess } from '../../../v0/util/index';
 import { HTTP_STATUS_CODES } from '../../../v0/util/constant';
 import stats from '../../../util/stats';
-import type {
-  DeliveryJobState,
-  DeliveryV1Response,
-  ProxyMetdata,
-  ProxyV1Request,
-} from '../../../types';
-import { BrazeError, BrazeResponseHandlerParams } from '../../../v0/destinations/braze/types';
+import type { DeliveryJobState, DeliveryV1Response, ProxyMetdata } from '../../../types';
+import {
+  BrazeError,
+  BrazeEvent,
+  BrazeProxyV1Request,
+  BrazeResponseHandlerParams,
+} from '../../../v0/destinations/braze/types';
+import { isBrazeEcommerceEventName } from '../../../v0/destinations/braze/ecommerceUtil';
 
 const DESTINATION = 'braze';
 
 const SUCCESS_MESSAGE = `Request for ${DESTINATION} Processed Successfully`;
 const failureMessage = (status: number): string =>
   `Request failed for ${DESTINATION} with status: ${status}`;
+
+// Ops-facing metric names — kept as constants so dashboards and alerts have a
+// single definition site to grep for.
+const METRIC_PARTIAL_FAILURE = 'braze_partial_failure';
+const METRIC_DELIVERED_WITH_WARNING = 'braze_delivered_with_warning';
+const METRIC_DELIVERY_ABORTED = 'braze_delivery_aborted';
 
 // Braze's `endpointPath` value for the /users/track endpoint — the only
 // endpoint whose response carries per-entry `errors[]` correlatable back to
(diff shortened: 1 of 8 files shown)


src/v1/destinations/braze/networkHandler.ts
```diff
@@ -103,56 +120,107 @@ const buildWarnedIndexMaps = (errors: BrazeError[]): Record<TrackInputArray, War
...
+  // The body is an unvalidated echo of what we sent, so the declared type is a
+  // statement of intent rather than a guarantee — check before trusting it.
+  return Array.isArray(events) ? events : [];
+};
+
+// True when the item we sent at `index` was built by the recommended-ecommerce
+// path, identified by its Braze event name — only that path emits those names,
+// so it separates ecommerce events from the legacy custom events that share the
+// same `events[]` array after chunking.
+const isEcommerceEventAt = (sentEvents: BrazeEvent[], index: number): boolean =>
```
shekhar-rudder: Lets inline `isEcommerceEventAt`

shekhar-rudder (dismissed):

shekhar-rudder (approved):

PR #5492, opened 2026-08-18
feat(google_adwords_offline_conversions): surface per-conversion partial-failure details

shekhar-rudder:

When Google Ads rejects some conversions in a batch it replies **HTTP 200** with a `partialFailureError` (a `google.rpc.Status`). Its `details[].errors[]` carry, per conversion: the specific `errorCode` enum, that conversion's own `message`, the `location` of the offending field, and the `requestId`.

The v1 GAOC response handler used only `partialFailureError.message` — which by construction describes just the **first** error — and stamped that one string onto **every** failed event in the batch. Everything else was dropped before it reached error reporting.

This PR parses the structured details and gives each failed event its own error:

```
before: Multiple errors in 'details'. First error: An internal error has occurred., at conversions[15]
after:  An internal error has occurred. [internalError: INTERNAL_ERROR] at conversions[15] (requestId: f4J_sjHfhbgNieU4pkBOqg)
```

New shared util `src/v0/util/googleUtils/partialFailure.ts` (`parsePartialFailure` / `formatGoogleAdsErrors`), wired into the GAOC v1 `networkHandler`.

N/A — came out of a support investigation into `An internal error has occurred., at conversions[N]` on GOOGLE_ADWORDS_OFFLINE_CONVERSIONS.

That error is currently **225,692 events across 92 destinations / 67 workspaces in 14 days**, and its rate co-moves across unrelated Google Ads accounts (0.51% → 6.82% of delivery attempts while traffic stayed flat), which points at Google's side rather than customer payloads.

We could not confirm that from our own data, because we discard exactly the fields needed to:

1. **Tell which documented error it is.** Google documents `InternalErrorEnum.InternalError`; `INTERNAL_ERROR` and `TRANSIENT_ERROR` are both explicitly retryable ([Error Types](https://developers.google.com/google-ads/api/docs/best-practices/error-types)). Without `errorCode` we cannot tell which we are being sent — and therefore cannot justify a retry policy.
2. **Escalate to Google.** Google asks for the request id when reporting an issue; we were dropping it.
3. **Tell events apart.** Every failed event in a batch got an identical message naming one conversion's index, so a customer debugging event #15 saw an error about event #1.

**Diagnostic only — delivery semantics are unchanged.** `statusCode` (400/200 per event), `status`, `statTags` and the success path are byte-for-byte identical. Only the per-event `error` **string** changes. Making these failures retryable is deliberately **not** in this PR: it would change delivery behaviour for 92 destinations and deserves its own review.

Two proxy fixtures were updated for the richer strings. The raw Google response echoed under `destinationResponse` is deliberately left untouched.

N/A

New `src/v0/util/googleUtils/partialFailure.ts`. It is destination-agnostic (`operationField` is a parameter, defaulting to `conversions`) so `google_adwords_enhanced_conversions`, which has the identical gap with `conversion_adjustments`, can adopt it without a move. Only GAOC is wired here.

Three correctness guards worth reviewer attention, all added in response to review:

- **Batch misalignment.** `combineBatches` (`src/v0/util/index.js`) can merge two differently sized requests under one deduped, re-sorted metadata array, so `metaDataArray[i]` does not always correspond to Google's `conversions[i]`. Positional attribution is therefore applied **only** when `results.length === metaDataArray.length`; otherwise we fall back to request-wide errors so no event is blamed for another's failure. (The pre-existing `results?.[i]` statusCode mapping has the same exposure; this PR does not widen it.)
- **Bounded output.** Request-wide errors are replayed onto every failed event, which is O(events x errors); with `MAX_CONVERSIONS_PER_BATCH = 2000` that is a plausible response blow-up. Capped at `MAX_ERRORS_PER_EVENT = 3` (`(+N more)`) and `MAX_ERROR_LENGTH = 1024`, with the request id appended after truncation so it always survives.
- **Index resolution.** Only the outermost path element is used as the operation index, so `conversions.user_identifiers[2]` is not misread as conversion 2.

**PII:** Google's `trigger` field echoes the value that caused the error (can be a hashed email or click id) and is deliberately **never** put into the per-event `error` string — only `message`, `errorCode`, field path and `requestId`. Verified end-to-end: a mock response carrying a sentinel `trigger` value produced error strings with no trace of it. Note for precision that `trigger` *is* still present in the raw `destinationResponse` echoed alongside the response — that is **pre-existing** behaviour, identical on the old image, and untouched here.

### E2E verification against the live Google Ads API

Run on the local `mini` stack with the transformer built from this branch, proxying to the **real** Google Ads v23 API (customer `4219454086`, real OAuth account), and compared against the stock `v1.133.2` image with the byte-identical request.

Batch of two conversions, both crafted to fail so **nothing is written to the account** (Google returned `results: [{}, {}]`, confirmed): `conversions[0]` has a valid `UPLOAD_CLICKS` conversion action but an unparseable gclid; `conversions[1]` is identifier-only.

Google's real `partialFailureError.message` names only the first error, the gclid one. Result:

```
OLD  job 201 [400] Multiple errors in 'details'. First error: The imported gclid could not be decoded...
OLD  job 202 [400] Multiple errors in 'details'. First error: The imported gclid could not be decoded...
                   ^ job 202 has NO gclid at all -- it is told to fix a field it never sent

NEW  job 201 [400] The imported gclid could not be decoded... [conversionUploadError: UNPARSEABLE_GCLID]
                   at conversions[0].gclid (requestId: DJDJHdnP4xEeAU-CvwA6zg)
NEW  job 202 [400] Make sure you've turned on enhanced conversions for leads...
                   [conversionUploadError: CUSTOMER_NOT_ENABLED_ENHANCED_CONVERSIONS_FOR_LEADS]
                   at conversions[1].user_identifiers (requestId: DJDJHdnP4xEeAU-CvwA6zg)
```

Verified in the same live run:

- **Delivery semantics unchanged** — `statusCode` identical on both images (`[400, 400]`).
- **PII holds against real data** — Google really did return a `trigger` (`{"stringValue": "THIS-IS-NOT-A-REAL-GCLID"}`); that value appears nowhere in the per-event error strings.
- **Misalignment guard** — replaying with 1 metadata entry against 2 conversions correctly fell back to the summary instead of making a wrong positional claim.
- **Cap** — 10 request-wide errors render as 3 + `(+7 more)`, 252 chars, request id preserved.

**Memory:** previously all failed events in a batch shared one string *reference*; they now each hold their own. Worst case with the caps is ~1.1 KB/event, so ~2.2 MB for a 2000-event batch — bounded, but not free.

## Known review findings (not fixed here)

Surfaced by review and consciously left out of scope:

1. **The alignment guard detects a size mismatch, not a reordering.** `removeDuplicateMetadata` sorts metadata by `jobId` while `conversions` keep router order, so if router input is ever not jobId-ascending, lengths can match while positions do not. Fully closing this needs a per-index jobId cross-check, which the Google response does not give us. The guard is a floor, not a guarantee, and the code comment says so.
2. **`statusCode` is still assigned positionally even when the guard trips.** `results?.[i]` decides 400 vs 200 regardless of alignment, so in a merged batch the wrong event can still be marked failed. This is **pre-existing** and unchanged by this PR — the new guard just makes it visible. It is the more consequential half (abort vs deliver) and probably deserves its own fix.
3. **`INTERNAL_ERROR`/`TRANSIENT_ERROR` are still aborted as 400 and never retried**, despite Google documenting them as retryable. That is the actual data-loss bug behind the 225K events; deliberately excluded here because it changes delivery behaviour fleet-wide.

### Developer checklist

### DIFF
--- a/src/v0/util/googleUtils/partialFailure.test.ts
+++ b/src/v0/util/googleUtils/partialFailure.test.ts
@@ -1,14 +1,14 @@
-import {
-  formatGoogleAdsErrors,
-  getErrorCodeLabel,
-  getFieldPath,
-  getOperationIndex,
-  parsePartialFailure,
-  GoogleAdsError,
-  PartialFailureError,
-} from './partialFailure';
+import { formatGoogleAdsErrors, parsePartialFailure } from './partialFailure';
 
-const failureWith = (errors: GoogleAdsError[], requestId?: string): PartialFailureError => ({
+// Shapes are declared locally rather than imported: the module exports only the two functions
+// under test, and nothing should be exported purely to give this file a handle on it.
+type TestError = {
+  errorCode?: Record<string, string>;
+  message?: string;
+  location?: { fieldPathElements?: { fieldName?: string; index?: number }[] };
+};
+
+const failureWith = (errors: TestError[], requestId?: string) => ({
   code: 3,
   message: 'summary message',
   details: [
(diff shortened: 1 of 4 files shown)


maheshkutty (approved):
LGTM

PR #5495, opened 2026-08-19
fix(braze): reject malformed email at transform time

shekhar-rudder:
## What

Braze answers a malformed address with:

```
"type": "The value provided for the email field is not a valid email."
```

…at **delivery** time, so it lands as a delivery failure rather than something the customer can spot in their event stream.

This adds `formatEmail` in `src/v0/destinations/braze/util.ts` and calls it from the `case 'email'` branch in `getUserAttributesObject` (`transform.ts:178`) — the single choke point for `user_attributes.email` on **both** the identify and track paths (called at `transform.ts:279` and `:338`). Malformed addresses now abort with an `InstrumentationError` at transform time and never reach delivery.

Validation uses `validator.isEmail`, already a direct dependency (`package.json:128`) and already the convention in `clickup/util.js:48` and `sendinblue/util.js:38`.

## Behaviour

| Input | Before | After |
|---|---|---|
| `USER@Example.com` | lowercased, sent | unchanged — lowercased, sent |
| `not-an-email` | sent, Braze rejects on delivery | aborted, 400 instrumentation error |
| `123` / `{}` / `[]` | aborted (`email must be a valid string`) | unchanged — aborted |
| `null` | passed through (unsets field in Braze) | unchanged — passed through |
| `undefined` | passed through | unchanged — passed through |

The offending address is deliberately **kept out of the error message**: the full payload is already visible alongside the error in Live Events, and the message itself ends up in logs and metrics where the PII does not belong.

## Scope — deliberately narrow

Only `user_attributes.email` is validated. Two other paths can still carry an unvalidated address and were **left alone on purpose**:

- `processGroup`'s `subscriptionGroup.emails` (`transform.ts:456-460`), only live when `enableSubscriptionGroupInGroupCall` is on.
- The RETL/VDM path, which returns traits as-is at `transform.ts:150-152` and skips field handling entirely.

Happy to extend to either in a follow-up.

## ⚠️ Rollout risk worth a second opinion

Throwing aborts the **whole event**, so an identify carrying a good `userId` and ten good traits is now dropped for one malformed email. Previously Braze's `/users/track` returned `201` with a per-field `errors` array and **still applied the other attributes** — so for those events this trades a partial success for a full drop.

That is the intended behaviour ("never reaches delivery"), but it is a real change for any workspace currently sending malformed addresses at volume. Worth a look at Braze error rates before merging if you want to size the blast radius.

## Testing

- **Unit** (`braze.util.test.ts`): 22 new `formatEmail` cases — 6 valid formats (plus-addressing, subdomains, dotted local parts, hyphenated domains), 9 invalid, 4 non-string types, null/undefined passthrough, and a guard that the address does not leak into the error message. `283 passed`.
- **Integration** (`processor/data.ts`): 3 new cases — identify aborts, track aborts (proves the shared user-attributes path is covered), and a valid uppercase email still succeeds and lowercases, guarding against over-rejection. `101 passed`.
- `npx tsc --noEmit` and `eslint` clean.
- Checked `validator.isEmail` for ReDoS: sub-millisecond on 50KB pathological inputs (it short-circuits on length).

### DIFF
--- a/src/v0/destinations/braze/util.ts
+++ b/src/v0/destinations/braze/util.ts
@@ -102,6 +102,17 @@ const formatGender = (gender: unknown) => {
  * The offending address is deliberately kept out of the error message; the full payload is
  * already visible alongside the error in Live Events, and the message itself ends up in logs
  * and metrics where the PII does not belong.
+ *
+ * `blacklisted_chars: '"'` narrows `isEmail` to Braze's rule that the local part "cannot
+ * contain double quotes" -- `"` is absent from the character set in Braze's own published
+ * validation regex. RFC-legal quoted addresses such as `"user name"@example.com` pass
+ * `isEmail`'s default options, but Braze rejects them.
+ *
+ * The option only ever rejects addresses Braze also rejects: `isEmail` admits `"` in the
+ * local part solely via its fully-quoted branch, and a local part opening with `"` can never
+ * match Braze's regex. Note Braze validates only the segment preceding `+`, so widening the
+ * blacklist further would over-reject -- e.g. Braze accepts `user+{tag}@example.com`.
+ * https://www.braze.com/docs/user_guide/channels/email/email_setup/email_validation
  */
 const formatEmail = (email: unknown) => {
   if (!isDefinedAndNotNull(email)) {
@@ -113,7 +124,7 @@ const formatEmail = (email: unknown) => {
   }
 
   const formattedEmail = email.toLowerCase();
-  if (!validator.isEmail(formattedEmail)) {
+  if (!validator.isEmail(formattedEmail, { blacklisted_chars: '"' })) {
     throw new InstrumentationError(
       'Invalid email, the email provided is not a valid email address',
     );
--- a/src/v0/destinations/braze/util.ts
+++ b/src/v0/destinations/braze/util.ts
@@ -97,14 +97,15 @@ const formatGender = (gender: unknown) => {
  * customer can spot in their event stream. Validating here turns it into an instrumentation
  * error at transform time, so the bad event never reaches delivery.
  *
- * `null` is passed through untouched — Braze reads an explicit null as "unset this field".
+ * `null`/`undefined` pass through untouched — Braze reads an explicit null as "unset this
+ * field", and the caller's own guard already decides which of the two reach here.
  * The offending address is deliberately kept out of the error message; the full payload is
  * already visible alongside the error in Live Events, and the message itself ends up in logs
  * and metrics where the PII does not belong.
  */
 const formatEmail = (email: unknown) => {
-  if (email === null) {
-    return null;
+  if (!isDefinedAndNotNull(email)) {
+    return email;
   }
 
   if (typeof email !== 'string') {
--- a/src/v0/destinations/braze/util.ts
+++ b/src/v0/destinations/braze/util.ts
@@ -1,6 +1,7 @@
 /* eslint-disable no-param-reassign, @typescript-eslint/naming-convention */
 import _ from 'lodash';
 import get from 'get-value';
+import validator from 'validator';
 import { ConfigurationError, InstrumentationError, isDefined } from '@rudderstack/integrations-lib';
 import stats from '../../../util/stats';
 import logger from '../../../logger';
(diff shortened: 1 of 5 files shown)


src/v0/destinations/braze/util.ts
```diff
@@ -88,6 +89,39 @@ const formatGender = (gender: unknown) => {
...
+  if (!isDefinedAndNotNull(email)) {
+    return email;
+  }
+
+  if (typeof email !== 'string') {
+    throw new InstrumentationError('Invalid email, email must be a valid string');
+  }
+
+  const formattedEmail = email.toLowerCase();
+  if (!validator.isEmail(formattedEmail)) {
```
ItsSudip: `validator.isEmail` with default options is looser than Braze's documented rules in one cheaply-fixable way: it accepts quoted local parts (`"user name"@example.com` → `true` on validator 13.15.22), while the [Braze doc](https://www.braze.com/docs/user_guide/channels/email/email_setup/email_validation#email-validation) says the local part "cannot contain double quotes".

```suggestion
  if (!validator.isEmail(formattedEmail, { blacklisted_chars: '"' })) {
```

Please also add a quoted-address case to the invalid list in the tests.

The other gaps vs the doc don't seem worth closing here — these all pass `validator.isEmail` but are rejected per Braze's doc, and will still fail at delivery time (same as today):

- Unicode in the domain: `hans@m端ller.com` — Braze accepts Unicode "only for the local part"; the domain must be Punycode
- Gmail local part shorter than 2 chars: `a@gmail.com`
- Microsoft-domain rules (msn/hotmail/outlook/live), e.g. a local part not starting with a letter/number/underscore: `+user@outlook.com`
- TLD not in ICANN's list: `user@example.notarealtld`

Each would need custom logic — and notably `domain_specific_validation: true` is not the answer for the Gmail one: it enforces validator's 6–30 char rule, which is *stricter* than Braze's documented ≥2, so it could drop addresses Braze accepts. Fine to leave these as-is.

shekhar-rudder: Applied in 3667ff9 — and Braze's own published validation regex confirms it: `"` is absent from the local-part character set, so this isn't just a prose rule.

Added three quoted-address cases to the invalid list: `"user name"@example.com`, `"user"@example.com`, `"user@internal"@example.com`.

On why it can't over-reject: `blacklisted_chars` is tested against the raw local part *before* `isEmail`'s fully-quoted branch (`validator/lib/isEmail.js:156-161`), and `isEmail` admits `"` in a local part **only** via that branch — so every address this newly rejects opens with `"`, which can never match Braze's regex. I ran a differential against a JS port of that regex over 2,000,000 randomised local parts: 10,085 newly rejected, **0** of which Braze would accept.

Two things that came out of the regex rather than the prose:

- **Deliberately not widening the blacklist to `{`/`}`**, even though Braze's charset excludes them and `validator`'s allows them. The doc adds that "the validation test checks whether the local part preceding the `+` matches the regular expression" — so Braze *accepts* `user+{tag}@example.com`. `blacklisted_chars: '"{}'` measured ~2.5k false positives per 500k random local parts. I've noted this in the code comment so it doesn't get "improved" later.
- **`+user@outlook.com` isn't a Microsoft-domain rule** — a leading `+` leaves an empty pre-`+` segment, which fails the regex at *any* domain. `!email.startsWith('+')` closes it with 0 false positives over the same runs. Happy to add it, or leave it out to keep this PR narrow — your call.

One caveat for whoever picks up the remaining gaps: the doc's prose allowlist (`+ & # _ - ^ /`) contradicts its own regex, which also permits `! $ % ' * = ? \` | ~`. Hardening off the prose would over-reject badly.

ItsSudip: On the open question: let's add the `!email.startsWith('+')` guard. It's provably safe — Braze's regex requires at least one character before the `+`, so an empty pre-`+` segment can never match at any domain — and you've already measured 0 false positives over your runs. Please add a test case alongside it (e.g. `+user@example.com` in the invalid list).

shekhar-rudder: I checked this before implementing and my earlier claim was wrong — sorry, I led you to a bad conclusion. The pre-`+` rule is **not** general; it belongs to the Microsoft-domain section. Your original diagnosis ("Microsoft-domain rules, e.g. `+user@outlook.com`") was right.

Running Braze's two published regexes verbatim in Ruby (they're Ruby regexes — `\A`/`\z` and the `&&` class intersection don't port to JS cleanly):

| local part | general regex | microsoft regex (pre-`+` segment) |
|---|---|---|
| `+user` | **match** | no match |
| `+` | **match** | no match |
| `user` | match | match |

`+` sits in the general regex's *leading* character class:

```
\A([a-zA-Z0-9_\-\^+$'\&#\/!%\*=\?`\|~]|…)
         here ──────────^
```

So `+user@example.com` is a valid Braze address at every non-Microsoft domain, and an unconditional `!email.startsWith('+')` would drop good events — the exact over-rejection failure mode this PR is trying to avoid, except it discards the whole event rather than one field.

My "0 false positives over 2,000,000 local parts" number was worthless: the harness applied the pre-`+` truncation at every domain, which baked the error into both sides of the differential. That's the only way it could have come back 0.

Scoping the guard to Microsoft domains would be correct per the doc, but I don't think it's provably safe either — the doc says domains *containing* `msn`/`hotmail`/`outlook`/`live`, and a literal substring match also catches `olive.com` and `transmission.com`. Whether Braze really substring-matches is untested, so I'd rather not guess. Left the gap as you originally suggested.

Pushed 2f1ece7 instead: corrects the comment, and pins `+user@example.com` in the **valid** list so the guard doesn't get added back later.

---

The same wrong premise infected one more line of that comment, in the opposite direction. I'd written that Braze accepts `user+{tag}@example.com` — it doesn't. The general regex validates the *whole* local part, and `{` isn't in the character set:

| local part | general regex |
|---|---|
| `user+tag` | match |
| `user+{tag}` | **no match** |

So `blacklisted_chars: '\"{}'` would have been a *correct* tightening, not the over-rejection I claimed (my ~2.5k/500k measurement was invalid for the same reason as above). I've only dropped the wrong claim rather than widen the blacklist — that's a behaviour change and this thread was about the `+` guard. Happy to add `{}` here or in a follow-up, your call.

ItsSudip (approved):

PR #5506, opened 2026-08-21
fix(delivery): cap the destination response the proxy echoes back (INT-6978)

shekhar-rudder:
## Summary

Alternative to #5481, fixing the same P1 (`router_transformerproxy_invalid_response{reason="missing output"}` on `FB_CUSTOM_AUDIENCE`) at its cause rather than at the response boundary. #5481 catches the `RangeError` after the oversized response has been built and replaces it with a fallback; this PR stops the oversized response from being built.

**Root cause.** A proxy delivery response carries the destination's own body back to rudder-server verbatim — in `destinationResponse` for v0, in `response[].error` for v1. Nothing bounds that body, and on the v1 path it is repeated **once per job in the batch**, so what the transformer has to serialize grows as `batchSize x bodySize`. Past V8's ~512MB maximum string length `JSON.stringify` throws `RangeError: Invalid string length`, Koa answers with a body that has no `output` field, and rudder-server records that as missing output.

The echo exists for debugging. Beyond the first few KB it stops serving that purpose, so it is capped rather than removed.

## The change

Additive — **no existing production file is modified except one line in the delivery controller.** The diff is a new module, its tests, that one call, and a metric registration.

`capDeliveryV1Errors` (`src/services/destination/deliveryResponseCap.ts`) caps the `error` a v1 delivery response carries, at **50KB**, configurable via `PROXY_DESTINATION_RESPONSE_MAX_BYTES`. It is the module's only export; the truncation itself is private, so a job state cannot be capped without going through the memo below.

**One call site**, in `DeliveryController.deliverToDestinationV1`, immediately before `ctx.body`. That is the single point every v1 response reaches, whichever producer built it:

| producer | reaches the cap via |
| --- | --- |
| the v0→v1 adaptation | `deliver`'s return |
| a native v1 handler's own per-job states (`braze`'s `buildJobStates`, `hs`'s per-item states) | `deliver`'s return |
| the batching framework | `deliver`'s early return |
| a handler that throws carrying the body (facebook) | `deliver`'s catch → `handlevV1DeliveriesFailureEvents` |
| an error that escapes `deliver` entirely | the **controller's own catch** → `handlevV1DeliveriesFailureEvents` |

The last row is why the cap cannot live inside `deliver`: `handlevV1DeliveriesFailureEvents` has two callers and only one of them is `deliver`. Placing it at `ctx.body` also puts it at the last point before serialization, so no producer added later can forget it.

### The memo

Capping entry by entry would itself amplify. Producers hand back `response[]` as N references to one error string (or N equal strings a per-job map built), so truncating each independently allocates N distinct copies of the capped string and rescans the full body N times — at 50KB a job, re-creating in miniature the very `batchSize x bodySize` growth this exists to remove.

So `capDeliveryV1Errors` carries a one-entry memo. One entry is enough because the shared case is N *consecutive* hits, and `!==` on strings compares by value, so it also collapses equal-but-distinct strings. Measured at 2000 jobs x 2MB: **142MB and 657ms without it, 1.5MB and 0.8ms with it**.

### Why 50KB and not 100KB

The cap is **per job**, so the response still serializes to `batchSize x cap` — the cap lowers the ceiling, it does not remove the multiplication. That ceiling has to stay under V8's ~512MB limit at the largest batch the proxy is asked to deliver, and the batch that produced INT-6978 was 6000 jobs:

```
6000 x 100KB = ~600MB  ->  still RangeError: Invalid string length
6000 x  50KB = ~300MB  ->  serializes
```

100KB would have left the reported failure reproducible at the batch size that reported it. `deliveryResponseCap.test.ts` asserts `DEFAULT_MAX_BYTES * 6000 < 512MB` so the two numbers cannot drift apart silently.

### Why the cap sits at the response boundary and not earlier

The literal earliest point is right after `processAxiosResponse`, before the body reaches the network handler. That would be wrong: the parsed body is the handler's **classification input**, not only an echo.

- `facebookUtils/networkHandler.js` — `errorResponseHandler` reads `response.error.code` and `error_subcode` to choose auth vs retryable vs abort and to raise `ConfigurationAuthError`. Against a truncated string `response.error` is `undefined`, so every large FB error body would be misclassified and **OAuth refresh would never fire**.
- `v1/destinations/emarsys/networkHandler.js:15` — `const { errors } = destinationResponse.response.data`, a `TypeError` on a string.
- `nativeBatching/delivery.ts` — `ctx.response` is both what the per-item verdicts are computed from and what the whole-batch `TransformerProxyError` carries (`{ status: ctx.status, response: ctx.response }`); truncating it loses per-item attribution.

All of that would break on exactly the oversized responses the guard exists for. So the cap goes downstream of every handler, where the body is purely an echo.

## Properties worth checking in review

- **No change to normal responses.** A value within the cap is returned *by reference, unchanged* — same shape, same type, same bytes. Only a payload already too big to be useful is ever altered.
- **Byte-based, character-safe truncation.** `slice` counts UTF-16 code units, under-counts multi-byte characters and can leave a lone surrogate. The cap measures bytes with `Buffer.byteLength` and walks back off UTF-8 continuation bytes (`0x80..0xBF`). Tested by sweeping every cut offset across a 3-byte character and an emoji.
- **Per-job routing survives.** `statusCode` and `metadata` are untouched — only the `error` text is trimmed. This is the main behavioural difference from #5481, which replaces all per-job outcomes with a uniform retryable 500.

## Behaviour changes

An oversized echo becomes truncated text with a `...[truncated: N bytes exceeded the M byte limit]` suffix, so it is no longer parseable JSON. Nothing in this repo parses either field back (checked), and rudder-server treats both as opaque text.

## Scope: v1 only, deliberately

**v0 delivery responses are not capped, and should not be.** The target is the multiplication, not the echo:

- `ProxyV0Request.metadata` is a single `ProxyMetdata`; `ProxyV1Request` overrides it to `ProxyMetdata[]` (`src/types/destinationTransformation.ts`). One v0 request is one job, so a v0 response carries the body **once** — the response is the size of the body, with no `batchSize x bodySize` growth to bound.
- A single body is already bounded upstream: axios rejects anything over `MAX_CONTENT_LENGTH` (`src/adapters/network.js`, 100MB by default), roughly five times under the string length that breaks serialization.

So capping v0 would trade real debugging detail for no safety. It falls out of the structure: `deliverToDestinationV0` simply never calls `capDeliveryV1Errors`. A test drives the v0 route with a 2MB body and asserts the echo survives byte for byte, and it fails if v0 capping is introduced. The premise to re-check if that ever changes: a v0 route accepting an array `metadata`, or `MAX_CONTENT_LENGTH` raised past ~512MB.

### Residual risk

- **`message` and `statTags` are not capped**, on any path. Both are per-response rather than per-job, so neither amplifies.
- **The bound is `batchSize x 50KB`, not absolute.** A batch materially larger than 6000 jobs whose body genuinely exceeds 50KB would still be at risk. A hard guarantee independent of batch size would need an aggregate budget divided across `response[]` rather than a per-job limit; lowering `PROXY_DESTINATION_RESPONSE_MAX_BYTES` is the lever meanwhile.

## Observability

New counter `proxy_destination_response_truncated`, tagged `destType`, registered in `createMetrics` so it is present in `/metrics` from startup. It advances by the number of job states truncated — so the number answers "how many events lost detail", not "how many times the memo missed" — via one `stats.counter` call per response rather than one `increment` per job. Counter only, no log line: a destination chronically above the cap would truncate on every delivery, which at proxy throughput is spam rather than signal.

## Deliberately out of scope

**The `JSON.stringify` de-duplication.** Three sites serialize the *same* destination body once per job inside a `.map` — `nativeIntegration.deliver`'s v0→v1 adaptation, `postTransformation.handlevV1DeliveriesFailureEvents`, and `braze`'s `buildJobStates` — allocating `batchSize` independent copies before the response ever reaches the controller. Hoisting the call out of each map is measured at **602MB → 2MB of heap and 272ms → 1ms** on a 2MB body across 300 jobs, for byte-identical output.

That is a real win but an independent one: it is upstream of the cap, the memo cannot recover it (the copies exist by the time the controller runs), and it touches three service files. Keeping it out leaves this PR additive. Worth a follow-up.

## Known review findings

Raised in review, not addressed in this PR:

1. **`capDeliveryResponse` can return more than `maxBytes`.** When the budget cannot fit the truncation notice, `end` clamps to 0 and the notice alone is returned — 52 bytes for `maxBytes=5`. Only reachable via a misconfigured `PROXY_DESTINATION_RESPONSE_MAX_BYTES`; the doc comment overstates the contract.

## Test plan

- [x] `src/services/destination/__tests__/deliveryResponseCap.test.ts` — 23 tests, all through the public `capDeliveryV1Errors`: byte-vs-code-unit counting, character-boundary sweeps for 3-byte and 4-byte characters, pass-through-by-reference, idempotence, env-var parsing including non-positive values, the `50KB x 6000 < 512MB` bound, the memo (one allocation for 300 shared entries, and for 100 equal-but-distinct ones), per-job-state counting, and mixed over/under-cap batches
- [x] `src/controllers/__tests__/delivery.test.ts` — the cap driven through the **real HTTP route**, once per producer: the v0→v1 adaptation, a native v1 handler's own per-job states, the batching framework, and the **throw** path, each with a 2MB body; plus an in-cap response left byte-identical with no metric, and the **v0 route left uncapped at 2MB**
- [x] All four producer-path cap tests verified to **fail** when the single `capDeliveryV1Errors` call is removed, and the v0 guard verified to **fail** against a build that caps v0 — they are real regression guards, not tautologies
- [x] `npx tsc --noEmit` clean, `eslint` clean, prettier clean
- [x] Full unit suite: 0 new failures (6 suites fail on `develop` too — sandbox/ivm, missing `dist/mappingsSandbox.bundle.js`)

Not yet live-tested against mini — happy to do that before this leaves draft if we decide to take this over #5481.

### DIFF
--- a/src/services/destination/deliveryResponseCap.ts
+++ b/src/services/destination/deliveryResponseCap.ts
@@ -1,0 +1,143 @@
+import { DeliveryJobState } from '../../types/index';
+import stats from '../../util/stats';
+import { parseEnvInt } from '../../util/utils';
+
+/**
+ * Bounds the `error` a v1 delivery response carries back to rudder-server.
+ *
+ * That field is where the proxy reports why a job failed, and it is usually the destination's own
+ * response body verbatim. A v1 response repeats it once per job in the batch (`response[].error`),
+ * so what the transformer has to serialize grows as `batchSize x bodySize`. At a few MB per body
+ * and a few thousand jobs that crosses V8's ~512MB maximum string length, `JSON.stringify` throws
+ * `RangeError: Invalid string length`, and rudder-server receives a body with no `output` field —
+ * which it reports as `router_transformerproxy_invalid_response{reason="missing output"}`.
+ *
+ * The detail exists for debugging, and beyond the first few KB it stops serving that purpose, so it
+ * is capped rather than removed.
+ *
+ * # v1 only, deliberately
+ *
+ * v0 delivery responses carry the same echo in `destinationResponse` and are **not** capped. That
+ * is a decision, not an oversight, and the reason is the multiplication rather than the echo:
+ *
+ * - `ProxyV0Request.metadata` is a single `ProxyMetdata`; `ProxyV1Request` overrides it to
+ *   `ProxyMetdata[]`. One v0 request is one job, so a v0 response carries the body **once**. There
+ *   is no `batchSize x bodySize` growth to bound — the response is the size of the body.
+ * - A single body is already bounded upstream: axios rejects a response larger than
+ *   `MAX_CONTENT_LENGTH` (`src/adapters/network.js`, 100MB by default), which is roughly five times
+ *   under the string length that breaks serialization.
+ *
+ * So capping v0 would trade real debugging detail for no safety. If that ever changes — a v0 route
+ * that accepts an array `metadata`, or `MAX_CONTENT_LENGTH` raised past ~512MB — the premise above
+ * is what to re-check.
+ */
+
+/**
+ * 50KB, chosen against the batch size rather than picked for readability.
+ *
+ * The cap is per job, so what the response serializes to is still `batchSize x MAX_BYTES` — the cap
+ * lowers the ceiling, it does not remove the multiplication. That ceiling has to stay under V8's
+ * ~512MB maximum string length at the largest batch the proxy is asked to deliver, and the batch
+ * that produced INT-6978 was 6000 jobs:
+ *
+ *   6000 x 100KB = ~600MB  -> still `RangeError: Invalid string length`, the original bug
+ *   6000 x  50KB = ~300MB  -> serializes
+ *
+ * So 100KB would have left the reported failure reproducible at the batch size that reported it.
+ * Raising `PROXY_DESTINATION_RESPONSE_MAX_BYTES` above 50KB is safe only where batches stay well
+ * under 6000; a guarantee independent of batch size would need an aggregate budget divided across
+ * `response[]` rather than a per-job limit.
+ */
+const MAX_BYTES = parseEnvInt(process.env.PROXY_DESTINATION_RESPONSE_MAX_BYTES, 50 * 1024);
+
+/**
+ * Cap one `error` string to at most `maxBytes` UTF-8 **bytes**.
+ *
+ * Module-private: `capDeliveryV1Errors` is the only entry point, so there is no way to cap a job
+ * state without going through the memo that keeps the capping itself from amplifying.
+ *
+ * Returns the error unchanged when it already fits, so the caller can apply it unconditionally and
+ * a normal response keeps its exact bytes.
+ *
+ * Bytes rather than `String.prototype.slice`, which counts UTF-16 code units: that under-counts
+ * multi-byte characters, so a byte budget expressed in code units is wrong by up to 4x, and a cut
+ * between the halves of a surrogate pair leaves a lone surrogate behind.
+ */
+const capDeliveryResponse = (error: string, maxBytes: number): string => {
+  // `JSON.stringify` returns `undefined` at runtime for an absent body, despite its TypeScript
+  // return type, and a job state built from one carries that through. Nothing to cap.
+  if (typeof error !== 'string') return error;
+
+  // Allocation-free, and the overwhelming majority of errors never exceed the budget.
+  const bytes = Buffer.byteLength(error);
+  if (bytes <= maxBytes) return error;
+
+  const notice = `...[truncated: ${bytes} bytes exceeded the ${maxBytes} byte limit]`;
+  // `slice` first so the copy is bounded: this runs on exactly the oversized values the cap exists
+  // to keep out of memory, and `Buffer.from(error)` would copy all of one. The bound is safe
+  // because every UTF-16 code unit encodes to at least one UTF-8 byte, so the first `maxBytes` code
+  // units always contain the first `maxBytes` bytes.
+  const buf = Buffer.from(error.slice(0, maxBytes));
+  let end = Math.max(0, maxBytes - Buffer.byteLength(notice));
+  // Back off any UTF-8 continuation byte (0x80..0xBF) so the cut lands on a character boundary.
+  while (end > 0 && buf[end] >= 0x80 && buf[end] <= 0xbf) {
+    end -= 1;
+  }
+
+  return buf.subarray(0, end).toString() + notice;
+};
+
+/**
+ * Cap every `error` in a v1 delivery response, in place. **The only place the cap is applied.**
+ *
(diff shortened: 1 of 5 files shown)


src/services/destination/deliveryResponseCap.ts
```diff
@@ -0,0 +1,143 @@
...
+ *
+ *   6000 x 100KB = ~600MB  -> still `RangeError: Invalid string length`, the original bug
+ *   6000 x  50KB = ~300MB  -> serializes
+ *
+ * So 100KB would have left the reported failure reproducible at the batch size that reported it.
+ * Raising `PROXY_DESTINATION_RESPONSE_MAX_BYTES` above 50KB is safe only where batches stay well
+ * under 6000; a guarantee independent of batch size would need an aggregate budget divided across
+ * `response[]` rather than a per-job limit.
+ */
+const MAX_BYTES = parseEnvInt(process.env.PROXY_DESTINATION_RESPONSE_MAX_BYTES, 50 * 1024);
```
ItsSudip: Two data points from rudder-server that I think argue for a lower default than 50 KB:

1. **Anything above 10 KB per job is discarded downstream.** `router/worker.go` trims every per-job body to 10 KB before it reaches jobsDB (`trimmedResponse := string(lo.Slice([]byte(routerJobResponse.respBody), 0, int(10*bytesize.KB)))`), and the v1 adapter reads only `metadata.jobId`, `metadata.dontBatch`, `statusCode` and `error` per entry. So a 10 KB default is lossless for the consumer — the extra 40 KB per job never survives.
2. **The `6000 × 50 KB` guard doesn't cover the largest v1 batch we ship.** `google_adwords_remarketing_lists`' Data Manager path batches `DATA_MANAGER_BATCH_SIZE = 10000` jobs per proxy request. `10,000 × 50 KB = 512,000,000` chars against V8's `536,870,888` limit leaves ~24 MB for 10,000 echoed `metadata` objects (~2.4 KB each, and `secret` alone can eat most of that). GARL hit this exact `RangeError` on 2026-08-08 (`missing output` on that destType), so it's a live case, not hypothetical. At 10 KB the same batch is ~100 MB + metadata.

Related: the server's trim is a naive byte slice, so the UTF-8 boundary care here is undone for anything over 10 KB anyway — one more reason the two limits should agree.

Suggest `10 * 1024` as the default (env override stays), and updating the guard test to the 10,000-job batch.

shekhar-rudder: As discussed on the call, for now lets keep it 50KB

ItsSudip (dismissed):

ItsSudip (approved):

PR #5519, opened 2026-08-25
fix(tiktok_audience): keep batch_data aligned with id_schema

arpl:

TikTok reads each `batch_data` row **positionally** against `id_schema`. `prepareIdentifiersPayload` built the row by iterating the `identifiers` object and pushing only truthy values, which misaligned it in two independent ways:

1. **A missing identifier produced a shorter row**, shifting every later id left. A user with only an email was sent as a one-entry row against a four-entry schema, so TikTok read the email hash as an `AAID_SHA256`. This is the customer report.
2. **`id_schema` is sorted, the row was not.** The row followed the `identifiers` object's own key order, so even a *fully populated* user could be emitted out of order — `AAID` and `PHONE` silently swapped. This was not in the report and is the more insidious of the two, since nothing about the record looks wrong.

The fix walks the sorted schema instead of the object, and emits `{}` for any slot the user has no usable value for.

```ts
const idSchema = Object.keys(identifiers).sort();
for (const fieldName of idSchema) {
  let identifier: BatchIdentifier = {};   // holds the slot open
  if (value) { /* validate, normalize, hash */ }
  identifiersList.push(identifier);
}
```

Events whose slots are *all* placeholders still abort with `No identifiers found` — a row of nothing but placeholders identifies no one.

### Why `{}` and not a reordered `id_schema`

Two independent confirmations that the placeholder is the contract:

- **This repo already does it correctly elsewhere.** The legacy `audienceList` path in `transform.ts` maps over the schema and emits `{}` for a missing trait:
  ```js
  destinationFields.map((f) => trait[f] ? { id: hashIdentifier(f, trait[f]), audience_ids: [audienceId] } : {})
  ```
  The two paths disagreed; the older one was right. This change makes the `record` path match it.
- **Segment's TikTok Audiences destination** does the same, with a comment stating the rule: *"Even if email and advertising_id are empty we want to include them into the array as we seek to comply with the ID Schema defined outside of this function."*

Not yet filed — reported by Chatbooks in a shared Slack channel and triaged in #<internal thread>. Happy to link one before merge.

Restore correct identifier attribution for TikTok audiences. Any record with a missing or invalid identifier currently has its remaining ids attributed to the wrong identifier types, which corrupts audience matching silently — TikTok returns `200` either way. Customers only notice as unexplained low match rates.

The affected customer worked around it by rebuilding their connection to match on email only, which is exactly the configuration in which the bug cannot appear (a single-slot schema is always aligned). Enriching with any additional identifier re-exposes it.

`batch_data` rows now always have exactly `id_schema.length` entries, where previously they had only as many entries as the record had non-null identifiers. This is the fix, and it is a wire-format change for every `record`-path TikTok audience payload with a sparse or unsorted identifier set.

`test/integrations/destinations/tiktok_audience/router/data.ts` — **"Native Test 18: record events: invalid identifiers"** had frozen the old behaviour into its expected payload (two ids against a four-entry `id_schema`, i.e. the misalignment itself) and had always passed. Its fixture is updated; please sanity-check that fixture rather than assuming it was collateral damage.

N/A

N/A — one new local type, `BatchIdentifier = Identifier | IdentifierPlaceholder`, so a placeholder is representable.

Payload size grows for sparse records (an extra `{}` per unfilled slot). Grouping is unchanged: it already keys on `id_schema`, and rows are now uniformly that length.

<hr>

## Verification

**Unit** — 12/12 in `recordTransform.test.ts`, including 3 new alignment cases. All three were confirmed to **fail before the fix** (row length 1 vs schema 4; AAID/PHONE hashes swapped), so they can actually catch a regression.

**Component** — 16/16 for `tiktok_audience`. Repo-wide `tsc --noEmit`, eslint and prettier clean.

**End-to-end on a local stack, against the real TikTok API.** Transformer built from this branch and deployed into the stack, verified the patched code was actually in the running container, then drove a real rETL sync from a warehouse source into a live TikTok audience with `AAID_SHA256`, `EMAIL_SHA256`, `PHONE_SHA256` mapped and nulls seeded across every gap shape. The payload was captured off the wire between rudder-server and the transformer:

```
 20 ok                                    (every row length == id_schema length)
 11 ["AAID_SHA256","EMAIL_SHA256","PHONE_SHA256"]

  5  -           | EMAIL_SHA256 | -              both null
  5  -           | EMAIL_SHA256 | PHONE_SHA256   <- leading gap: the reported bug
  5  AAID_SHA256 | EMAIL_SHA256 | -              trailing gap
  5  AAID_SHA256 | EMAIL_SHA256 | PHONE_SHA256   fully populated
```

Sync result: **20 changed → 20 successful, 0 failed / dropped / invalid.** TikTok accepted the `{}` placeholders.

Running the identical payload against the current published transformer image reproduces both bugs:

```
                     BEFORE                          AFTER
row 0  AAID_SHA256   jared@example.com   <- bug      (no entry)
       EMAIL_SHA256  (no entry)                      jared@example.com
row 1  AAID_SHA256   marketing@example   <- bug      (no entry)
       EMAIL_SHA256  +15551234567        <- bug      marketing@example.com
       PHONE_SHA256  (no entry)                      +15551234567
row 2  AAID_SHA256   +15559998888        <- bug      aaid-value-here
       PHONE_SHA256  aaid-value-here     <- bug      +15559998888
```

Row 2 is bug #2: all identifiers present, `AAID` and `PHONE` swapped.

## Linear ticket 
[Resolve INT-7054](https://linear.app/rudderstack/issue/INT-7054/fix-tiktok-audience-identifier-schema-alignment)

### Developer checklist

- [ ] **No breaking changes are being introduced.** — see *changes to existing behaviour* above: the emitted `batch_data` shape changes. That is the fix, but it is a wire-format change worth a reviewer's eyes.

- [x] Is the PR limited to 10 file changes? — 4 files

### DIFF
--- a/src/v0/destinations/tiktok_audience/recordTransform.test.ts
+++ b/src/v0/destinations/tiktok_audience/recordTransform.test.ts
@@ -234,3 +234,100 @@ describe('processTiktokAudienceRecords tiktok_audience record edge cases', () =>
     expect(id).toBe(md5('abcdef'));
   });
 });
+
+describe('processTiktokAudienceRecords batch_data / id_schema positional alignment', () => {
+  type BatchedJSON = {
+    batchedRequest: {
+      body: {
+        JSON: {
+          batch_data: ({ id: string; audience_ids: string[] } | Record<string, never>)[][];
+          id_schema: string[];
+        };
+      };
+    };
+  };
+
+  const readPayload = (response: RouterTransformationResponse) =>
+    (response as unknown as BatchedJSON).batchedRequest.body.JSON;
+
+  it('null identifiers are sent as {} placeholders so each slot lines up with id_schema', () => {
+    const event = buildBaseEvent({
+      identifiers: {
+        AAID_SHA256: null,
+        EMAIL_SHA256: 'user@example.com',
+        IDFA_SHA256: null,
+        PHONE_SHA256: null,
+      },
+    });
+
+    const { failedResponses, successfulResponses } = processTiktokAudienceRecords([event]);
+    expect(failedResponses).toHaveLength(0);
+    expect(successfulResponses).toHaveLength(1);
+
+    const { batch_data: batchData, id_schema: idSchema } = readPayload(successfulResponses[0]);
+    expect(idSchema).toEqual(['AAID_SHA256', 'EMAIL_SHA256', 'IDFA_SHA256', 'PHONE_SHA256']);
+    expect(batchData[0]).toHaveLength(idSchema.length);
+    expect(batchData[0]).toEqual([
+      {},
+      { id: sha256('user@example.com'), audience_ids: ['23856594064540489'] },
+      {},
+      {},
+    ]);
+  });
+
+  it('identifiers supplied out of alphabetical order are re-ordered to match id_schema', () => {
+    const event = buildBaseEvent({
+      identifiers: {
+        PHONE_SHA256: '+15551234567',
+        EMAIL_SHA256: 'user@example.com',
+        AAID_SHA256: 'AAID-VALUE',
+      },
+    });
+
+    const { failedResponses, successfulResponses } = processTiktokAudienceRecords([event]);
+    expect(failedResponses).toHaveLength(0);
+    expect(successfulResponses).toHaveLength(1);
+
+    const { batch_data: batchData, id_schema: idSchema } = readPayload(successfulResponses[0]);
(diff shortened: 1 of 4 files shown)


maheshkutty (approved):

maheshkutty:
going via hotfix 
https://github.com/rudderlabs/rudder-transformer/pull/5520

PR #5532, opened 2026-08-27
feat(reddit_audience): add Reddit Custom Audiences destination

arpl:
Syncs warehouse audience membership into Reddit Ads Custom Audiences via `PATCH /custom_audiences/{audience_id}/users`, on the native batching framework.

Spec: rudderlabs/rudder-specs#148 · Config: rudderlabs/rudder-integrations-config#2694

## The shape that drives the design

Reddit's body is a **positional matrix**, not named objects — `column_order` is declared once per request and every `user_data` row must align to it positionally.

Two consequences:

- `ADD` and `REMOVE` **share one URL**, unlike `iterable_audience` which gets free separation from distinct subscribe/unsubscribe endpoints.
- Rows carrying different identifier sets cannot travel together, because they need different `column_order`.

Both are handled with `internalGroupKey = ${action_type}|${columnOrder.join(',')}` — precisely the discriminator the framework documents for "action type when multiple actions share a URL". `wrapBody` reads the group-invariant fields off `bodies[0]`, which is safe because grouping guarantees every body in a group carries the same pair.

## Hashing is pinned to Reddit's published vectors

Reddit's email rule is unusual — lowercase, strip the `+alias`, then strip **all** non-alphanumerics from the username, on *every* domain (not Gmail-only dot-stripping). Getting it wrong produces a silently **zero match rate** with no error from Reddit, so all four of Reddit's own published test vectors are asserted end-to-end through `processAudienceRecord`.

There is also a guard that **MAIDs are not case-folded**: IDFA is uppercase hex and AAID is lowercase, so a well-meaning "normalize everything" edit would silently break every iOS device.

## Misaligned-matrix guard

**Verified live against the Reddit API: a row whose length disagrees with `column_order` is answered `204`, not `400`** — both a 2-value row under a 1-column order and a 1-value row under a 2-column order. A misalignment therefore ships silently, shifts every value into the wrong column, matches nobody, and leaves nothing for the delivery spec's 400 path to classify.

Grouping already guarantees alignment, so a mismatch is a bug in this file rather than bad customer data — `wrapBody` asserts it and fails loudly. **Mutation-checked**: removing the guard fails exactly the three tests covering it.

## Delivery

`204` is the documented success and there is no per-record envelope, so no `perItem` correlation exists. The framework default is nearly right; `delivery.ts` refines only what OAuth and Reddit's taxonomy need:

- `401` → `authExpired` (token refresh), **derived from the response body**, not the status — matching the existing `reddit` handler's `error.reason === 'UNAUTHORIZED'` check, per the batching-framework guidance never to infer auth from a status code. An unrecognized 401 falls through rather than burning a refresh.
- `403` → abort naming the `adsedit` scope / blocked User-Agent. `404` → abort naming a deleted audience.
- `failureReason` joins `error.fields[]` onto `error.message`. Reddit returns two different 400 shapes — schema violations carry `fields[]`, semantic ones only `message` — and the extractor handles both.

No `src/v1/destinations/reddit_audience/` tree: new destinations write only `delivery.ts`.

A `User-Agent` is sent on every request — Reddit maps a missing or generic agent onto both 403 and 429.

## Verification

- **55 unit + 12 component tests**, `npm test` 5,718 passing / 108 suites, `test:ts component` 2,404 passing, lint clean.
- **Live**: the emitted payload was accepted by a real Reddit ad account — `204` for ADD, REMOVE, and a two-column row — and the audience's status advanced `PENDING → NOT_ENOUGH_MATCHES_ERROR`, proving Reddit *ingested and matched over* the rows rather than no-oping.
- **Parity**: `/routerTransform` output diffed **byte-for-byte** against those live-accepted bodies (harness in the specs PR). This caught a real defect — `adAccountId` was required on the destination schema, which would have rejected every record from a wizard-configured destination; every component fixture happened to set it, so only an end-to-end call surfaced it.


arpl:
Updated after a cross-repo review panel — 13 findings survived adversarial verification and are fixed here and in the sibling PRs. Every fix is mutation-checked (reverted, test confirmed red, restored). Summary and the full finding list: rudderlabs/rudder-specs#148

am6010 (approved):

ItsSudip (approved):

PR #5542, opened 2026-09-02
docs(google_adwords_enhanced_conversions): spec for INT-6579 conversion action batching

shekhar-rudder:
Spec only — no code changes. Design document for [INT-6579](https://linear.app/rudderstack/issue/INT-6579).

📄 `docs/superpowers/specs/2026-09-02-int-6579-gaec-conversion-action-batching-design.md`

## Problem

`transform.ts:61` puts the conversion action **name** into the delivery request's `params`, and the native batching framework builds its composite grouping key from the entire `params` object (`processBatchedDestination.ts:113`). Every distinct conversion name therefore forces its own batch — a destination with K configured conversions makes ≥K requests where Google's `uploadConversionAdjustments` accepts one of up to 2000 adjustments.

The name can't just be deleted: the proxy uses it at delivery time to resolve the Google resource name (`networkHandler.ts:70`). The fix is a relocation, not a removal.

## Design summary

- **Transform** (feature-flagged on `DEST_GAEC_CONVERSION_ACTION_IN_BODY_ENABLED_WORKSPACE_IDS`): carry the name as `conversionActionName` on each adjustment; drop `event` from `params`. Remaining params are all destination-level, so the key collapses to one group per destination.
- **Proxy**: resolve all distinct names in one GAQL `WHERE name IN (?)` query, mirroring GAOC's existing `batchFetchConversionActions` (`google_adwords_offline_conversions/utils.js:611`), with the current 24h cache plus a short-TTL negative entry.
- **Per-item failure containment**: adjustments whose name doesn't resolve are dropped from the outgoing payload, then Google's `results` array is re-expanded back to the original indices with `{}` at the dropped positions. That keeps `delivery.ts` and `v1/networkHandler.ts` working **unmodified**, and means one stale conversion name aborts only its own jobs instead of every co-batched event.
- **Rollout**: merge flag-unset so the tolerant proxy ships first, then ramp. Rollback is a flag unset — in-flight new-shape jobs still deliver via the `?? params.event` fallback.

## Open for review

The three decisions worth pushing back on are the per-item drop/re-expand (vs. simply aborting the whole batch), the feature flag (vs. two sequenced PRs), and porting GAOC's batch fetch (vs. parallel per-name SDK calls). Alternatives and trade-offs for each are in the spec.

https://claude.ai/code/session_01RasAAA3UURwu44L9EemCZm

### DIFF
--- a/docs/superpowers/specs/2026-09-02-int-6579-gaec-conversion-action-batching-design.md
+++ b/docs/superpowers/specs/2026-09-02-int-6579-gaec-conversion-action-batching-design.md
@@ -1,0 +1,269 @@
+# INT-6579 — Move the conversion action out of GAEC proxy params
+
+Ticket: <https://linear.app/rudderstack/issue/INT-6579>
+Destination: `GOOGLE_ADWORDS_ENHANCED_CONVERSIONS` (GAEC)
+
+## Problem
+
+`transform.ts` puts the conversion action **name** into the delivery request's `params`:
+
+```ts
+// src/v0/destinations/google_adwords_enhanced_conversions/transform.ts:61
+deliveryRequest.params = { event, customerId, accessToken, loginCustomerId, subAccount };
+```
+
+The native batching framework groups payloads by a composite key built from the **entire** `params`
+object:
+
+```ts
+// src/services/destination/nativeBatching/processBatchedDestination.ts:113
+const key = stableStringify({
+  endpoint: payload.endpoint,
+  method: payload.method,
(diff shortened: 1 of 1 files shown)


shekhar-rudder:
Closing — spec/design docs shouldn't go through PR review unless explicitly requested. The spec remains on branch `feature/int-6579-update-google-ads-enhanced-conversions-network-handler` at `docs/superpowers/specs/2026-09-02-int-6579-gaec-conversion-action-batching-design.md`.

PR #5543, opened 2026-09-02
refactor(destination-integration): rename BatchDestination to DestinationIntegration

shekhar-rudder:
## What

Renames the batching framework's base class `BatchDestination` → **`DestinationIntegration`**, and the surrounding symbol vocabulary with it. No behaviour change.

## Why

Two reasons, and the second is the one that actually motivated it:

1. **It outgrew "batch".** Since #5418 moved delivery response handling onto the class, and #5541 made delivery gate on the same predicate as the router transform, the class now covers a destination's full router-stage contract: input schema, per-event transform, batch strategy, and delivery response policy.
2. **It is not a destination.** It *has* one — `constructor(destination: Destination, connection?: Connection)`, where `Destination` is a real, distinct type.

The codebase had already drifted this way on its own: every leaf class is `<Dest>Integration`, the loader reads `.Integration`, and the LLD's terminology table called the base class `RouterIntegration`.

## Renames

| Old | New |
| --- | --- |
| `BatchDestination` | `DestinationIntegration` |
| `BatchDestinationConstructor` | `DestinationIntegrationConstructor` |
| `processBatchedDestination` | `processDestinationIntegration` |
| `batchedDestinationsMap` | `destinationIntegrationsMap` |
| `isBatchingFrameworkEnabled` | `isDestinationIntegrationEnabled` |
| `getBatchingFrameworkGaDestinations` | `getGaDestinationIntegrations` |
| `getBatchDestinationHandler` | `getDestinationIntegrationHandler` |
| `batchDestinationHandlerMap` | `destinationIntegrationHandlerMap` |
| `useBatchingFramework` | `useDestinationIntegration` |
| `src/services/destination/nativeBatching/` | `src/services/destination/destinationIntegration/` |

The directory is `destinationIntegration/` rather than the shorter `integration/` on purpose: it sits beside `nativeIntegration.ts` and `cdkV2Integration.ts`, which are `DestinationService` implementations, and a bare `integration/` there reads like a third service.

## Deliberately NOT renamed

- **The `{DEST}_BATCHING_FRAMEWORK_ENABLED_WORKSPACE_IDS` env vars** — live in rudder-devops `customer-objects`. Renaming them would silently revert enrolled workspaces to the legacy path, which is a far worse outcome than an inconsistent name.
- **`VDMV2ObjectDestination`** and its `vdmV2ObjectDestination.ts` filename — kept by request. The file moves into the renamed `destinationIntegration/` directory and the class still extends `DestinationIntegration`; only its own name is unchanged.
- **The `batching: true` capability key** in `features.ts` — it describes a capability of the destination, and it is data rather than a symbol. (It is not in `FeaturesConfig`, so it is not on the wire to rudder-server either way.)
- **`CHANGELOG.md`** and the dated design spec under `docs/superpowers/specs/` — frozen historical artifacts.
- **The `batching-framework*` skill and LLD directory names** — the *framework* is still called the batching framework; only the class vocabulary changed.

Happy to fold any of these in if reviewers would rather have full consistency.

## Docs corrected in passing

The rename rewrote lines in docs that had gone stale when #5541 removed the separate delivery flag, so leaving them would have restated false claims under new names:

- Three docs still described a separate `{DEST}_BATCHING_FRAMEWORK_DELIVERY_ENABLED_WORKSPACE_IDS` flag that no longer exists in `src/`. Now: delivery has no flag of its own and is gated on the same predicate.
- Three docs instructed hand-editing `destinationIntegrationsMap.ts` to enrol a destination. That file is derived from `features.ts` via `getGaDestinationIntegrations()` and cannot be hand-edited. Now: mark `batching: true` in `features.ts`.
- The LLD called the base class `RouterIntegration` and pointed at `src/services/destination/routerIntegration.ts`, which never existed.

## Verification

- `tsc --noEmit` clean on both `tsconfig.json` and `tsconfig.test.json`
- 707 unit tests / 51 suites pass (framework, constants, helpers, controllers, and all migrated destinations)
- 38 api tests pass
- 243 component tests pass across the 8 destinations with component data (posthog, custom_audience, test_destination, google_adwords_enhanced_conversions, iterable_audience, reddit_audience, braze_audience, customerio)
- `prettier --check .` clean repo-wide (`format:ci`), 0 eslint errors
- Verified no string literal, metric name, statTag, log message, JSON key, env var, or test fixture *value* was altered — the diff is identifiers, import paths, and prose only

https://claude.ai/code/session_01Lg5DBihMSoGisjLAwD2DTo


src/services/destination/__tests__/deliverDestinationIntegration.test.ts
krishna2020, line 1: Move deliveryTest.ts also to this location

krishna2020: Also move `delivery.test.ts` to destinationintegration/__tests__/

shekhar-rudder: Moved in d7fe3dc — but to `src/services/destination/deliveryTest.ts` rather than into `__tests__/`.

`deliveryTest.ts` is production code, not a test: it holds `DeliveryTestService`, which `src/controllers/delivery.ts` imports for the transformer-proxy comparison path. Putting it under `__tests__/` would bury a live service in a test directory. So it now sits beside the other destination services, which also gets rid of the misspelled `src/services/delivertTest/` folder. Import in the controller updated; the file's own relative imports were unaffected since the directory depth is the same.

shekhar-rudder: Done in d7fe3dc — `delivery.test.ts` now lives at `src/services/destination/destinationIntegration/__tests__/delivery.test.ts`, alongside the other three suites there. Relative imports re-based for the extra directory level (`./delivery` → `../delivery`, `../../../util/stats` → `../../../../util/stats`, etc.).

krishna2020 (approved):

PR #5546, opened 2026-09-02
feat(features): declare transformerProxy capability for GA destinations

shekhar-rudder:
## What

Adds an optional `transformerProxy` flag to `DestinationCapabilities` and declares it for the **40 destinations that already have the transformer proxy enabled fleet-wide**, plus the `TEST_DESTINATION` fixture. `/features` gains a top-level map:

```json
{ "routerTransform": { ... },
  "supportTransformerProxyV1": true,
  "transformerProxy": { "CUSTOMERIO": true, ... } }
```

Pairs with rudderlabs/rudder-server#7337, which reads the map and only falls back to `Router.<DEST>.transformerProxy` when a destination is not declared.

## Behaviour change

**None.** All 40 declared destinations already have the env set to `true`, and the rudder-server side treats the capability as an alternative to the env rather than a replacement. The observable change comes later, when the env entries are removed and delivery keeps working.

## Deliberately excluded

Screened all 47 destination directories containing a `networkHandler` or `delivery.ts` for an actually-exported handler:

- **`GA` — a false positive.** `src/v0/destinations/ga/networkHandler.js` exports only `gaResponseHandler`, a deletion-API response parser. It exports no handler class, so `networkHandlerFactory` resolves GA to `genericNetworkHandler`. It is not proxy-capable despite the filename. The new invariant test exists to keep this class of mistake out.
- **`CLICKSEND`, `MONDAY`, `POSTSCRIPT`, `FB`** — real proxy handlers, but proxy has never been enabled for them in any environment. Declaring them would turn proxy delivery on for the first time, which is a rollout, not a config consolidation. Worth doing, separately, with its own validation.
- **`REDDIT_AUDIENCE`** — pre-GA. It has no destination definition in rudder-integrations-config, and `reddit_audience/networkHandler.ts` is marked as the legacy handler until GA.
- **`AF`, and `ACTIVE_CAMPAIGN` / `ATTENTIVE_TAG` / `KLAVIYO` / `KOCHAVA` / `MAILCHIMP` / `REFINER` / `SINGULAR` / `SLACK` / `ZENDESK`** — proxy is enabled for these (AF fleet-wide, the rest per-customer) but they have no handler of their own and are delivered by `genericNetworkHandler`, so they cannot honestly declare the capability. They stay env-driven, and their env entries cannot be cleaned up until each is decided.

## Test plan

- `src/features.test.ts`: **9 passed**, `tsc --noEmit` clean, prettier/eslint clean.
- Three new assertions, the first of which is the one that matters:
  - every destination declaring `transformerProxy` genuinely implements it — checked from the export shape, which is what fails for GA. Asserted from source rather than by requiring `networkHandlerFactory`, which would pull the whole `v0/util` runtime (and `isolated-vm`) into a previously light unit suite.
  - the never-enabled destinations and GA stay off the map
  - `AF` stays off the map
- `test/apitests/service.api.test.ts` and `src/services/__tests__/misc.test.ts` compare against `defaultFeaturesConfig` itself rather than a literal, so the new key needs no snapshot updates.
- Two adjacent suites (`src/services/__tests__/misc.test.ts`, `src/cdk/v2/utils.test.ts`) fail locally on an `isolated-vm` NODE_MODULE_VERSION mismatch; verified identical failures on the base commit with the change stashed, so they are a local ABI problem and not from this diff. CI is the arbiter.

## Follow-ups (not in this PR)

- Remove `Router.<DEST>.transformerProxy` for these 40 from rudderstack-operator and rudder-devops once a transformer image at or above this release is pinned everywhere. That is the change that actually deletes the drift.
- Decide `CLICKSEND` / `MONDAY` / `POSTSCRIPT` / `FB` individually.
- Decide what to do about the 10 proxy-enabled destinations with no handler of their own.
- Delete the two dead `Router.throttler.<DEST>.transformerProxy` keys.

https://claude.ai/code/session_01HbD4xkXp922cKx6zAVQnbg


src/features.ts
krishna2020, line 1: For destinations like AF that are using default network handler, can we do the same in DestinationIntegration framework too ? By having default delivery logic and needed integrations will override it.

krishna2020, line 1: Are there any cases where proxy has been not GA'ed yet but limited to only few workspaces/environments while reviewing the helm charts ?

shekhar-rudder: DestinationIntegration already reuses the default network handler proxy code. Are you suggesting to enable proxy for more destination types?

shekhar-rudder: I didn't find any such destinations

krishna2020: yes for these

> AF, and ACTIVE_CAMPAIGN / ATTENTIVE_TAG / KLAVIYO / KOCHAVA / MAILCHIMP / REFINER / SINGULAR / SLACK / ZENDESK — proxy is enabled for these (AF fleet-wide, the rest per-customer) but they have no handler of their own and are delivered by genericNetworkHandler

shekhar-rudder: I don't see the point in enabling proxy for destinations that don't have any custom logic. It only adds an additional network call between the server and transformer without providing any real benefit.

krishna2020 (approved):

PR #5550, opened 2026-09-03
fix: raise 4xx instead of 500 for untrusted set-value paths

shekhar-rudder:
## Root cause

`set-value` validates **every** segment of the path it is given and throws `Cannot set unsafe key: "<key>"` for `__proto__`, `constructor` and `prototype`. Amplitude passed raw customer trait names straight in:

```js
set(rawPayload, `user_properties.${trait}`, get(traits, trait));  // am/transform.js:312
```

A trait literally named `constructor` therefore crashed the router transformation. The throw is a plain `Error`, so `generateErrorObject` classified it as `errorCategory=transformation` / `statusCode=500` — a **retryable** status for a payload that can never succeed, so the job was retried until its TTL and re-notified Bugsnag on every attempt.

Not a code regression: `am/transform.js` was last touched 2025-10-16 and `set-value` has been pinned `^4.1.0` since the Amplitude refactor. The trigger was a payload shape that had not been seen before.

**Impact was bounded and has already stopped.** `transformer_event_transform_failure{destType="AM"}` for that workspace shows ~292 failures in a single burst on 2026-09-02 (~16:30–19:30 UTC) and zero in the 8h before this PR, with no data at now-1d/-2d/-3d/-7d.

## Fix

Amplitude was one instance of a general problem, so this fixes the class rather than the symptom.

`setValueForUntrustedPath` (`src/v0/util/index.js`) performs the write and, when `set-value` rejects the *path*, re-raises it as an **`InstrumentationError` (4xx)** instead of letting a plain `Error` become a 500. A key the customer chose is their input, not our defect: the event aborts once with an actionable message in the errors table, rather than being retried until its TTL.

**No list of reserved keys is maintained anywhere in this repo.** `set-value` enforces that policy but never exposes it — I checked 4.1.0 (the latest published version, 2023-07-12): its only options are `arrays`, `separator`, `split`, `merge`, `preservePaths`, none of which relaxes the check; `validateKey` is called unconditionally on every segment; and the entire exported surface is `split`, `cache`, `clear`, with no predicate. So on failure the wrapper asks `set-value` the same question again against a throwaway target. If it still refuses, the path is what it objected to → 4xx. Otherwise the failure came from the target object (frozen, exotic setter, …) and the original error is re-thrown untouched, so a genuine 5xx-worthy bug on our side still surfaces as one. That stays correct for free if the library's rules change, and the happy path pays nothing because the probe only runs after a write has already failed.

### Call sites updated

Every place an untrusted key reaches `set-value`:

| File | Key source |
|---|---|
| `am/transform.js` (traits, `group_type`) | customer traits / `integrations.Amplitude.groups` |
| `zendesk/transform.ts` ×2 | customer trait + organization field names |
| `customerio/util.ts` ×2 | customer traits, `user_properties` |
| `customerio/v2/util.ts` | customer traits |
| `revenue_cat/transform.js` | customer trait names |
| `fb/transform.js` | customer property keys |
| `hs/util.ts` | customer-supplied `externalId.identifierType` |
| `sources/braze/transform.js` | inbound webhook property names |
| `sources/message.js` (`setProperty`) | covers the shopify / slack / monday source transforms, which build `properties.${key}` from inbound payloads |

Call sites whose keys come from **our own** `mapping.json` / config deliberately stay on plain `set` — there a throw means a mapping bug we want to see as a 500, not customer data.

### Deliberately not changed

- **`extractCustomFields`** keeps its own inline reserved-word guard. The `set` inside `src/v0/util/index.js` is `setValue` from `@rudderstack/integrations-lib`, which treats the whole path as one literal key and **never throws** — a different function from the `set-value` package. Verified:
  ```
  integrations-lib setValue('user_properties.constructor') -> OK    {"user_properties.constructor":1}
  set-value        set('user_properties.constructor')      -> THROW Cannot set unsafe key: "constructor"
  ```
  Its guard is data hygiene, not crash protection, and changing it would alter behaviour for 32 callers.
- **`engage/utils.js`, `woopra/utils.js`** use lodash `set`, which silently no-ops on these keys instead of throwing. Not affected.
- `groups[groupInfo.group_type] = ...` in `getResponseData` uses plain assignment, so a `__proto__` group type reassigns that local object's prototype rather than creating an own key. Pre-existing, cannot reach `Object.prototype`, left alone.

## Known gap: `<mapping>[<customer key>]` prototype-chain lookups

There is a **second, independent** crash with the same trigger that this PR does **not** fix:

```js
if (eventPropsToPathMapping[k]) {          // k is a customer property name
  let rudderEventPath = eventPropsToPathMapping[k];
  if (rudderEventPath.includes('sub')) {   // TypeError: not a function
```

A plain lookup walks the prototype chain, so `eventPropsToPathMapping['constructor']` returns `Object` itself — truthy — and `.includes` then blows up, still as a 500. Measured for fb:

| property key | lookup resolves to | outcome |
|---|---|---|
| `constructor`, `__proto__`, `toString` | `Object.prototype` member | **TypeError → 500 (unfixed)** |
| `prototype` | `undefined` | reaches `setValueForUntrustedPath` → 4xx ✅ |

So in fb specifically, only `prototype` is covered by this PR; `constructor` — the exact key from the original alert — still 500s. The same pattern appears in ~25 further places (`eventNameMapping[event]` in tiktok_ads, klaviyo, drip, ometria, facebook_offline_conversions, sendgrid, adobe_analytics, …), where it can also cause a silent mis-mapping rather than a crash. Find them with:

```
grep -rnE "(Mapping|MAPPING|mapping)\[(k|key|trait|prop|field|event)\]" src/v0/destinations src/sources
```

Fix is `Object.prototype.hasOwnProperty.call(map, key)`. Left for a separate change.

## Testing

- `src/v0/util/index.test.js` — tests for `setValueForUntrustedPath`: accepted paths behave identically to raw `set-value`; rejected paths raise an `InstrumentationError` that `getErrorStatusCode` reports as **400**; the error names the offending path and keeps the library message; `Object.prototype` stays clean; target-side failures propagate unchanged and are *not* reclassified. It also asserts `set-value` still throws for those paths, so the suite fails loudly if a library upgrade changes the rules the wrapper depends on.
- `src/v0/destinations/am/transform.test.ts` — reproduces the production case end to end and asserts it is now an `InstrumentationError` naming `user_properties.constructor`, plus nested keys, group `group_type`, ordinary traits unaffected, and non-pollution of `Object.prototype`.
- `src/v0/destinations/fb/transform.test.ts` — asserts a `prototype` property yields a 4xx rather than a 500.
- Full JS unit suite, full TS/component suite, `tsc --noEmit`, eslint and prettier all clean.

https://claude.ai/code/session_015WLXFyWF9E2J76MH35bEHS


src/v0/util/index.js
```diff
@@ -1363,6 +1366,53 @@ const generateExclusionList = (mappingConfig) =>
...
+ */
+const setValueForUntrustedPath = (target, setPath, value, options) => {
+  try {
+    return setValuePkg(target, setPath, value, options);
+  } catch (error) {
+    try {
+      setValuePkg({}, setPath, value, options);
+    } catch {
+      // set-value objects to the path itself, not to `target`
+      throw new InstrumentationError(
```
maheshkutty, line 1408: Why to check against empty object?
I see the lib [send error](https://github.com/jonschlinkert/set-value/blob/master/index.js#L28C22-L28C44) `Cannot set unsafe key` which we catch it and return instrumentation error directly

shekhar-rudder: **Why check against an empty object?** To differentiate bad customer input from everything else. The catch sees *every* error `set-value` can throw, not just the unsafe-key one — e.g. a frozen sub-object or a throwing setter on the target gives `TypeError: Cannot add property address, object is not extensible` on a perfectly valid path. That is a bug on our side and should stay a 500 (retryable), while a customer-chosen key like `address.constructor` should become an `InstrumentationError` (4xx, abort once).

Replaying the same write against `{}` separates the two: `{}` is a known-good target, so if the write *still* throws, the library is objecting to the path itself → customer input → InstrumentationError. If it succeeds, the failure came from the target → re-throw the original error → 500.

**Why not match on the `Cannot set unsafe key` message?** Substring matching is fragile — the lib exports no error class or predicate, just plain `Error`s, so if a version bump rewords the message we would silently fall back to retryable 500s (the exact bug this PR fixes). And it is not the only path-rejection case: the same `validateKey` also throws `TypeError: Object keys must be strings or symbols`, and there might be more in future versions. The probe keeps the library itself as the single source of truth for what it rejects, and it only runs after a write has already failed, so the happy path pays nothing.

maheshkutty (approved):

PR #5551, opened 2026-09-04
fix(openai_ads): reject events older than OpenAI's ingest window

shekhar-rudder:
## What

Rejects OpenAI Ads events whose `timestamp_ms` falls outside OpenAI's documented ingest window — the last 7 days, and no more than 10 minutes in the future — instead of sending them and letting OpenAI reject the batch.

## Why

OpenAI enforces the window **per batch, atomically**. One violating event fails the entire request:

```
422 event_timestamp_too_old
{"error":{"code":"event_timestamp_too_old",
  "message":"event_timestamp_ms must be within the last 7 days."}}
```

We cannot attribute that failure to a single job. The delivery spec maps `400` and `422` alike to `retry({ dontBatch: true })` (`delivery.ts:57-63`) without reading `param`, so the batch gets split blindly, one event at a time. At `MAX_BATCH_SIZE = 1000`, a single bad event turns one request into 1001 — and the bad one is aborted at the end of that anyway.

Checking the window before the request turns that whole cycle into one `InstrumentationError` against the one event that is actually bad.

## Scope

Both halves of the window are enforced, off a single `Date.now()` read, at OpenAI's own values — `7 days` and `10 minutes` — with no safety margin on either.

The one trade-off worth naming: OpenAI re-evaluates the window when it *receives* the batch, which is later than when we build the event. For the age bound that only makes us more permissive than OpenAI (safe — an event that ages out in transit is OpenAI's call, not ours). For the future bound it runs the other way: an event between 10m00s and 10m00s + transit-delay in the future is rejected here though OpenAI might have accepted it on arrival. That band is narrow and anything in it is clock skew that is anomalous on its own; the realistic offenders are unit-confusion timestamps (µs/ns read as ms), which land thousands of years out. Adding slack to close the band would reintroduce the fudge constant this branch deliberately removed, so it is left strict.

### DIFF
--- a/src/v0/destinations/openai_ads/utils.ts
+++ b/src/v0/destinations/openai_ads/utils.ts
@@ -19,6 +19,9 @@ import {
   CUSTOMER_ACTION_DATA_TYPE,
   CUSTOM_EVENT_SENTINEL,
   DESTINATION,
+  EVENT_AGE_SAFETY_MARGIN_MS,
+  MAX_EVENT_AGE_MS,
+  MAX_EVENT_FUTURE_SKEW_MS,
   STANDARD_EVENT_DATA_TYPES,
 } from './config';
 import mappingConfig from './data/OPENAI_ADSConfig.json';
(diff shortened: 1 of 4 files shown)


src/v0/destinations/openai_ads/utils.ts
```diff
@@ -270,6 +271,21 @@ const resolveSourceUrl = (payload: EventBasePayload, actionSource?: string): str
...
 };
 
+const resolveTimestampMs = (payload: EventBasePayload): number => {
+  const timestampMs = payload.timestamp_ms;
+  // Narrowing only. `constructPayload` declares this mapping `required` and rejects a missing or
+  // unparseable timestamp before we get here, so this branch is not reachable in practice.
+  if (typeof timestampMs !== 'number' || !Number.isFinite(timestampMs)) {
+    throw new InstrumentationError('OpenAI Ads timestamp is required and must be a valid date');
+  }
+  if (timestampMs < Date.now() - MAX_EVENT_AGE_MS) {
```
maheshkutty: Should we also add a check for the future 10 minutes?

shekhar-rudder: Yes — added in f3b0125bd.

maheshkutty (dismissed):

maheshkutty (approved):

PR #5553, opened 2026-09-04
fix(openai_ads): resolve standard event names without a mapping row

shekhar-rudder:
## What

An event whose name is already a standard OpenAI event name resolves on its own name instead of aborting, whether or not the destination has `eventMapping` rows configured.

## Why

`resolveEventMapping` looked the source event name up in `config.eventMapping` and threw if it missed. `eventMapping` is optional in the destination's `schema.json` (only `rudderAccountId` is required) and the UI saves an empty table by default — so a destination in that state drops **100%** of its traffic, one `OpenAI Ads event mapping not found for <event>` per event, with nothing at save time to warn you.

But the empty table is only the extreme case of a more general one. `eventMapping` exists to *translate* names — to say "our `Purchase` is OpenAI's `order_created`". An event literally named `order_created` needs no translating, so the absence of a row for it is not a decision to exclude it; it is a row nobody had any reason to add. A partially-configured table hits exactly this: rows for the names that needed translating, and a standard-named event alongside them that silently aborted.

So the fallback is unconditional on the table's contents: if no row matches and the name is already a standard OpenAI event, take it at face value.

## Only a literal match

The name is compared case-insensitively — matching how the mapping lookup already compares `from` — and nothing else. No whitespace/hyphen folding, no synonyms, no heuristics. `order_created` and `ORDER_CREATED` resolve; `Order Created` does not, and still needs a mapping row like any other name in a house style that isn't OpenAI's.

Keeping it literal is what makes the unconditional fallback safe. `eventMapping` is the only event gate on the router path — the destination definition's `eventFilteringOption` / `whitelistedEvents` / `blacklistedEvents` sit under `destConfig.web`, the ui-config groups them as "Client-side event filtering" gated on device mode, and rudder-server has no reference to them, so that filtering is SDK-side only. Anything the fallback resolves therefore gets delivered, and this is a conversions API, so that means billing and optimising on it.

With a literal match the only events that newly flow are ones a source deliberately named in OpenAI's own snake_case. Under name folding that would not have held: `src/sources/shopify/config.js` emits events named verbatim `"Checkout Started"` and `"Order Created"`, and `shopify` is in the destination's `supportedSourceTypes` — a Shopify connection mapping only `Order Created` would have started sending `checkout_started` too. Literal matching leaves both of those needing a row, as they should.

## Notes for the reviewer

- **`page` and `screen` are not covered by this.** `getSourceKey` uses `message.name` for them, which is a page title (`"Home"`, `"Docs Page"`) and essentially never a literal OpenAI event name. So the rescue is `track`-only in practice. Defaulting `page` to `page_viewed` would be a coherent follow-up but is a bigger semantic call than this PR should make — flagging it rather than smuggling it in.
- **An explicit mapping still wins.** The table is searched first; the fallback only runs when nothing matched. A row remapping `order_created` to something else keeps working.
- **`EVENT_DATA_TYPE_BY_EVENT` moved to a null prototype.** It is a plain object literal indexed by a config-controlled key, which is a bug class this repo has been bitten by (an event named `constructor` resolving to `Object`). It is not reachable today — the zod schema rejects those values first — but the invariant shouldn't depend on a validation layer two files away staying where it is. There are regression tests pinning `constructor`, `__proto__` and `toString`.
- **The `custom` sentinel deliberately falls through to the error.** `STANDARD_EVENTS` is derived from `STANDARD_EVENT_DATA_TYPES` and `CUSTOM_EVENT_SENTINEL` is a separate const, so an event named `custom` is not a standard event and aborts. That matters because the `customEventName` guard only runs on the mapped branch — if `custom` were ever in that list the fallback would emit `type: 'custom'` with no `custom_event_name`. Nothing in the type system pins this, so there is a test for it.
- **`isStandardEvent` is a type predicate**, not an `as` cast, so the resolved name is genuinely narrowed rather than asserted.

## Testing

- `npx jest src/v0/destinations/openai_ads` — 40 passed
- `npm run test:ts -- component --destination=openai_ads` — 9 passed
- All 12 standard events are covered exhaustively by looping `STANDARD_EVENTS` and asserting the derived `data.type` against `STANDARD_EVENT_DATA_TYPES`, so a 13th event wired to the wrong data type fails the suite.
- Plus: case-insensitive matching (`ORDER_CREATED`, `Order_Created`); that `Order Created` and other non-OpenAI-format names are still rejected; that a standard name absent from a **non-empty** table resolves while a non-standard one alongside it still aborts; that an explicit row wins; and the prototype-chain names above.

### DIFF
--- a/src/v0/destinations/openai_ads/utils.ts
+++ b/src/v0/destinations/openai_ads/utils.ts
@@ -19,6 +19,8 @@ import {
   CUSTOMER_ACTION_DATA_TYPE,
   CUSTOM_EVENT_SENTINEL,
   DESTINATION,
+  EVENT_DATA_TYPES,
+  STANDARD_EVENTS,
   STANDARD_EVENT_DATA_TYPES,
 } from './config';
 import mappingConfig from './data/OPENAI_ADSConfig.json';
@@ -38,10 +40,12 @@ import type {
 const ACTION_SOURCE_SET = new Set<string>(ACTION_SOURCES);
 const CURRENCY_RE = /^[A-Z]{3}$/;
 const PUNCTUATION_REGEX = /[\x21-\x2F\x3A-\x40\x5B-\x60\x7B-\x7E]/g;
-const EVENT_DATA_TYPE_BY_EVENT = {
-  ...STANDARD_EVENT_DATA_TYPES,
+// Null-prototype so an event type of `constructor` or `toString` misses instead of resolving to
+// something off Object.prototype. The zod schema already rejects those before they reach the
+// lookup, but the invariant should not depend on a validation layer two files away staying put.
+const EVENT_DATA_TYPE_BY_EVENT = Object.assign(Object.create(null), STANDARD_EVENT_DATA_TYPES, {
   [CUSTOM_EVENT_SENTINEL]: CUSTOM_EVENT_SENTINEL,
-} as const;
+}) as Record<string, (typeof EVENT_DATA_TYPES)[number]>;
 
 type OpenAIAdsMappingConfig = {
   hashedUserMappings: MappingEntry[];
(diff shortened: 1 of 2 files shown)


src/v0/destinations/openai_ads/utils.ts
```diff
@@ -38,10 +40,12 @@ import type {
 const ACTION_SOURCE_SET = new Set<string>(ACTION_SOURCES);
 const CURRENCY_RE = /^[A-Z]{3}$/;
 const PUNCTUATION_REGEX = /[\x21-\x2F\x3A-\x40\x5B-\x60\x7B-\x7E]/g;
-const EVENT_DATA_TYPE_BY_EVENT = {
-  ...STANDARD_EVENT_DATA_TYPES,
+// Null-prototype so an event type of `constructor` or `toString` misses instead of resolving to
+// something off Object.prototype. The zod schema already rejects those before they reach the
+// lookup, but the invariant should not depend on a validation layer two files away staying put.
+const EVENT_DATA_TYPE_BY_EVENT = Object.assign(Object.create(null), STANDARD_EVENT_DATA_TYPES, {
```
maheshkutty, line 50: Why to complicate this with Object.assign, can't we use?
```suggestion
const EVENT_DATA_TYPE_BY_EVENT: Record<string, typeof EVENT_DATA_TYPES[number]> = {
  [CUSTOM_EVENT_SENTINEL]: CUSTOM_EVENT_SENTINEL,
  ...STANDARD_EVENT_DATA_TYPES,
};
```

shekhar-rudder: The suggestion isn't equivalent to what's there — it silently reintroduces the prototype chain, which is the one thing the line exists to prevent.

`Object.create(null)` here isn't incidental complexity, it's the whole content of the change. With a plain object literal the table inherits from `Object.prototype`, so the lookup at `utils.ts:458` resolves for keys nobody ever put in it:

- `EVENT_DATA_TYPE_BY_EVENT['constructor']` → the `Object` function
- `EVENT_DATA_TYPE_BY_EVENT['__proto__']` → `Object.prototype`
- `EVENT_DATA_TYPE_BY_EVENT['toString']` → a function

and each of those flows straight into `data.type`.

The `Record<string, ...>` annotation doesn't close that. It's erased at runtime, and with `noUncheckedIndexedAccess` off (`tsconfig.json:95`) it arguably makes things worse: it types the lookup as always returning a valid data type, so nothing downstream is ever forced to consider a miss.

This is a bug class we've been bitten by before in this repo — a config-controlled key indexed into a plain object literal resolving to something off `Object.prototype` instead of missing cleanly.

To be clear on the risk today, I agree it's unreachable: `to` is pinned to `z.union([z.enum(STANDARD_EVENTS), z.literal('custom')])` (`types.ts:14`) and every router input is `safeParse`d against that schema before `transformEvent` runs (`processDestinationIntegration.ts:37`). So this is defence-in-depth. But the thing it defends against is someone loosening that enum in a different file — a change that would look entirely harmless in review, precisely because the reason it has to stay strict lives over here. That's the argument for holding the invariant at the lookup rather than borrowing it from the validation layer.

Not a novel pattern in the codebase either — same shape at `src/features.ts:163`, a `Record<string, string>` registry keyed by a config-supplied name.

maheshkutty (dismissed):

maheshkutty (approved):

PR #5564, opened 2026-09-08
fix(mp): treat 409 on create deletion task as already scheduled

shekhar-rudder:
## What

Mixpanel's create-deletion-task API (`POST /api/app/data-deletions/v3.0/`) answers **409 Conflict** when a deletion task already exists for a requested `distinct_id`. `createDeletionTask` treated any non-2xx as a `NetworkError`, so the regulation-worker recorded the job as `failed` and retried it.

Retries send the identical request, so they get the same 409. The job burns every attempt and is then **aborted** — recording a failed deletion for a user Mixpanel is already deleting.

This change treats 409 as success for **single-`distinct_id`** requests, matching the 404 ("nothing left to delete") handling already present in the `custify`, `intercom` and `iterable` deletion handlers.

## Why it matters

A duplicate deletion request — the same user covered by an earlier request — is enough to trigger it. Once it happens the outcome is deterministic: every attempt 409s, the job aborts, and the deletion is recorded as failed even though the vendor is deleting the user.

It also fires the P0 `regulations-error-spike` alert for MP. MP has a single active workspace, so the alert's documented single-customer waiver applies and one aborted job reads as "100% of active MP destinations failing all deletions".

Observed in production: one destination type, one deletion job covering a single user, 5 consecutive attempts, all 409, then abort. The immediately preceding deletions for the same destination succeeded (201), so this was not a credential or connectivity problem.

## Scope — deliberately narrow

The 409 is **not** swallowed for multi-id requests.

Mixpanel does not document 409 for this endpoint, so we cannot tell whether a conflict rejects the whole request or only the conflicting id. If it rejects the whole request, swallowing it would mark every other user in the batch as deleted when none of them were — a silent compliance gap. Failing the batch is the safe side of that unknown.

Worth confirming the exact semantics with Mixpanel support; if a conflict only skips the conflicting id, this can be widened to the whole batch.

## Tests

Two component cases added:

- single `distinct_id` + 409 → job succeeds
- two `distinct_ids` + 409 → job still fails (locks in the narrow scope)

`mp` component suite: 109 passed, 0 failed.

Note: `custom_audience` component tests fail on a clean `develop` checkout, unrelated to this change.

### DIFF
--- a/src/v0/destinations/mp/config.js
+++ b/src/v0/destinations/mp/config.js
@@ -65,6 +65,9 @@ const MAX_PAYLOAD_SIZE_BYTES = 1024 * 1024; // 1MB
 // Delete user
 const DEL_MAX_BATCH_SIZE = 1000;
 const DISTINCT_ID_MAX_BATCH_SIZE = 1999;
+// Returned by the create deletion task api when a deletion task already exists for one or more of
+// the requested distinct_ids.
+const DELETION_TASK_ALREADY_EXISTS_STATUS = 409;
 
 const DESTINATION = 'MP';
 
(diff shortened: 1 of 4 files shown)


shekhar-rudder:
Closing this — capturing the investigation as a Linear ticket instead for now. The branch `worktree-mp-deleteusers-409-already-exists` stays pushed if we want to revive the change.

PR #5567, opened 2026-09-09
feat(logger): gate request/response payload logs behind event level

ItsSudip:

- Adds an `event()` logger wrapper for the new most-verbose `event` level (above `debug`), with a `console.debug` fallback for `LOGGER_IMPL=console`.
- `requestLog`/`responseLog` (proxy delivery request/response payload diagnostics) now emit at `event` level instead of `info`. The `LOG_DEST_IDS`/`LOG_WSP_IDS` metadata filter is unchanged.
- Raises the default `LOG_LEVEL` from `warn` to `info`.
- Log hygiene from the INT-6828 PII audit: braze custom-attribute failure log → `debug` with the error message only; shopify tracker identifier log → `debug`; trackingPlan event-schema fetch failure logs `error.message` instead of `JSON.stringify(error)` (which yields `{}` for Errors).
- Adds `src/logger.test.js` covering event-level gating and request/response log behavior; updates `network.test.js` diagnostics assertions from `info` to `event`.

Resolves INT-6828

Payload-bearing request/response diagnostics must not be visible at `info` so the production log level can be raised from `warn` to `info` without storing event payloads (PII) in logs. Emitting them now requires an explicit `LOG_LEVEL=event` *and* a matching destination/workspace ID.

Activation of the event level depends on `@rudderstack/integrations-lib` picking up rudderlabs/rudder-integrations-lib#486 (adds `event` to `LOGLEVELS`); until that version is released and bumped here, event-level logs are simply suppressed — no behavior regression.

Yes: request/response payload diagnostics move from `info` to `event` level; the default log level (when `LOG_LEVEL` is unset) becomes `info`. Deployments that set `LOG_LEVEL` explicitly are unaffected by the default change.

N/A (a follow-up bump of `@rudderstack/integrations-lib` will activate the event level once rudderlabs/rudder-integrations-lib#486 is released)

`event(...)` added to `src/logger.js`, same gating pattern as the existing `debug/info/warn/error` wrappers.

Validation: `src/logger.test.js` + `src/adapters/network.test.js` 54/54, braze + shopify suites 433/433, `tsc --noEmit` clean, prettier + eslint clean on changed files; full pre-commit suite (`test:ts:silent && test:js:silent && lint-staged`) passed on commit.

<hr>

### Developer checklist

### Reviewer checklist

https://claude.ai/code/session_01Bw2AqSTVYA57BfPSJRq2fS

### DIFF
--- a/src/logger.js
+++ b/src/logger.js
@@ -8,7 +8,7 @@ const { getMatchedMetadata } = require('./util/logger');
 // LOGGER_IMPL can be `console` or `winston`
 const loggerImpl = process.env.LOGGER_IMPL ?? 'winston';
 
-let logLevel = (process.env.LOG_LEVEL ?? 'warn').toLowerCase();
+let logLevel = (process.env.LOG_LEVEL ?? 'info').toLowerCase();
 
 const logger = structuredLogger({
   level: logLevel,
@@ -119,6 +119,14 @@ const log = (logMethod, logArgs) => {
   logMethod(message);
 };
 
+const event = (...args) => {
+  const logger = getLogger();
+  if (LOGLEVELS.event <= LOGLEVELS[logLevel]) {
+    // the console impl has no event method; fall back to debug
+    log(logger.event ?? logger.debug, args);
+  }
+};
+
 const debug = (...args) => {
   const logger = getLogger();
   if (LOGLEVELS.debug <= LOGLEVELS[logLevel]) {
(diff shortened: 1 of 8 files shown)


src/sources/shopify/tracker/transform.js
```diff
@@ -257,7 +257,7 @@ const isIdentifierEvent = (event) =>
 const processIdentifierEvent = async (event, metricMetadata) => {
   const cartToken =
     typeof event.cartToken === 'string' ? event.cartToken.split('?')[0] : event.cartToken;
-  logger.info(`{{SHOPIFY::}} writeKey: ${metricMetadata.writeKey}, cartToken: ${cartToken}`, {
```
krishna2020, line 260: let's remove this log altogether

src/logger.js
achettyiitr, line 1: Can we decouple requestLog with 
- Just logging without depending on filteredMetadata at an event log level.
- In case of filteredMetadata, write it to a file and upload to s3 in periodic basis.

src/sources/shopify/tracker/transform.js
ItsSudip: Removed the log altogether in 7a58267d2.

src/logger.js
ItsSudip: Decoupled in 7a58267d2 — `requestLog`/`responseLog` now emit at the event level unconditionally (`LOG_LEVEL=event` is the only gate; no `filteredMetadata` dependency). The `LOG_DEST_IDS`/`LOG_WSP_IDS` allowlist now gates only the second half: sampled file staging + periodic gzip upload to S3, which is implemented in the stacked PR #5575 (per-worker JSONL staging, request/response pairing by requestId, redaction, 10-min flush cycle).

src/logger.js
```diff
@@ -119,6 +118,14 @@ const log = (logMethod, logArgs) => {
   logMethod(message);
 };
 
+const event = (...args) => {
+  const logger = getLogger();
+  if (LOGLEVELS.event <= LOGLEVELS[logLevel]) {
+    // the console impl has no event method; fall back to debug
```
achettyiitr: Let's check if we can get rid of it, if not being used?

ItsSudip: Removed in b32e84c0d — LOGGER_IMPL was referenced nowhere outside src/logger.js (no compose/Dockerfile/docs/tests), so the console impl and the event-level fallback are gone; all wrappers use the winston structured logger directly. If any environment still sets LOGGER_IMPL=console it simply becomes a no-op (winston structured logs).

ItsSudip: Done.

achettyiitr (approved):

PR #5578, opened 2026-09-11
fix(mp): handle 409 from create deletion task in deleteUsers

achettyiitr:

Fixes MP `deleteUsers` for the create-deletion-task path (`userDeletionApi: 'task'`) when Mixpanel answers **409 Conflict**.

- **409 with `conflicting_distinct_ids`:** those ids already have a deletion task running at Mixpanel, so they are dropped and the rest of the batch is resubmitted. The request succeeds once no ids are left. Any other failure fails the request as before, including a 409 that names none of the requested ids.
- **Rate limit:** requests to the create-deletion-task API are now sent one at a time, at least 1s apart. Mixpanel limits this API to 1 request per second (`429 Too Many Requests, this API is limited to 1 req/s`), and the previous `Promise.all` sent all batches at once.
- **Logging:** when conflicts are dropped, an info log records the conflicting and remaining counts and Mixpanel's task ids. It never logs user ids.

The `/engage` delete-profile path is unchanged.

Resolves INT-7111

Every non-2xx from create-deletion-task was a `NetworkError`, so the regulation-worker marked the job `failed` and retried a byte-identical request. That request got the same 409 every time, and the job was aborted after `REGULATION_DELETION_MAX_FAILED_ATTEMPTS`. It was recorded as a failed deletion for a user Mixpanel was already deleting, and it fired the `regulations-error-spike` P0.

The ticket's open question was whether a 409 rejects the whole request or only the conflicting id. Tested against a real Mixpanel project:

| Request | Mixpanel |
|---|---|
| `[A]` | 200, task created |
| `[A]` again | 409, `conflicting_distinct_ids: [A]` |
| `[A, C]` (C new) | 409, `conflicting_distinct_ids: [A]` only |
| `[C]` straight after | 200, so C was **not** scheduled by the previous request |

**A 409 rejects the whole request.** Treating it as success would silently skip every other user in the batch, which is why only the ids Mixpanel names are dropped. Mixpanel's 409 body is:

```json
{"status":"error","error":{"conflicting_distinct_id_count":1,"conflicting_distinct_ids":["A"],"conflicting_task_ids":["<task id>"],"error":"1 of these distinct IDs are already being deleted by a request that is still running. Wait for <task id> to finish, then submit them again."}}
```

End-to-end reproduction (local stack: data-regulation-service → regulation-worker → transformer → Mixpanel): a second deletion regulation for a user with a running task got 409 on all 5 attempts and was aborted. That matches the production incident.

- A 409 that names conflicting ids no longer fails the job; those users count as deleted, since Mixpanel is deleting them. This matches the guarantee a 200 gives today: task creation, not completion.
- Create-deletion-task batches are sent sequentially, at least 1s apart, instead of all at once.

N/A

N/A

- Each extra request (a further batch of more than 1999 ids, or a resubmit after a 409) waits 1s. A typical job (at most 1000 users per regulation job) is one request, plus one when a conflict is resubmitted.
- The loop always terminates: every resubmit strictly shrinks the id list, and a 409 that names none of the ids fails immediately.
- Tests: the two component cases that resubmit wait on the real 1s timer, so they switch Jest back to real timers through `mp/mocks.ts`. Destinations earlier in the shared component suite turn on fake timers to pin the system time and never switch them off.

<hr>

### Developer checklist

### Reviewer checklist

https://claude.ai/code/session_01TtFmebUFukLMqURdNQW2vX

### DIFF
--- a/src/v0/destinations/mp/deleteUsers.js
+++ b/src/v0/destinations/mp/deleteUsers.js
@@ -1,14 +1,19 @@
 const lodash = require('lodash');
 const { ConfigurationError, NetworkError } = require('@rudderstack/integrations-lib');
 const { handleHttpRequest } = require('../../../adapters/network');
+const logger = require('../../../logger');
 const { isHttpStatusSuccess } = require('../../util');
 const { DEL_MAX_BATCH_SIZE, DISTINCT_ID_MAX_BATCH_SIZE } = require('./config');
 const { executeCommonValidations } = require('../../util/regulation-api');
 const { getDynamicErrorType } = require('../../../adapters/utils/networkUtils');
 const tags = require('../../util/tags');
-const { JSON_MIME_TYPE } = require('../../util/constant');
+const { HTTP_STATUS_CODES, JSON_MIME_TYPE } = require('../../util/constant');
 const { getUserIdBatches } = require('../../util/deleteUserUtils');
 const { getBaseEndpoint, getCreateDeletionTaskEndpoint } = require('./util');
+
+// The create deletion task API allows one request per second; a faster request gets
+// 429 "Too Many Requests, this API is limited to 1 req/s".
+const DELETION_TASK_REQUEST_INTERVAL_MS = 1000;
 
 const deleteProfile = async (userAttributes, config) => {
   const endpoint = `${getBaseEndpoint(config)}/engage`;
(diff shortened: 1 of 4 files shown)


src/v0/destinations/mp/deleteUsers.js
```diff
@@ -86,16 +91,23 @@ const createDeletionTask = async (userAttributes, config) => {
...
-    batchEvents.map(async (batchEvent) => {
-      const request = {
-        distinct_ids: batchEvent,
-        compliance_type: complianceType,
-      };
+  let requestCount = 0;
+  // Requests are sent one at a time to stay within the API's rate limit.
+  for (const batchEvent of batchEvents) {
+    let distinctIds = batchEvent;
+    while (distinctIds.length > 0) {
```
krishna2020, line 98: with the loop here to finish all Ids & with wait time per request, Is the server-transformer timeout gonna get respected still ?

krishna2020: Is there a knob to send only 1 batch from regulation server instead ?

